PR #18, opened 2019-10-29
Update README.md

rix1:
Indicate required and optional arguments.


quietshu:
Replied in https://github.com/zeit/swr/pull/20#issuecomment-547330727

PR #20, opened 2019-10-29
Update `useSWR`-section in readme

clentfort:
Clarify that the passed in fetcher-function must return a promise and is optional


quietshu:
Currently, I don't think we can mark `fetcher` as optional safely because we don't include any `fetch` lib, nor using the native `fetch` function. You can only skip that parameter if there's a `{ fetcher: ... }` option in the config.

clentfort:
I removed the `?`.

PR #28, opened 2019-10-29
Code cleanup and update comments

quietshu:
Unrelated code from ZEIT's product codebase should be removed (the `shouldShowError` part). Also improved the comments.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -170,7 +170,7 @@ function useSWR<Data = any, Error = any>(
           config.onSuccess(newData, key, config)
         }
 
-        // if the revalidation happened eariler than local mutations,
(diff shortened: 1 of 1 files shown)


src/use-swr.ts
```diff
@@ -165,11 +165,12 @@ function useSWR<Data = any, Error = any>(
...
           newData = await CONCURRENT_PROMISES[key]
 
-          // recover error toasts (only for the original request)
+          // trigger the success event
+          // (only do it for the original request)
           config.onSuccess(newData, key, config)
         }
 
-        // If the revalidation happened eariler than local mutations,
+        // if the revalidation happened eariler than local mutations,
```
pacocoursey:
```suggestion
        // if the revalidation happened earlier than local mutations,
```

PR #31, opened 2019-10-29
Add example of implementing optimistic UI with mutate

sergiodxa:
This example uses `mutate` to implement an optimistic UI, the API randomly "fail" to add the data of the POST so it some cases the UI will return to the previous state, in most cases it will work


examples/optimistic-ui/pages/index.js
```diff
@@ -0,0 +1,37 @@
...
+    event.preventDefault()
+    // mutate current data to optimistically update the UI
+    // the fetch below could fail, in that case the UI will
+    // return to the previous state after SWR revalidate it
+    mutate('/api/data', [...data, text])
+    // send text to the API
+    await fetch('/api/data', {
+      method: 'POST',
+      body: JSON.stringify({ text })
+    })
```
quietshu: If we mutate & revalidate before sending the request, the revalidation request (GET) might arrive earlier than the API request (POST), so it could still be out to dated.

Instead, I think we can do this if we want to do local mutation optimistically:
```js
import { mutate, trigger } from '@zeit/swr'

mutate('/api/data', data, false) // set `shouldRevalidate` to false
await fetch(API)
trigger('/api/data') // trigger a revalidation manually
```

rauchg:
Amazing @sergiodxa – We miss you!

quietshu:
I think we'll need READMEs for all the examples in the future, but it's good to merge them for now.

PR #35, opened 2019-10-29
Adding types to ConfigInterface for callbacks

aequasi:


src/types.ts
```diff
@@ -1,4 +1,4 @@
-export interface ConfigInterface {
+export interface ConfigInterface<Data = any, Error = any> {
```
aequasi: Wasnt sure if `Error` should default to `any` or `Error`

aequasi: Added defaults to avoid a BC break

quietshu: Yeah, I think it's good to leave the possibility of throwing anything. There might be some interesting use cases.

quietshu:
You may need to run `yarn lint:fix` to fix the prettier error 🙏

aequasi:
Fixed! Thanks

PR #36, opened 2019-10-29
feature/basic-typescript-example: add basic-typescript example

marcelmokos:
Looking forward to use SWR with typescript with my colleagues in the near future.


examples/basic-typescript/pages/[user]/[repo].tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+import Link from 'next/link'
+import fetch from '../../libs/fetch'
```
Comment (aequasi) [inline examples/basic-typescript/pages/[user]/[repo].tsx]: Is this a pattern worth continuing? Would it make more sense to do this here instead?
```suggestion
import fetch from 'isomorphic-unfetch';
```

Comment (marcelmokos) [reply examples/basic-typescript/pages/[user]/[repo].tsx]: I have used `example/basic` as a template. Both examples are using a custom implementation of fetch that is directly returning `json`. 

https://github.com/zeit/swr/blob/master/examples/basic/pages/%5Buser%5D/%5Brepo%5D.js

Comment (aequasi) [reply examples/basic-typescript/pages/[user]/[repo].tsx]: Doh, I forgot about the Json step!

PR #37, opened 2019-10-29
Adding new 'args' config

aequasi:
Adds the ability to pass arguments to the fetcher function

- [x] Updated documentation
- [x] Added example
- [x] Added test

Looking for more examples to show off the possibilities for this


Timer:
What we're trying to achieve would be a syntax sugar way of doing this:
```js
const fn = useCallback(
  key => {
    doSomething(key, a, b);
  },
  [a, b],
);
const { data, error } = useSWR('my-key', fn);
```

(where `key` & `fn` are a part of the internal dependency arrays)

aequasi:
Hi @timer, I had an offline conversation with @quietshu about this in slack and we landed on this "api"

I can add the deps, thats an easy change.

aequasi:
Hm. that changed caused an infinite loop in the tests...

aequasi:
Updated to the new suggested format: 

```tsx
function useSWR<Data = any, Error = any, Fn extends fetcherFn<Data> = fetcherFn<Data>>(key: keyInterface, fn?: Fn, config?: ConfigInterface<Data, Error>): responseInterface<Data, Error>;
function useSWR<Data = any, Error = any, Fn extends fetcherFn<Data> = fetcherFn<Data>>(key: keyInterface, fn?: Fn, deps?: Parameters<Fn>, config?: ConfigInterface<Data, Error>): responseInterface<Data, Error>;
```

src/types.ts
```diff
@@ -1,3 +1,4 @@
+export type fetcherFn<Data> = (...args: any) => Data
```
clentfort: At least the first argument must always be a string.

src/use-swr.ts
```diff
@@ -57,34 +58,40 @@ const mutate: mutateInterface = function(key, data, shouldRevalidate = true) {
...
-function useSWR<Data = any, Error = any>(
+function useSWR<
+  Data = any,
+  Error = any,
+  Fn extends fetcherFn<Data> = fetcherFn<Data>
+>(
   key: keyInterface,
-  fn?: Function,
+  fn?: Fn,
+  deps?: Parameters<Fn>,
```
clentfort: The react team recommends against having a custom `deps` as the lint-rule will not pick these up automatically. https://twitter.com/dan_abramov/status/1140316424898043904 The recommended way is to pass in a function wrapped via `useMemo` or `useCallback`.

clentfort: Also `deps` being not the last argument would be a very surprising API-design since the deps array is usually the last array in hooks.

aequasi: This isnt a "custom" `deps`. Its just an array. `Parameters` is a typescript type to put the arguments of a function in a tuple.

the eslint plugin for exhaustive deps is expecitng `deps` to be immediately after the function. Hands may be tied here for that

quietshu:
> I do have a concern here, in that we can't pass a function that returns deps for this, not allowing: https://swr.now.sh/#dependent-fetching. I'd love to find a way to allow this while still working with the eslint plugin. Any suggestions?

I think we should still include deps inside the `key` function e.g.:

```
useSWR(() => dep ? 'key' : throwOrNull, () => fetch(dep.id), [dep])
```

examples/fetch-with-credentials/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "basic",
```
quietshu:
```suggestion
  "name": "fetch-with-credentials",
```

It causes a warning in CI.

src/use-swr.ts
```diff
@@ -57,34 +58,40 @@ const mutate: mutateInterface = function(key, data, shouldRevalidate = true) {
   }
 }
```
```suggestion
function depsCompare (arr1, arr2) {
  if (arr1.length !== arr2.length) return false
  for (let i = 0; i < arr1.length; ++i) {
    if (arr1[i] !== arr2[i]) return false
  }
  return true
}

```

src/use-swr.ts
```diff
@@ -124,6 +131,8 @@ function useSWR<Data = any, Error = any>(
   const errorRef = useRef(false)
   const unmountedRef = useRef(false)
   const keyRef = useRef(key)
+  const fnRef = useRef(fn)
```
```suggestion
```

src/use-swr.ts
```diff
@@ -124,6 +131,8 @@ function useSWR<Data = any, Error = any>(
...
+  const fetcherArgsRef = useRef(fetcherArgs)
   const dataRef = useRef(data)
```
```suggestion
const _fn = useCallback(fn, [key, ...fetcherArgs])

```

src/use-swr.ts
```diff
@@ -157,7 +166,7 @@ function useSWR<Data = any, Error = any>(
               if (loading) config.onLoadingSlow(key, config)
             }, config.loadingTimeout)
           }
-          CONCURRENT_PROMISES[key] = fn(key)
+          CONCURRENT_PROMISES[key] = fn(...fetcherArgs)
```
```suggestion
          CONCURRENT_PROMISES[key] = _fn(...fetcherArgs)
```

src/use-swr.ts
```diff
@@ -194,6 +203,8 @@ function useSWR<Data = any, Error = any>(
               trigger(key, false)
             }
             keyRef.current = key
+            fnRef.current = fn
```
```suggestion
```

src/use-swr.ts
```diff
@@ -222,7 +233,7 @@ function useSWR<Data = any, Error = any>(
       loading = false
       return true
     },
-    [key]
+    [key, fn, ...fetcherArgs]
```
```suggestion
    [key, _fn]
```

src/use-swr.ts
```diff
@@ -237,11 +248,18 @@ function useSWR<Data = any, Error = any>(
 
     const _newData = cacheGet(key)
 
-    // update the state if the cache changed OR the key changed
-    if ((_newData && !deepEqual(data, _newData)) || keyRef.current !== key) {
+    // update the state if the cache changed OR the key changed OR the args changed OR the fetcherFn has changed
+    if (
+      (_newData && !deepEqual(data, _newData)) ||
+      keyRef.current !== key ||
+      fnRef.current !== fn ||
```
```suggestion
```

src/use-swr.ts
```diff
@@ -237,11 +248,18 @@ function useSWR<Data = any, Error = any>(
...
+      fetcherArgsRef.current !== fetcherArgs
```
```suggestion
      !depsCompare(fetcherArgsRef.current, fetcherArgs)
```

src/use-swr.ts
```diff
@@ -237,11 +248,18 @@ function useSWR<Data = any, Error = any>(
...
+    ) {
       setData(_newData)
       dataRef.current = _newData
       keyRef.current = key
+      fnRef.current = fn
```
```suggestion
```

src/use-swr.ts
```diff
@@ -277,6 +295,8 @@ function useSWR<Data = any, Error = any>(
         })
         dataRef.current = newData
         keyRef.current = key
+        fnRef.current = fn
```
```suggestion
```

src/use-swr.ts
```diff
@@ -332,7 +352,7 @@ function useSWR<Data = any, Error = any>(
         clearTimeout(id)
       }
     }
-  }, [key, config.refreshInterval, revalidate])
+  }, [key, config.refreshInterval, revalidate, fn, ...fetcherArgs])
```
```suggestion
  }, [key, config.refreshInterval, _fn])
```

src/use-swr.ts
```diff
@@ -351,7 +371,12 @@ function useSWR<Data = any, Error = any>(
     // `key` might be changed in the upcoming hook re-render,
     // but the previous state will stay
     // so we need to match the latest key and data
-    data: keyRef.current === key ? data : undefined,
+    data:
+      keyRef.current === key &&
+      fnRef.current === fn &&
```
```suggestion
```

src/use-swr.ts
```diff
@@ -351,7 +371,12 @@ function useSWR<Data = any, Error = any>(
...
+      fetcherArgsRef.current === fetcherArgs
```
```suggestion
      depsCompare(fetcherArgsRef.current, fetcherArgs)
```

PR #42, opened 2019-10-30
Add React-Native Support

### LINKED ISSUE #34: Support for `React-Native`
Hi, I read the code and you're using ReactDOM. I wonder if you plan to support `React Native`?

n3tr:
resolve #34 

- Create Native-specific extension file to export `unstable_batchedUpdates` function from `react-native` package
  - https://facebook.github.io/react-native/docs/platform-specific-code#native-specific-extensions-ie-sharing-code-with-nodejs-and-web
- Check for browser variables `window` and `document`
- I tried to add example for `react-native` as part of this repo but unable symlink the package locally (probably due to Haste or duplicated react copy). Thus, I created a separated repo for demo and testing instead.
  - Demo: https://github.com/n3tr/ReactNativeSWRExample
 
Screenshot 2019-10-31 at 12 30 06 AM

https://github.com/n3tr/ReactNativeSWRExample/blob/master/App.js#L5-L14


src/libs/reactBatchedUpdates.native.ts
```diff
@@ -0,0 +1,5 @@
+// eslint-disable-next-line
+// @ts-ignore
+import { unstable_batchedUpdates } from 'react-native'
+
+export { unstable_batchedUpdates }
```
sztadii: It was a problem to export it in that way?
```export { unstable_batchedUpdates } from 'react-native'``
Ceause you did it in `src/libs/reactBatchedUpdates.ts`

PR #45, opened 2019-10-30
Stricter typing for fetcher-function

clentfort:
**Summary:**

At the moment the fetcher function is very loosely typed, this could
easily lead to problem, i.e. when passing in a function that does not
return a promise.

**Test plan:**

Run `yarn test` and verify the tests are still passing.


clentfort:
CI fails on a an unrelated file/also fails on an unchanged master.

aequasi:
the fetcher function imo shouldn't require an async function.

quietshu:
Could you please run `yarn lint:fix` to fix the prettier errors?

aequasi:
A valid use-case that would throw type errors with this change: (Taken from @quietshu)

```tsx
const useDateNow = (refreshInterval: number): number => {
    return useSWR('now', () => Date.now(), {refreshInterval}).data // `() => Date.now()` does not return a promise
}

const DynamicAgo = ({ts}) => {
    const now = useDateNow(1000)
    return ms(Math.max(0, now - ts))
}
```

Issue #39, opened 2019-10-30
Enabling Suspense hangs browser


fnky:
I have tried to setup an example project using the SWR hook with GraphQL, as explained in the documentation. However, enabling suspense in the options of `useSWR` causes the browser to hang.

I have an [example project on CodeSandbox](https://codesandbox.io/s/swr-movies-example-zgunu) which shows this. Simply go to either `screens/movies.tsx` or `screens/movie-details.tsx` and uncomment the line with the `// { suspense: true }`.

adwd:
I have same problem but without GraphQL. Minimum app for reproduction is here => https://codesandbox.io/s/react-suspense-with-zeitswr-r5z2k

Nyura95:
Me too. Just a bug with this feature I think 

To reproduce, copy/paste the code below:

<img width="564" alt="Capture d’écran 2019-10-30 à 12 56 15" src="https://user-images.githubusercontent.com/10028203/67856373-ae146280-fb14-11e9-8672-f0913ffc136f.png">

getMessages runs a loop each time and Suspense rests in fallback

erikjung:
Looks like a duplicate of #46

fnky:
Closing this issue as it seems to be resolved in #49 — there were a "Close" keyword, but it wasn't separated for each issue, so it only closed the first issue regarding this.

PR #49, opened 2019-10-30
Fix suspense hanging

### LINKED ISSUE #46: Suspense is not resolved and the requests are looping
```javascript
import React, { Suspense } from 'react';
import useSWR from '@zeit/swr'

function StarWarsSuspense() {
  return (
    <Suspense fallback={<div>loading ...</div>}>
      <StarWars />
    </Suspense>
  )
}

function StarWars() {
  const { data } = useSWR('http://swquotesapi.digitaljedi.dk/api/SWQuote/RandomStarWarsQuote', fetch, { suspense: true });

  return (
    <div>
      {data.starWarsQuote}
    </div>
  );
}

export default StarWarsSuspense;
```
With this code, i never get the final component, the placeholder stay forever, looking on chrome developer console, the requests are looping.

I made this code using the readme example, i tried change de fetch function to a random promise to see if i get out of fallback, without no results.

I doing something wrong or this is really something that should not happen ?

quietshu:
The problem is caused by `useHydration`. 

When enabling SSR with frameworks like Next.js, we have to match the SSRed result with the client rendered result on hydration, so we skipped from reading the cache.  
But when suspense is enabled, we need to read from cache before the component is mounted.

Close #46 #39

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -116,7 +116,9 @@ function useSWR<Data = any, Error = any>(
   }
 
   // stale: get from cache
-  let [data, setData] = useState(useHydration() ? undefined : cacheGet(key))
+  let [data, setData] = useState(
+    config.suspense ? cacheGet(key) : useHydration() ? undefined : cacheGet(key)
+  )
   let [error, setError] = useState()
   let [isValidating, setIsValidating] = useState(false)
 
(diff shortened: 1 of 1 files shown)


src/use-swr.ts
```diff
@@ -116,7 +116,9 @@ function useSWR<Data = any, Error = any>(
   }
 
   // stale: get from cache
-  let [data, setData] = useState(useHydration() ? undefined : cacheGet(key))
+  let [data, setData] = useState(
+    config.suspense ? cacheGet(key) : useHydration() ? undefined : cacheGet(key)
+  )
```
```suggestion
  )
  // it is fine to call `useHydration` conditionally here
  // because `config.suspense` should never change
```

RobertBorg:
I think you intended for this fix to be part of 0.1.2 release but it's not afaik.

quietshu:
Published `swr@0.1.5` 🎉

PR #53, opened 2019-10-31
Use `fast-deep-equal`

quietshu:
https://github.com/epoberezkin/fast-deep-equal is better than `deep-equal` in both bundle size and performance:

```
fast-deep-equal x 325,485 ops/sec ±0.57% (86 runs sampled)
deep-equal x 73,699 ops/sec ±0.55% (86 runs sampled)
```

From bundlephobia.com:
```
fast-deep-equal: 830B MINIFIED
deep-equal: 7.5kB MINIFIED
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "typescript": "3.6.4"
   },
   "dependencies": {
-    "deep-equal": "1.1.0",
     "lodash.throttle": "4.1.1",
     "ms": "2.1.2"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -36,7 +36,6 @@
     "typescript": "3.6.4"
   },
   "dependencies": {
-    "deep-equal": "1.1.0",
```
ycjcl868: need install `fast-deep-equal`

quietshu: Oh thanks a lot! I don’t know why it was missed :)
Will fix

PR #56, opened 2019-11-01
Improve export for IntelliSense

aequasi:
This is an odd change, but without this, the IntelliSense doesn't know that `useSWR` is the export from this package (it tries to import from `swr/dist/use-swr` instead),

Functionally, this only adds a new export to the package: `useSWR`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,5 @@
 export * from './use-swr'
-export { default } from './use-swr'
+export { default as useSWR } from './use-swr'
 export { useSWRPages } from './use-swr-pages'
 export {
   ConfigInterface,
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -1,5 +1,5 @@
 export * from './use-swr'
-export { default } from './use-swr'
+export { default as useSWR } from './use-swr'
```
quietshu:
```suggestion
import { default as useSWR } from './use-swr'
```

aequasi: Want to also export `useSWR` as a non-default? Or just this fix?

quietshu: @aequasi just this fix I think 🙏

quietshu:
Thanks!

PR #63, opened 2019-11-03
add @types/lodash.throttle package

Go7hic:
add @types/lodash.throttle package

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "typescript": "3.6.4"
   },
   "dependencies": {
+    "@types/lodash.throttle": "4.1.1",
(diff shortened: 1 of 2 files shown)


quietshu:
Thank you for your contribution! Could you please update `yarn.lock` as well? (same for #64)
Also the latest version should be `"@types/lodash.throttle": "4.1.6",`.

Go7hic:
> Thank you for your contribution! Could you please update `yarn.lock` as well? (same for #64)
> Also the latest version should be `"@types/lodash.throttle": "4.1.6",`.
It‘s my wrong, i just change code online. I should clone to local install and run it

PR #64, opened 2019-11-03
add @types/ms package

Go7hic:
yarn add @types/ms --dev


quietshu:
We decided to remove `ms` from the dependencies, see #111

PR #67, opened 2019-11-03
Fix error broadcasting

quietshu:
cc @pacocoursey as we discussed somewhere else:

Currently if there're multiple SWR hooks sharing the same key, and one of them revalidates, it will broadcast the updated state to others (so we can deduplicate concurrent requests). But it only broadcasts data updates, not errors. This PR fixes the bug.

Also, some tiny performance improvements are made (using `unstable_batchedUpdates` and reducing the number of cache reads), and related tests are added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,12 +22,11 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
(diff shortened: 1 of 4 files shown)


quietshu:
Thanks!

PR #68, opened 2019-11-03
Fix suspense related bugs

### LINKED ISSUE #57: Bug: Suspense Data undefined Race Condition
I'm not sure if this is a bug, or if this behavior is intended, but I was running into an issue with the line of code below. 

So, I'm using `useSWR(`/api/users/${id}`, fetch, {suspense: true})` where the key is dynamic based on the id of the object I want to get (in this contrived example the user's id). 

My code looks something like this exampled:

```
function useUser(id) {
  const user = useSWR(`/api/users/${id}`, fetch, {suspense: true})
  console.log(user)
  return user
}
```

When this hook is called multiple times with different IDs, the logs would look like this:

```
// render 1 with user id 1
{ id:1 }
// render 2 with user id 2
undefined
// render 3
{ id: 2 }
```

The reason behind this is the two LOC referenced below.

So, my question is, in render 2 on L342 `typeof data = Object`, not undefined because the data is reset on line 358 (afterwards). Should there be a line before line 342 that checks to see if the `keyRef.current` and the new `key` are then different and reset data there so that the revalidate can happen immediately? 

I quickly implemented this locally and this fixed my issue and prevented render 2 happening where the data was undefined.

Thanks! Loving this library so far.

https://github.com/zeit/swr/blob/9b51b7d066335d7249d114e2b9a2cb5d633795b6/src/use-swr.ts#L342
https://github.com/zeit/swr/blob/9b51b7d066335d7249d114e2b9a2cb5d633795b6/src/use-swr.ts#L358

### LINKED ISSUE #66: BUG: Infinite loop with suspense on rejected promise
See: https://codesandbox.io/s/swr-playground-zloyr

Steps to reproduce:
- Use SWR
- Set suspense to true
- Set shouldRetryOnError to false
- Reject HTTP request (or just promise)

Actual result:
Endless loop of retrying

Expected result:
Error boundary should catch error'ed promise

quietshu:
Fixes #57, fixes #66.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -339,14 +339,28 @@ function useSWR<Data = any, Error = any>(
   }, [key, config.refreshInterval, revalidate])
 
   // suspense (client side only)
-  if (config.suspense && typeof data === 'undefined') {
+  if (
+    config.suspense &&
+    typeof data === 'undefined' &&
+    typeof error === 'undefined'
+  ) {
     if (typeof window !== 'undefined') {
-      if (!CONCURRENT_PROMISES[key]) {
+      if (typeof CONCURRENT_PROMISES[key] === 'undefined') {
         // need to trigger revalidate immediately
         // to throw the promise
         revalidate()
       }
-      throw CONCURRENT_PROMISES[key]
+      if (
+        CONCURRENT_PROMISES[key] &&
+        typeof CONCURRENT_PROMISES[key].then === 'function'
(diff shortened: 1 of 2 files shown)


jcarroll2007:
Is there anything I could do to help move this along?

quietshu:
@jcarroll2007 thanks, I moved this PR to #82 and it is ready for reviewing & testing!

PR #82, opened 2019-11-06
Fix suspense bugs

### LINKED ISSUE #57: Bug: Suspense Data undefined Race Condition

### LINKED ISSUE #66: BUG: Infinite loop with suspense on rejected promise

quietshu:
Bug fixes
- fixed #57 and added a test case for it
- fixed #66 and added a test case for it

Also
- introduced cache for errors to support suspense
- changed `noDedupe` to `dedupe` (undocumented API), which is more clear and removes 1 hook inside  (`useCallback`)
- improved code readability

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -116,7 +116,9 @@ function useSWR<Data = any, Error = any>(
   }
 
   // stale: get from cache
-  let [data, setData] = useState(useHydration() ? undefined : cacheGet(key))
+  let [data, setData] = useState(
+    config.suspense ? cacheGet(key) : useHydration() ? undefined : cacheGet(key)
+  )
   let [error, setError] = useState()
   let [isValidating, setIsValidating] = useState(false)
 
@@ -335,7 +337,7 @@ function useSWR<Data = any, Error = any>(
   }, [key, config.refreshInterval, revalidate])
 
   // suspense (client side only)
-  if (config.suspense && !data) {
+  if (config.suspense && typeof data === 'undefined') {
     if (typeof window !== 'undefined') {
       if (!CONCURRENT_PROMISES[key]) {
         // need to trigger revalidate immediately
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -446,6 +446,7 @@ function useSWR<Data = any, Error = any>(
         CONCURRENT_PROMISES[key] &&
         typeof CONCURRENT_PROMISES[key].then === 'function'
       ) {
+        console.log(CONCURRENT_PROMISES[key], key)
         // if it is a promise
         throw CONCURRENT_PROMISES[key]
       }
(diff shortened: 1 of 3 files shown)


src/use-swr.ts
```diff
@@ -361,21 +413,60 @@ function useSWR<Data = any, Error = any>(
...
       }
     }
   }, [key, config.refreshInterval, revalidate])
 
-  // suspense (client side only)
-  if (config.suspense && typeof data === 'undefined') {
-    if (typeof window !== 'undefined') {
+  // suspense
+  if (config.suspense) {
+    if (IS_SERVRE)
```
pacocoursey:
```suggestion
    if (IS_SERVER)
```

src/use-swr.ts
```diff
@@ -33,26 +33,34 @@ import SWRConfigContext from './swr-config-context'
 import isDocumentVisible from './libs/is-document-visible'
 import useHydration from './libs/use-hydration'
 
+const IS_SERVRE = typeof window === 'undefined'
```
pacocoursey:
```suggestion
const IS_SERVER = typeof window === 'undefined'
```

src/use-swr.ts
```diff
@@ -361,21 +413,60 @@ function useSWR<Data = any, Error = any>(
...
+        // to get the promise
         revalidate()
       }
-      throw CONCURRENT_PROMISES[key]
+
+      if (
+        CONCURRENT_PROMISES[key] &&
+        typeof CONCURRENT_PROMISES[key].then === 'function'
+      ) {
+        console.log(CONCURRENT_PROMISES[key], key)
```
```suggestion
```

PR #88, opened 2019-11-06
Add an autocomplete suggestions examples

### LINKED ISSUE #84: [Question] Async autocomplete example
I'm looking to implement SWR in a project but I'm wondering if it would be beneficial to use this for a type of async autocomplete component and if so how would that look, all the current examples are based on that you have static data that you get when rendering but how would it look when you have data you want to load based on what you type like a autocomplete?

sergiodxa:
It's using Reach UI Combobox and fetching a list of countries from an API

Resolves #84


examples/autocomplete-suggestions/pages/index.js
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <div style={{ textAlign: 'center' }}>
+      <h1>Country Search</h1>
+      <Combobox>
+        <ComboboxInput
+          className="country-search-input"
+          onChange={handleSearchTermChange}
+          aria-label="Countries"
+        />
+        {!isValidating && countries.length > 0 && (
```
quietshu:
```suggestion
        {countries && countries.length > 0 && (
```

There're some cases that `isValidating` is true but we have the data (e.g.: when focusing on the tab).

PR #89, opened 2019-11-06
Use `@zeit/ncc`

quietshu:
So we can have a smaller package build (`-m`).

https://github.com/zeit/ncc


quietshu:
This might affect the `react-native` peer dependency. Made this PR on hold.

Issue #93, opened 2019-11-07
how to use post method and pass params in swr?


nai6514531:

monman2l:
const fetcher = params => url => post(url, params)
const { data, error } = useSWR('/api/user', fetcher({key: val}))

This way?

sergiodxa:
SWR is more intended to get data, not update it, usually you have another function to update and after it you will run mutate to update the cache and trigger a revalidation (SWR will fetch it again)

sergiodxa:
This example https://github.com/zeit/swr/tree/master/examples/optimistic-ui does something similar, update and then revalidate

quietshu:
```js
function App() {
  const { data } = useSWR('/api/data', fetch) // <--------------GET

  return <>
    {data}
    <button onClick={() => {
      fetch('/api/data', { method: 'POST', body: ... })  // <---POST
    }}>update data</button>
  </>
}
```

You can just use `fetch` for POST requests. I don't see the benefit of allowing POST in SWR.  
And most importantly, most of the great features will be gone with POST (because it's not a "pure" action like GET, which is perfect for fetch-as-you-render):
- stale-while-revalidate (SWR)
- focus revalidate / polling
- automatically request deduplication
- caching
- ...

But speaking of revalidation/mutation, it's still a thing for allowing POST together with SWR. Currently we need to do this:

```js
mutate('/api/data', newData, false) // local mutate without revalidation
await sendPost('/api/data', newData) // POST request
trigger('/api/data') // revalidate
```

Maybe we can simplify it to something like:

```js
mutate('/api/data', newData, sendPost)
```

which will do the exact same thing as above.

nicolasschabram:
Maybe it'd be worth checking out `react-query`'s [useMutation()](https://github.com/tannerlinsley/react-query#usemutation) hook for reference as well.

iammarkps:
I have a use case with firebase onCall function(which it's method is POST). Because it supports authentication by default, I have to use this workaround.

```
import firebase from '../lib/firebase'

export const fetchSubmissionData = async (id: string) => {
  const getSubmission = firebase
    .app()
    .functions('asia-east2')
    .httpsCallable('getDetailedSubmissionData')

  return await getSubmission({ submission_id: id })
}
```
Then I can use
```
 const { data } = useSWR(`${id}`, fetchSubmissionData)
```

My function is onCall because it needs to verify user before sending back data.

Any idea on improving this? @quietshu

quietshu:
[...]

PR #98, opened 2019-11-08
Args as key

### LINKED ISSUE #23: Example for `urql`?
Is anyone able to write an example for use with `urql`? I'm currently experimenting with `swr`, recently taking up the JAMstack mindset.

### LINKED ISSUE #40: Using GraphQL doesn't always revalidate
There's a problem with revalidating data in some circumstances which can be seen in [this example project](https://codesandbox.io/s/swr-movies-example-zgunu). This is based on the 

1. Navigate to a details page by clicking on one of the items.
2. See the data load, and correctly show.
3. Go back to the previous page.
4. Click on another item.
5. See that it sometimes doesn't revalidate/update the state for the given page.

Note that the project uses `@reach/router` which provides a param `slug` through the `MovieDetailsScreen` component, as it is a route component.

My theory is that the revalidation doesn't happen because the "key" (GraphQL) is always the same even though the provided variable changes and therefore doesn't revalidate, but I don't know exactly the inner workings of `useSWR` with regards to GraphQL.

quietshu:
This is a simple and efficient implementation of a frequently requested feature. 

Currently if you want to pass outside args to fetcher (which may change during the lifecycle):
```js
useSWR('key', () => fetcher(...args))
```
It doesn't work, because **the cache is indexed by the key**. Which means if your `args` change, a SWR refetch will not be triggered.

### Use Cases
GraphQL:
```js
const personQuery = `
  query Person(id: ID, name: String) {
    person(id: $id, name: $name) {
      id
    }
  }
`

// Reuse same query with different variables.
request(personQuery, { variables: { id: 42 } })
request(personQuery, { variables: { name: "Alice" } })
```

Multiple params:
```js
const token = useToken()

// SWR will still read the stale data even if `token` changes
const { data } = useSWR('/api/user', url => fetchAPI(url, token))
```

### New API

This implementation doesn't break any existing API or behavior, but allows you to use an array as the `key` param:

```js
useSWR([...args], fetcher, options)
```

And `arg` can be any value, object or even function. SWR will use those args as the key of the cache, and pass them as arguments to `fetcher`:

```js
fetcher(...args)
```

#### Notes
Same as React Hook deps, **don't** pass anonymous objects/arrays/functions, e.g.:
```js
const x = props.x

// this is wrong, because deps are compared shallowly (===)
useSWR(['hello', {value: x}], fetcher) 

// you can use useMemo
const v = useMemo(() => ({value: x}), [x])
useSWR(['hello', v], fetcher)

// or handle the logic inside fetcher
useSWR(['hello', x], (str, v) => fetcher(str, { value: v }))
```

Related: #37, #93.
Closes #40, closes #23.

### DIFF
--- a/src/libs/hash.ts
+++ b/src/libs/hash.ts
@@ -1,5 +1,5 @@
 // use WeakMap to store the object->key mapping
-// so the objects can be garbadge collected.
+// so the objects can be garbage collected.
 // WeakMap uses a hashtable under the hood, so the lookup
 // complexity is almost O(1).
 const table = new WeakMap()
--- a/src/libs/hash.ts
+++ b/src/libs/hash.ts
@@ -1,0 +1,28 @@
+// use WeakMap to store the object->key mapping
+// so the objects can be garbadge collected.
+// WeakMap uses a hashtable under the hood, so the lookup
+// complexity is almost O(1).
+const table = new WeakMap()
+
+// counter of the key
+let counter = 0
+
+// hashes an array of objects and returns a string
+export default function hash(args: any[]): string {
+  let key = 'arg'
+  for (let i = 0; i < args.length; ++i) {
+    let _hash
+    if (typeof args[i] !== 'object') {
+      _hash = String(args[i])
+    } else {
+      if (!table.has(args[i])) {
+        _hash = counter
+        table.set(args[i], counter++)
+      } else {
+        _hash = table.get(args[i])
+      }
+    }
+    key += '@' + _hash
+  }
+  return key
+}
--- a/test/use-swr.test.tsx
+++ b/test/use-swr.test.tsx
@@ -189,7 +189,7 @@ describe('useSWR', () => {
     expect(container.textContent).toMatchInlineSnapshot(`"err err err"`)
   })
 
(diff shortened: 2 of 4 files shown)


aequasi:
Closes #37 , would re-implement type-fixes in a separate PR

src/libs/hash.ts
```diff
@@ -0,0 +1,28 @@
+// use WeakMap to store the object->key mapping
+// so the objects can be garbadge collected.
```
pacocoursey:
```suggestion
// so the objects can be garbage collected.
```

test/use-swr.test.tsx
```diff
@@ -188,6 +188,39 @@ describe('useSWR', () => {
     await act(() => new Promise(res => setTimeout(res, 100)))
     expect(container.textContent).toMatchInlineSnapshot(`"err err err"`)
   })
+
+  it.only('should accept object args', async () => {
```
pacocoursey:
```suggestion
  it('should accept object args', async () => {
```

quietshu: Thanks for pointing this out! 👍

Issue #11, opened 2019-10-28
Server-side rendering


klarstrup:
Any plans to add SSR support?

Southclaws:
I think this should be trivial once cache API is exposed, as mentioned in #4
> Finally, we plan to export the cache APIs from the lib, and add options to the config soon. So you can override it with your own implementation like LRU cache, or just manipulate the cache object.
You could replace it with an isomorphic cache.

mattfwood:
I imagine the long-term plan for this library is to leverage Suspense + SSR once Suspense is fully released, and I think Next.js will likely do the same. But I agree that in the meantime it would be nice for it to have support in Next.js or other SSR setups to fetch data ahead of time.

quietshu:
Yeah indeed. When we have Suspense supported in server side, we will have this feature out of the box (by just using `suspense: true`).

> But I agree that in the meantime it would be nice for it to have support in Next.js or other SSR setups to fetch data ahead of time.

And yes, we’ve been doing this internally w/ SWR and Next.js for a long time in our dashboard. Will definitely adding the feature to this library. PR soon, stay tuned!

quietshu:
Hey everyone! With the support of the new `initialData` option in v0.1.10, you can now use `getInitialProps` with SWR to do SSR in Next.js without changing too much:

```js
App.getInitialProps = async getInitialProps () {
  const data = await fetcher('/api/data')
  return { data }
}

function App (props) {
  const initialData = props.data
  const { data } = useSWR('/api/data', fetcher, { initialData })

  return <div>{data}</div>
}
```

In this example, everything works the same just like a normal SSR Next.js app. But it’s also fully powered by SWR in the client side, which means the data can be dynamic and update itself over time / user interactions.

Let us know what you think about this :)

quietshu:
Closing this issue  🎉

mirshko:
@quietshu heya! 

Seems theres an issue when using isomorphic-unfetch returning an error "Only Absolute URLs are supported" when fetching from getInitialProps.

Any idea about best way around this.

sergiodxa:
That is not an issue of SWR. It's how sending HTTP requests works server side, you can't use `/api/users` because the server doesn't know to which host send the request, you need to add `localhost:3000` or your production domain.

Maybe you could use env variables to know if you are in development use `localhost:3000` and if you are in production use your domain. You should also be able to dynamically get the host using the headers of the request if you want to make it dynamic but not sure how such a good practice is that.

[...]

PR #106, opened 2019-11-09
Adding initialData

aequasi:
Can partially help in resolving the issue: #11 in the interim 

(worded weirdly to avoid a auto-close)

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -18,7 +18,8 @@ import {
   triggerInterface,
   mutateInterface,
   broadcastStateInterface,
-  responseInterface
+  responseInterface,
+  fetcherFn
 } from './types'
 
 import defaultConfig, {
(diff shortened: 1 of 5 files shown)


src/use-swr.ts
```diff
@@ -157,7 +158,7 @@ function useSWR<Data = any, Error = any>(
   const shouldReadCache = config.suspense || !useHydration()
 
   // stale: get from cache
-  let [data, setData] = useState(shouldReadCache ? cacheGet(key) : undefined)
+  let [data, setData] = useState(config.initialData || (shouldReadCache ? cacheGet(key) : undefined))
```
quietshu: I like the idea, but I think `config.initialData` should be the fallback value. 

```suggestion
  let [data, setData] = useState((shouldReadCache ? cacheGet(key) : undefined) || config.initialData)
```

src/use-swr.ts
```diff
@@ -157,7 +158,7 @@ function useSWR<Data = any, Error = any>(
   const shouldReadCache = config.suspense || !useHydration()
 
   // stale: get from cache
-  let [data, setData] = useState(shouldReadCache ? cacheGet(key) : undefined)
+  let [data = config.initialData, setData] = useState((shouldReadCache ? cacheGet(key) : undefined) || config.initialData)
```
aequasi: So, without this `data = config.initialData`, the very first render doesn't have the correct data in it. This definitely has a code-smell, but I'm not sure how to do it better?

quietshu: Yeah normally the first render won't have cache available, so it will always be undefined. I think the solution will be either you provide `initialData` (e.g. from getInitialProps if you're using SSR) or use suspense.

quietshu:
BTW, can you add this new option to the readme?

PR #109, opened 2019-11-11
note on `key` parameter

shoeman22:
I'm using this library with a services layer, so rather than a string, I pass in something like this for the key my fetcher needs:

```js
const key = { service: 'myFavoriteService', method: 'getProducts', payload: { 'keyword': 'pants' }}
const { data, error }  = useSWR(key, myServiceFetcher)
```

Found out I need to just `JSON.stringify` my `key` to get the caching behavior to work properly:
```js
const { data, error }  = useSWR(JSON.stringify(key), () => myServiceFetcher(key))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,8 @@ const { data, error, isValidating, revalidate } = useSWR(key, fetcher, options)
 
 #### Parameters
 
-- `key`: a unique key string for the request (or a function / null) [(advanced usage)](#conditional-fetching)  
+- `key`: a unique key string for the request (or a function / null) [(advanced usage)](#conditional-fetching)
+   NOTE: `key` cannot be an object.  Suggest wrapping key in `JSON.stringify(key)` to accomodate.
 - `fetcher`: (_optional_) a Promise returning function to fetch your data [(details)](#data-fetching) 
 - `options`: (_optional_) an object of options for this SWR hook
 


quietshu:
`JSON.stringify` is not that efficient. You might want to check this: https://github.com/zeit/swr/tree/improve-readme#multiple-arguments

shoeman22:
@quietshu thanks, I'd missed that part of the readme and didn't realize `JSON.stringify` was a performance concern.  For my use case, I'll still need to do something on the `payload` part of the key, but I can pull the other parameters out to make it more efficient at least:

```js
const key = [  'myFavoriteService', 'getProducts', JSON.stringify({ 'keyword': 'pants' })]
```
I think I may also take a look at [fast-json-stringify ](https://www.npmjs.com/package/fast-json-stringify) to optimize the bit I still need to stringify as well.

Thanks

sergiodxa:
Instead of stringify the last element you could use `React.useMemo` to memoize and pass it as an object.

shoeman22:
> Instead of stringify the last element you could use `React.useMemo` to memoize and pass it as an object.

Could you show me pseudo code for what  you're thinking?  When I try `useMemo` with an object as a dependency, `useMemo` fires multiple times.

```js
const payloadKey = React.useMemo(() => stringify(payload), [payload])
```

This only runs once:
```js
const payloadKey = React.useMemo(() => stringify(payload), [stringify(payload)])
```
But is useless.

sergiodxa:
```js
function App({ keyword }) {
  // create a memoized payload
  const payload = React.useMemo() => {
    return { keyword }
  }, [keyword]);

  // here you pass the payload as any other element without stringifying it
  const { data } = useSWR([ "myFavoriteService", "getProducts", payload ], myServiceFetcher);
  // render something here
}
```

Something like that

shoeman22:
@sergiodxa oh ok, but it looks like there's no great way to use the actual `payload` object blindly as the memo key, right?

In my actual use case, I'm making my own hook that extends `useSWR` and blindly forwards `payload` so I don't know the structure of it ahead of time.  I'd need something like:

```js
const payloadKeyIfied = React.useMemo() => {
    return stringify(payload)
  }, [payload]);
```

But when I use an object as the memo dependency directly, it runs multiple times for the same looking payload.  From research, my understanding is this is just how JS handles objects -- even if they look the same, they are still different with regard to how useMemo checks them.

PR #115, opened 2019-11-13
Remove/inline dependencies

### LINKED ISSUE #111: Optimize dependencies
- remove `ms`
- inline lodash/throttle, no need to include the entire package

quietshu:
Closes #111.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3103,7 +3103,7 @@ ms@2.0.0:
   resolved "https://registry.yarnpkg.com/ms/-/ms-2.0.0.tgz#5608aeadfc00be6c2901df5f9861788de0d597c8"
   integrity sha1-VgiurfwAvmwpAd9fmGF4jeDVl8g=
 
-ms@2.1.2, ms@^2.1.1:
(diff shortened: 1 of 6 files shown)


yarn.lock
```diff
@@ -3103,7 +3098,7 @@ ms@2.0.0:
   resolved "https://registry.yarnpkg.com/ms/-/ms-2.0.0.tgz#5608aeadfc00be6c2901df5f9861788de0d597c8"
   integrity sha1-VgiurfwAvmwpAd9fmGF4jeDVl8g=
 
-ms@2.1.2, ms@^2.1.1:
+ms@^2.1.1:
```
aequasi: Shouldn't these have been removed entirely?

quietshu: Oh weird, I'll delete yarn.lock and regenerate one. Thanks!

quietshu: @aequasi I believe is that some dependency depends on `ms`... just tried again.

aequasi: Strange!

PR #120, opened 2019-11-13
Add example on how to use fetch + mutate to prefetch data

sergiodxa:


examples/prefetch/pages/index.js
```diff
@@ -0,0 +1,31 @@
...
+    data.forEach(project => {
+      fetch(`https://api.github.com/repos/${project}`)
+        .then(projectData => {
+          mutate(`/api/data?id=${project}`, projectData, false)
+        })
+    })
+  }, [data]);
+
+  return <div style={{ textAlign: 'center' }}>
+    <h1>Trending Projects</h1>
```
quietshu: Can you also add a preload tag here:
```js
<NextHead>
  <link preload="/api/data" as="fetch" crossOrigin="anonymous" />
</NextHead>
```

So it will preload even before JS starts to download.  I think these two (preload | prefetch) are good examples and worth mentioning in the readme and website.

sergiodxa: I was thinking of using a second example for the preload one, would you prefer to have them in the same example?

quietshu: Yeah, kinda because if you want to do some optimization, it's great to do both. But also we can add some other cool examples e.g.: 
- wrap the prefetch part as a function and call it when hovering the link
- call the prefetch outside of the component (so it starts to load before React does)

sergiodxa: I added all of this, and mentioned in the READMe and you may not need all of them at the same time

quietshu:
Let's merge it first 👍

PR #121, opened 2019-11-13
Add a prefetch example that uses a combination of fetch and mutate

samchamberland:
This PR adds a basic example of how one could prefetch data using a combination of `fetch` and `mutate`.


samchamberland:
Apologies, it looks like #120 beat me to it 🙃 
I can look into switching the example from prefetch to preload.

PR #132, opened 2019-11-18
Add Server Side Render Example

sergiodxa:
Based on https://github.com/zeit/swr/issues/11#issuecomment-554847046


examples/server-render/pages/index.js
```diff
@@ -0,0 +1,24 @@
...
+      data ? data.map(project => 
+        <p key={project}><Link href='/[user]/[repo]' as={`/${project}`}><a>{project}</a></Link></p>
+      ) : 'loading...'
+    }
+    </div>
+  </div>
+}
+
+Home.getInitialProps = async () => {
+  const data = await fetcher('http://localhost:3000/api/data')
```
quietshu: Hmmm this won't work for deployments... should we use a public API as the example? e.g.: https://pokeapi.co

sergiodxa: You are right, I changed it to use Pokeapi and render a list of pokemon in `/` and a detail in `/[pokemon]`

PR #134, opened 2019-11-19
Add async mutation support

quietshu:
With this change, the `data` parameter of `mutate` can be a promise, which returns the data or throws an error:

```js
mutate(key, Promise<Data>)
```

It can simplify scenarios like

```js
const newData = await update(...)   // some POST APIs return the updated data
mutate('/api/data', newData, false) // disable the revalidation here
```

to:

```js
mutate('/api/data', update(...)) // POST + mutate w/o revalidation
```

Related: #93.

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,6 +326,18 @@ function Profile () {
 }
 ```
 
+In the example above, when clicking the button, it will send the request, **then** update the client data and 
+try to fetch the latest one (revalidate).
+
+But many POST APIs will just return the updated data directly, so we don’t need to revalidate again.  
+Here’s an example showing the “local mutate - request - update” usage:
+
+```js
+mutate('/api/user', newUser, false)      // use `false` to mutate without revalidation
+mutate('/api/user', updateUser(newUser)) // `updateUser` is a Promise of the request,
+                                         // which returns the updated document
+```
+
 ### Suspense Mode
 
 You can enable the `suspense` option to use SWR with React Suspense:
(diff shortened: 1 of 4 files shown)


src/use-swr.ts
```diff
@@ -92,21 +92,42 @@ const broadcastState: broadcastStateInterface = (key, data, error) => {
...
   if (!key) return
 
   // update timestamp
   MUTATION_TS[key] = Date.now() - 1
 
-  // update cached data
-  cacheSet(key, data)
+  let data, error
+
+  if (_data && typeof _data.then === 'function') {
```
aequasi: You *can* await a non-promise too. Not sure if there is a performance issue though
```suggestion
  const data = await _data;
```

quietshu: It will defer the execution order (await non-promise values), similar to `process.nextTick`. But the major reason is that for promises (which should be mutation requests), we don't have to revalidate again after the request in most cases.

README.md
```diff
@@ -338,6 +338,18 @@ function Profile () {
 }
 ```
 
+In the example above, when clicking the button, it will send the request, **then** update the client data and
```
pacocoursey: I think this is a little confusing because we're referring to two different requests here, a POST request to change the data in your remote data source, and then another GET request (the revalidation).

Here's my attempt to clarify the steps:

```suggestion
Clicking the button in the example above will send a POST request to modify the remote data, locally update the client data and
```

quietshu: Thanks! That’s indeed clearer to me 👏

PR #135, opened 2019-11-19
Improve readme

quietshu:
- add SSR 
- explain that args need to be stable on every render

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ You can also use [global configuration](#global-configuration) to provide defaul
 - [Multiple Arguments](#multiple-arguments)
 - [Manually Revalidate](#manually-revalidate)
 - [Local Mutation](#local-mutation)
+- [SSR with Next.js](#ssr-with-nextjs)
 - [Suspense Mode](#suspense-mode)
 - [Error Retries](#error-retries)
 
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ useSWR('/api/data', url => fetchWithToken(url, token))
 ```
 
 This is **incorrect**. Because the identifier (also the index of the cache) of the data is `'/api/data'`, 
-so even if `token` changes, SWR will still get the same key and return the wrong data. 
+so even if `token` changes, SWR will still have the same key and return the wrong data. 
 
 Instead, you can use an **array** as the `key` parameter, which contains multiple arguments of `fetcher`:
 
@@ -265,9 +265,9 @@ Instead, you can use an **array** as the `key` parameter, which contains multipl
 useSWR(['/api/data', token], fetchWithToken)
 ```
 
-It solves the problem. The key of the request is now the combination of both values. SWR **shallowly** compares
-the arguments on every render, and triggers the validation if any of them has changed.  
-So keep in mind to not recreate objects when rendering:
+This solves the problem. The key of the request is now the combination of both values. SWR **shallowly** compares
+the arguments on every render, and triggers revalidation if any of them has changed.  
+Keep in mind that you should not recreate objects when rendering, as they will be treated as different objects on every render:
 
 ```js
 // Don’t do this! Deps will be changed on every render.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -251,22 +253,32 @@ function MyProjects () {
...
 ```js
-const token = props.token
-
 useSWR('/api/data', url => fetchWithToken(url, token))
 ```
 
-**This is incorrect**. Because the identifier of the data is `'/api/data'`, which is also the index of the cache. 
-When `token` changes, SWR will still treat it as the same key and request. 
+This is **incorrect**. Because the identifier (also the index of the cache) of the data is `'/api/data'`, 
+so even if `token` changes, SWR will still get the same key and return the wrong data.
```
pacocoursey:
```suggestion
so even if `token` changes, SWR will still have the same key and return the wrong data. 
```

README.md
```diff
@@ -251,22 +253,32 @@ function MyProjects () {
...
-Instead, you can use an array as the `key` parameter, which contains multiple arguments of `fetcher`:
+Instead, you can use an **array** as the `key` parameter, which contains multiple arguments of `fetcher`:
 
 ```js
 useSWR(['/api/data', token], fetchWithToken)
 ```
 
-This solves the problem. The identifier of the request is now the combination of both values. SWR **shallowly** compares
-the arguments on every render, and triggers the validation if any of them has changed.
+It solves the problem. The key of the request is now the combination of both values. SWR **shallowly** compares
```
pacocoursey:
```suggestion
This solves the problem. The key of the request is now the combination of both values. SWR **shallowly** compares
```

README.md
```diff
@@ -251,22 +253,32 @@ function MyProjects () {
...
+the arguments on every render, and triggers the validation if any of them has changed.
```
pacocoursey:
```suggestion
the arguments on every render, and triggers revalidation if any of them has changed.  
```

README.md
```diff
@@ -251,22 +253,32 @@ function MyProjects () {
...
 
 ```js
 useSWR(['/api/data', token], fetchWithToken)
 ```
 
-This solves the problem. The identifier of the request is now the combination of both values. SWR **shallowly** compares
-the arguments on every render, and triggers the validation if any of them has changed.
+It solves the problem. The key of the request is now the combination of both values. SWR **shallowly** compares
+the arguments on every render, and triggers the validation if any of them has changed.  
+So keep in mind to not recreate objects when rendering:
```
pacocoursey:
```suggestion
Keep in mind that you should not recreate objects when rendering, as they will be treated as different objects on every render:
```

README.md
```diff
@@ -251,22 +253,32 @@ function MyProjects () {
...
+```js
+// Don’t do this! Deps will be changed on every render.
+useSWR(['/api/user', { id }], query)
+
+// Make sure objects are stable
+const params = useMemo(() => ({ id }), [id])
+useSWR(['/api/user', params], query)
+```
+
+Dan Abramov explained deps very well in [his blog post](https://overreacted.io/a-complete-guide-to-useeffect/#but-i-cant-put-this-function-inside-an-effect).
```
pacocoursey:
```suggestion
Dan Abramov explains dependencies very well in [this blog post](https://overreacted.io/a-complete-guide-to-useeffect/#but-i-cant-put-this-function-inside-an-effect).
```

README.md
```diff
@@ -348,7 +382,10 @@ function App () {
 }
 ```
 
-Note in Suspense mode, `data` is always the fetch response (so you don't need to check if it's `undefined`). But if there's an error occurred, you need to use an [error boundary](https://reactjs.org/docs/concurrent-mode-suspense.html#handling-errors) to catch it.
+In Suspense mode, `data` is always the fetch response (so you don't need to check if it's `undefined`). 
+But if there's an error occurred, you need to use an [error boundary](https://reactjs.org/docs/concurrent-mode-suspense.html#handling-errors) to catch it.
```
pacocoursey:
```suggestion
But if an error occurred, you need to use an [error boundary](https://reactjs.org/docs/concurrent-mode-suspense.html#handling-errors) to catch it.
```

README.md
```diff
@@ -348,7 +382,10 @@ function App () {
...
+
+_Note that currently Suspense doesn’t work in SSR mode._
```
pacocoursey:
```suggestion
_Note that Suspense is not supported in SSR mode._
```

README.md
```diff
@@ -326,6 +338,28 @@ function Profile () {
 }
 ```
 
+### SSR with Next.js
+
+With the the `initialData` option, you pass an initial value to the hook. It works perfectly with many SSR solutions
```
sergiodxa:
```suggestion
With the `initialData` option, you pass an initial value to the hook. It works perfectly with many SSR solutions
```

pacocoursey: Classic! I miss the double 'the' every time.

PR #138, opened 2019-11-20
Clarify about HTTP RFC 5861

dancerphil:
Clarify that swr does not implement HTTP RFC 5861, but implements a cache invalidation strategy popularized by HTML RFC 5861, which may make reader confusing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
 
 SWR is a React Hooks library for remote data fetching.
 
-The name “**SWR**” is derived from `stale-while-revalidate`, a cache invalidation strategy popularized by [HTML RFC 5861](https://tools.ietf.org/html/rfc5861).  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,7 +21,7 @@
 
 SWR is a React Hooks library for remote data fetching.
 
-The name “**SWR**” is derived from `stale-while-revalidate`, a HTTP cache invalidation strategy popularized by [RFC 5861](https://tools.ietf.org/html/rfc5861).  
+The name “**SWR**” is derived from `stale-while-revalidate`, a cache invalidation strategy popularized by [HTML RFC 5861](https://tools.ietf.org/html/rfc5861).
```
quietshu:
```suggestion
The name “**SWR**” is derived from `stale-while-revalidate`, a cache invalidation strategy popularized by [HTTP RFC 5861](https://tools.ietf.org/html/rfc5861).  
```

Sounds good to me! But IMO this should be an HTTP RFC, not HTML : )

dancerphil: Good point.

Issue #143, opened 2019-11-21
Example for axios


Svish:
With `axios` it's handy to "define requests" using a request object, for example like this:

```ts
import axios, { AxiosRequestConfig } from 'axios';

const request: AxiosRequestConfig = {
  baseURL: 'https://api.example.com',
  withCredentials: true,
  method: 'GET',
  url: '/search',
  params: {
    term: 'foobar',
  },
};

const response = await axios(request);
const data = response.data;
```

But from what I can gather, the `key` I can pass in to `useSWR` has to be a string? I can't find any of your included examples that uses anything other than a simple string either. In the readme, there's a short mention of using arrays, which can include both strings and objects, but it doesn't mention what the fetcher would actually look like, and it also says the key comparison is shallow, so with my request object above here, I assume that means a change in `request.params.term` would not be picked up?

Would it be possible for someone to add an example of how one can use `useSWR` with axios and a request object like this? Or is it perhaps not possible to do that?

quietshu:
When using an array as the key, SWR will pass those items as arguments to the fetcher function:
```js
useSWR([A, B, C, ...], fetcher)
// calls fetcher(A, B, C, ...)
```

For your specific case, I'd do something like:
```js
// global config because this part won't change
const getRequest: AxiosRequestConfig = {
  baseURL: 'https://api.example.com',
  withCredentials: true,
  method: 'GET'
};

function App () {
  // inside component
  const params = useMemo(() => ({ term }), [term])
  useSWR(
    ['/search', params],
    (url, params) => axios({ ...getRequest, url, params }).then(res => res.data)
  )
}
```

or simpler:
```js
// global config because this part won't change
const getRequest: AxiosRequestConfig = {
  baseURL: 'https://api.example.com',
  withCredentials: true,
  method: 'GET'
};

const fetchTerm = (url, term) =>
  axios({ ...getRequest, url, params: { term } }).then(res => res.data)

function App () {
  // inside component
  useSWR(['/search', term], fetchTerm)
}
```

The major reason of shallow comparison is performance. So it will always render faster, and it will reuse the stale data if the same url + params pair has occurred before.

Thanks for the feedback! We will definitely improve the readme and create an example for axios.

Svish:
@quietshu Thank you for a quick reply!

[...]

PR #145, opened 2019-11-21
Axios examples

Svish:
I asked for some examples using axios in #143. Since the `useRequest` wrapper hook I made for my own project seems to work well, as far as my newb `useSWR` skills can gather, I decided to just copy the `basic` and `basic-typescript` examples, and change them to use this wrapper hook. 🙂

Hope I haven't missed anything crucial 🤔


Svish:
@sergiodxa Good catch. Deleted, and ignored. 🙂👍

quietshu:
Although `JSON.stringify` could make it work, I don't think we should recommend it.

`useRequest` is a nice abstraction :+1:

Svish:
@quietshu Could make it work? What would make it not work? Something I could fix? 🤔

examples/axios/pages/[user]/[repo].js
```diff
@@ -0,0 +1,39 @@
...
+  const { data } = useRequest(
+    id
+      ? {
+          url: '/api/data',
+          params: {
+            id
+          }
+        }
+      : null,
+    fetch
```
Comment (quietshu) [inline examples/axios/pages/[user]/[repo].js]: Is this `fetch` needed?

examples/axios-typescript/libs/useRequest.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export default function useRequest<Data = unknown, Error = unknown>(
+  request: GetRequest,
+  { initialData, ...config }: Config<Data, Error> = {}
+): Return<Data, Error> {
+  const { data: response, error, isValidating, revalidate } = useSWR<
+    AxiosResponse<Data>,
+    AxiosError<Error>
+  >(request && JSON.stringify(request), () => axios(request || {}), {
```
quietshu: I think we should add a comment here and mention it in the README (and the documentation in the future).
`JSON.stringify` is a hack and not the recommended way because:
1. it's unnecessary and slow (runs on every render)
2. it's not stable

2 means it can produce different keys for the same request. e.g.:
`JSON.stringify({ params: { a: 1, b: 2 } }) // -> {"params":{"a":1,"b":2}}`
`JSON.stringify({ params: { b: 2, a: 1 } }) // -> {"params":{"b":2,"a":1}}`

I think this will be better:
```js
useRequest(request, deps = [], config = {}) {
  useSWR(deps, () => axios(request || {}), ...)
  ...
}

// to use it:

useRequest({
  url: '/api/data',
  params: { id }
}, ['/api/data', id])

useRequest({
  url: '/api/data'
}, ['/api/data'])
```

Just like React's `useMemo`, `useEffect`, and `useCallback`.

But it's not an anti-pattern or "mistake" to use `JSON.stringify` here. I'm good with using it in this specific example and sometimes it is easier to use and working well for small apps indeed. That's why I think we need a comment in the example's code to clarify that.

examples/axios-typescript/libs/useRequest.ts
```diff
@@ -0,0 +1,52 @@
...
+  return useMemo(
+    () => ({
+      data: response && response.data,
+      response,
+      error,
+      isValidating,
+      revalidate
+    }),
+    [response, error, isValidating, revalidate]
+  )
```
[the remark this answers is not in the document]
Svish: Right, I wasn't quite sure, but removed it now 👍

examples/axios/pages/[user]/[repo].js
Comment (Svish) [reply examples/axios/pages/[user]/[repo].js]: Nope, and isn't even defined, so not sure why it didn't complain here 😕 Removed!

examples/axios-typescript/libs/useRequest.ts
Svish: I like your suggestion. I'll try to update it with that 🙂

Svish: @quietshu How does `useSWR` react if there's no key? Or if you pass it an empty array? Will it return the cached version of a different call that also was created with an empty array? Or will those arrays be different arrays, and if so, how would it be cached if the array next render "changes" to a new empty array? 🤔

quietshu: @svish SWR will "pause" the data fetching (not fetching anything and just return `undefined` as the data) if the key is an empty string, array or just null.

Svish: So it wouldn't really work then, because you'd need something more to use as a key. An empty deps array, in the case of a query with no parameter, would be ignored, and an array with for example an id of 2, would potentially be mixed up with a different query which also had an id of 2 (for example a post and a user with the same id). 😕

quietshu: Yeah that's why I also put the `url` in the array: `['/api/data', id]`. So it won't be mixed up.

Svish: Ooh, right, missed that. Thanks, and sorry for being a bother 😛

quietshu: No worries : )

Svish: @quietshu Hm, after looking at it more, I think it's easier to just use the `JSON.stringify`, because it's difficult to know exactly how much in the request object one should need to include in the array. And from what I can understand, `JSON.stringify` is fairly fast, and it should also be stable as long as the object is created the same way each time.

Would probably be good to add a comment about that though, as you mention, but not sure how it should be written. Maybe one of you guys could try to author something along the lines of what you think it should say, and I can add it to the code? 🤔

quietshu: @Svish No worries, I think it's fine for now.

pabacham: @Svish Hi. With a current approach, I stuck with an issue when I need to mutate some request but I don't know its params cause they are too high in a tree of components and there is no sense to put it down to children.

Svish: > @Svish Hi. With a current approach, I stuck with an issue when I need to mutate some request but I don't know its params cause they are too high in a tree of components and there is no sense to put it down to children.

Well, then you just need to do it differently then 😛 I'm currently only doing fetching with this library, so haven't done anything with mutate or such yet.

pabacham: Seems like there is a sense to pay attention to the "apollo-client" implementation of working with cache (__typename and etc) cc @quietshu

Issue #180, opened 2019-12-06
SWR requires a peer of react@16.11.0


Svish:
> swr@0.1.13 requires a peer of react@16.11.0 but none is installed.

Is it really necessary with such a specific peer dependency requirement as you have now?

```
  "peerDependencies": {
    "react": "16.11.0"
  }
```

The current React version is `16.12.0`, so on every `npm install` I now get warnings from `swr`. 😕

wendellhu95:
```diff
- "react": "16.11.0"
+ "react": "^16.11.0"
```

Svish:
@wendellhu95 That would be better, but still wouldn't accept `16.12.0`, would it?

Assuming react won't break anything until a new major version is released, and you do need at least version `16.11.0`, I guess it could be like this?

```
"react": ">= 16.11.0 < 17"
```

wendellhu95:
`^` can cover any upcoming minor versions. It's equivalent to `>= 16.11.0 < 17`.

Messed it up with `~`?

Svish:
@wendellhu95 Ah, then definitely `^` 👍

Just a newb who thought `^` was for the patch version and didn't even know `~` was a thing 😜

PR #181, opened 2019-12-06
Allow newer minor versions of React as peer dependency

Svish:
This PR opens up the peer dependency requirement to allow for newer minor releases of React without SWR starting to complain.


Svish:
Fixes #180

Issue #184, opened 2019-12-07
Performance: Component Re-rendered because of hook changes.


AnasBoulmane:
Make sure the state returned from the reducer is not a different object that equals by value but different by reference from the previous state:

![image](https://user-images.githubusercontent.com/17661697/70374958-27b02500-18f9-11ea-9097-52affdcc3d39.png)

quietshu:
I think it’s because of the change of `isValidating` (the loading state), which is also returned by the hook.

quietshu:
Will work on an improvement soon.

PR #185, opened 2019-12-07
Fix component Re-rendered. (#184)

AnasBoulmane:
* improve the mergedState reducer to return the previous state if it is equal to the new state by value

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -132,7 +132,8 @@ const mutate: mutateInterface = async (_key, _data, shouldRevalidate) => {
 }
 
 function mergeState(state, payload) {
-  return { ...state, ...payload }
+  const newState = { ...state, ...payload }
+  return deepEqual(newState, state) ? state : newState
 }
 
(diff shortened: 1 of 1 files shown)


quietshu:
I don't think this will help. Since there's a [`deepEqual` check](https://github.com/zeit/swr/blob/05779e6c892e9dac9007faa37044c3aa5fb8cd69/src/use-swr.ts#L297) before [dispatching the new state](https://github.com/zeit/swr/blob/05779e6c892e9dac9007faa37044c3aa5fb8cd69/src/use-swr.ts#L307). Have you tested it locally?

AnasBoulmane:
there isn't a deepEqual check before each dispatch of a new state
https://github.com/AnasBoulmane/swr/blob/716714791b0901c9a9c39dd705b0ac406b28bde2/src/use-swr.ts#L221
image
https://github.com/AnasBoulmane/swr/blob/716714791b0901c9a9c39dd705b0ac406b28bde2/src/use-swr.ts#L308
image

Issue #144, opened 2019-11-21
data is always undefined first, even when cached


herrstucki:
Whenever a key is used that has already cached data, `data` will be `undefined` anyway first and then immediately cause a re-render with the cached data.

I've set up an example here:
https://codesandbox.io/s/determined-almeida-3nwpv

You'll see that even for __cached__ results, the sequence will be like this:

```
CLICKED
data: undefined, isValidating: false <-- causes an unnecessary render with undefined data!
data: Hello Bar, isValidating: false
data: Hello Bar, isValidating: true
data: Hello Bar, isValidating: false
```
… where I would _expect_ a sequence like this:

```
CLICKED
data: Hello Bar, isValidating: false <-- Even this step could be optional?
data: Hello Bar, isValidating: true
data: Hello Bar, isValidating: false
```

sergiodxa:
This is because SWR uses useEffect to read from cache I believe, I know when you enable Suspense it will try to read from cache right away.

quietshu:
Amazing reproduction example! Thanks @herrstucki.

The major reason is that when changing the key, it's still the same hook. For example:
```js
const [state] = useState(key)
```

Only the initial `key` will affect the state. If the key has changed, it returns `undefined` (because of course it can't just return the previous state of the previous key), and uses `setState` to update it (inside `useEffect`), so the first render will be `undefined`, and set to the correct data in a second render.

However I do think this is fixable (https://github.com/zeit/swr/blob/master/src/use-swr.ts#L544-L547), by just trying to return the cached data (actually the `initialData` value, it also handles the hydration case), rather than `undefined`.

Will let you know when this gets implemented  👍

herrstucki:
Thanks @quietshu! 

I tried to figure out how things work by looking at the source but to be honest I didn't quite understand it at first glance 😅 

I'm not quite sure why a changed key even needs to update state in the first place … reading from cache is synchronous anyway, no? Maybe a simplified pseudo-code example of what I mean:

```ts
function useSWR(key, fetcher) {
  const [isValidating, setValidating] = useState(false);

  let currentData;
  if (cache.has(key)) {
    currentData = cache.get(key); // synchronous read
  }
  
  useEffect(() => {
    if (needsRevalidating(key)) {
       // will update cache, cache will notify "subscribers" via context or whatever
      triggerRevalidating(key)
      setValidating(true)
    }
  }, [key, setValidating])

  
  return { data: currentData, isValidating };
}
```

Issue #114, opened 2019-11-12
Rerender multiple times 


beidan:
when i use swr, i find it will rerender multiple times.

maybe **useSWR** use many **useState** ……
![image](https://user-images.githubusercontent.com/17068068/68670120-19374d80-0587-11ea-8320-4b3a45e688be.png)

but my application doesnot need so many states.
Will it affect performance? 
When not in use, whether  can remove **useState** or Stop updating the state ?

pacocoursey:
Yes, components using SWR will re-render multiple times to update the return values (data, error, isValidating).

Currently there is no way to avoid this, if you don't need all these extra options maybe SWR isn't best suited for your use case. May I ask what your use case is?

quietshu:
> maybe useSWR use many useState

It won’t cause multiple re-renders because it’s using batched updates. FYI: https://stackoverflow.com/questions/48563650/does-react-keep-the-order-for-state-updates/48610973#48610973

beidan:
There may be many requests for a page. But not every request requires loading . 
Can the loading state be customized ?

quietshu:
@beidan if you read this: https://github.com/zeit/swr/blob/master/src/use-swr.ts#L242-L263

Data, error and validating states are batched, there will be **only 1** re-render. Although we will change those `useState` to `useReducer` in the future, but that will not cause any difference. Maybe you see the re-rendering because of the revalidation.

> But not every request requires loading.

I think it's a basic requirement for a data-fetching library, to have the loading state, error state and the data.

beidan:
> @beidan if you read this: https://github.com/zeit/swr/blob/master/src/use-swr.ts#L242-L263
> 
> Data, error and validating states are batched, there will be **only 1** re-render. Although we will change those `useState` to `useReducer` in the future, but that will not cause any difference. Maybe you see the re-rendering because of the revalidation.
> 
> > But not every request requires loading.
> 
> I think it's a basic requirement for a data-fetching library, to have the loading state, error state and the data.

thanks ~ perfect

quietshu:
@beidan let me know if there's any other question! We're also going to ship #117 soon.

PR #186, opened 2019-12-08
Reduce re-renders by auto detecting state dependencies

### LINKED ISSUE #184: Performance: Component Re-rendered because of hook changes.

quietshu:
By moving _data, error, isValidating_ away from the state (`useReducer`) to a ref (`stateRef`), the hook won't trigger another re-render anymore when one of these values updates. 

Instead, we add getters to these values (implemented with `defineProperties`). Once a value is read, it will be marked as a dependency immediately. Thus, we can trigger a re-render if any dependency changes in the future.

With this change, `useSWR` will never cause unnecessary re-renders:

```js
const { data } = useSWR(...)
// will not re-render if `isValidating` or `error` changes
```

No changes needed from the user side (and Suspense mode is not affected).

Fixes #184. Related #144, #114.

Published at ~~`swr@0.1.13-beta.0`~~ `swr@0.1.15-beta.0`.

---

This PR has been running on our production for a couple of weeks, and it works very well  🎉

### DIFF
--- a/src/use-swr-pages.tsx
+++ b/src/use-swr-pages.tsx
@@ -150,7 +150,12 @@ export function useSWRPages<OffsetType = any, Data = any, Error = any>(
       ) {
         setPageSWRs(swrs => {
           const _swrs = [...swrs]
-          _swrs[id] = swr
+          _swrs[id] = {
+            data: swr.data,
+            error: swr.error,
+            revalidate: swr.revalidate,
+            isValidating: swr.isValidating
+          }
           return _swrs
         })
         if (typeof swr.data !== 'undefined') {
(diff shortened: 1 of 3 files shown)


quietshu:
Added more code optimization (memorize returned state and dispatch function), also fixed a bug causing SWRPages not updating in 5795f3325c669005ea67db19b2a6f6661f0f9da6.

Published a new version based on the current master to `swr@0.1.15-beta.0`.

quietshu:
Note: in the future we are going to simplify the useSWRPages hook (which is undocumented currently).

rubnvp:
maybe you could use Proxies (as Vue 3) instead of Object.defineProperty? (it won't support IE)

iamstarkov:
Seems like this PR breaks ssr when `initialData` is used. see #230

PR #203, opened 2019-12-16
Prevent polling when browser is offline

daneden:
I love `useSwr`! I’m using it to build a [London Underground status site](http://underground.lucid.toys/).

While testing the app on slow/no network speeds, I noticed that `useSwr` will continue polling endpoints even if the navigator is offline. This PR adds an additional boolean check in the `tick` function to ensure that requests aren’t sent if the browser is offline.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -476,7 +476,6 @@ function useSWR<Data = any, Error = any>(
     let reconnect = null
     if (config.revalidateOnReconnect) {
       const reconnectEvent = async () => {
-        console.log('reconnecting...')
         await softRevalidate()
       }
 
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -473,6 +473,7 @@ function useSWR<Data = any, Error = any>(
       timeout = setTimeout(tick, config.refreshInterval)
     }
 
+    // set up reconnecting when the browser regains network connection
     let reconnect = null
     if (config.revalidateOnReconnect) {
       const reconnectEvent = async () => {
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -476,11 +476,7 @@ function useSWR<Data = any, Error = any>(
     // set up reconnecting when the browser regains network connection
     let reconnect = null
     if (config.revalidateOnReconnect) {
-      const reconnectEvent = async () => {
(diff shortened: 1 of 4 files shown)


sergiodxa:
Maybe it will be great to detect when the user is back online to trigger a revalidation.

daneden:
I’m confused about why, but I’m getting an “Invalid hook call” error (as in [Rules of Hooks](https://reactjs.org/warnings/invalid-hook-call-warning.html)) when I test this locally. I’m not sure if it’s an issue with different React modules being loaded; I’m also struggling to get tests to run locally, but it looks like tests pass in Travis!

Let me know if you also run into problems.

quietshu:
Thanks a lot! Could you also create an option for it (same as `config.refreshWhenHidden`)? Because sometimes people use SWR for local states too so it can work in offline.

daneden:
@quietshu no problem! Added as a config option. I also took @sergiodxa’s suggestion and added an additional `revalidateOnReconnect` option that will add an event listener to the `online` event and trigger a fetch.

rauchg:
It just dawned on me we didn't have this capability. That's epic @daneden …

src/use-swr.ts
```diff
@@ -473,6 +473,16 @@ function useSWR<Data = any, Error = any>(
...
     }
 
+    // set up reconnecting when the browser regains network connection
+    let reconnect = null
+    if (config.revalidateOnReconnect) {
+      const reconnectEvent = async () => {
+        await softRevalidate()
+      }
+
+      reconnect = addEventListener('online', reconnectEvent)
```
```suggestion
      reconnect = addEventListener('online', softRevalidate)
```

And we can remove `reconnectEvent`.

quietshu: All other things looks good! Thank you @daneden ❤️

daneden: Ah nice, I wondered if it was necessary to wrap it in an async function. Will update now!

daneden: Ah nice! I wondered if I had to wrap this in an async function. Just pushed this change.

quietshu:
Thank you @daneden !

quietshu:
Published in `swr@0.1.15` 🎉

PR #207, opened 2019-12-18
Fix #182

### LINKED ISSUE #182: Dynamic polling interval
When changing `refreshInterval` dynamically:

```js
const [int, setInt] = useState(0)
const { data } = useSWR('/api', { refreshInterval: int })

return <>
  {data}
  <button onClick={() => setInt(int + 1000)}>update</button>
</>
```

Sometimes the received `data` will not be updated from the data source upon interval changes. Need to refine the interval change logic and remove `refreshInterval` from the `useEffect` deps, and use a ref for it.

Also we need a test to make sure it'll be stable in the future.

promer94:
close #182

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -134,6 +134,28 @@ function mergeState(state, payload) {
   return { ...state, ...payload }
 }
 
+function usePolling(callback: () => void, refreshInterval?: number | null) {
+  const callbackRef = useRef(callback)
+
+  useIsomorphicLayoutEffect(() => {
+    callbackRef.current = callback
+  }, [callback])
+
+  useIsomorphicLayoutEffect(() => {
+    let timer = null
+    function tick() {
+      callbackRef.current()
+      timer = setTimeout(tick, refreshInterval)
+    }
+    if (refreshInterval) {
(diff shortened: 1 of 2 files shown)


src/use-swr.ts
```diff
@@ -505,15 +506,27 @@ function useSWR<Data = any, Error = any>(
...
+      !errorRef.current &&
+      (config.refreshWhenHidden || isDocumentVisible()) &&
+      (!config.refreshWhenOffline && isOnline())
+    ) {
+      // only revalidate when the page is visible
+      // if API request errored, we stop polling in this round
+      // and let the error retry function handle it
+      await revalidate({ dedupe: true })
+    }
+  }
```
quietshu: `refresh` will be recreated in every render (and triggers the `useIsomorphicLayoutEffect` inside `usePolling`) so might not be efficient.

A suggestion is to add deps of `refresh` to `usePolling`:
```js
const refresh = ...

usePolling(refresh, config.refreshInterval, [config.refreshWhenHidden, config.refreshWhenOffline, revalidate]))
```

And inside `usePolling`, we can use the deps:

```js
    callbackRef.current = callback
  }, deps)
```

This way there're only 2 `useEffect`s being used.

PR #211, opened 2019-12-22
skip immediate revalidation when initalData is provided

### LINKED ISSUE #194: When using SSR and initialData, is it possible to prevent first validation after render?
In `getInitialProps()` I call my fetcher function which queries the db for data. I pass that initial data to `useSWR` as `initialData` option. However, as soon as it renders, it queries the server again with the same end point (returning the same data set). I would expect that it would not query on render if `initialData` is passed, or there would be an option to prevent this.

I would like to prevent the server getting hit twice in a short timeframe with the same endpoint.

addstar34:
This fixes #194.

I wasn't sure if this should be set via another option or not but if so let me know and I can update.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -382,16 +382,18 @@ function useSWR<Data = any, Error = any>(
     const softRevalidate = () => revalidate({ dedupe: true })
 
     // trigger a revalidation
-    if (
-      typeof latestKeyedData !== 'undefined' &&
-      !IS_SERVER &&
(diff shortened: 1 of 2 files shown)


iamstarkov:
@addstar34 do you think this PR could introduce a regression i described in #230 ?

andreoav:
After upgrading, if I set the `initialData` options, SWR does not send the request.
I'm not using with SSR. It is a CRA application.

Issue #213, opened 2019-12-26
Passing initial data when there's an error in the cache displays the error


addstar34:
If you render a component that errors and then unmount it and then remount it with initial data it will return the error from the cache not the initial data.

Here's an example component. I'm using next js and passing initial data via `getInitialProps`.

```
import React from 'react'
import useSWR from 'swr'

import fetcher from 'src/data/fetcher'

function Users({ initialData }) {
  const { data: users, error } = useSWR('/users', fetcher, {
    initialData,
  })

  return (
    <>
      <h1>Users</h1>
      {error ? (
        <div>{error.message}</div>
      ) : (
        <ul>
          {users &&
            users.data.map(i => <li key={i.id}>{i.name}</li>)}
        </ul>
      )}
    </>
  )
}

Instruments.getInitialProps = async function() {
  try {
    const initialData = await fetcher('/users')
    return { initialData }
  } catch (e) {
    return {}
  }
}

export default Users
```

PR #214, opened 2019-12-26
set initialError to undefined if intialData is provided

addstar34:
This addresses #213 where passing initialData when there's an error in the cache will still return the error.

I'm not sure if this is the correct way to fix this as it means the cache and state would be out of sync but maybe that's ok? It did pass all tests 🤷‍♂

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -187,7 +187,7 @@ function useSWR<Data = any, Error = any>(
   }
 
   const initialData = cacheGet(key) || config.initialData
-  const initialError = cacheGet(keyErr)
+  const initialError = config.initialData ? undefined : cacheGet(keyErr)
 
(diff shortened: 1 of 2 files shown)


quietshu:
I understand the issue you had, but I'm not sure about this fix.

The fundamental of useSWR is that each resource is defined by the key. The hook still considers it as the same request, even if `initialData` changes. And `initialData` is a **fallback** of the cached (current) data.

So the problem can be explained like this:
1. swr fetched resource and got an error, its state (error) was set
2. still the same swr, but its `initialData` changed

The thing that confused me was that in `getInitialProps` (SSR), if there's an error you still render the page but immediately request the same resource from client side (sounds like a retry right?). I think the recommend approach to do SSR is to just render an error page instead.

addstar34:
For my use case I was trying to still render the page but show an errorModal... I can work around this so I think this is fine to close.

Issue #189, opened 2019-12-09
Is there a way to reset pages of `useSWRPages`?


tem-tem:
Hi

I'm using swr for pagination in my local next.js project. My implementation is based on [this example](https://github.com/zeit/swr/blob/master/examples/pagination/pages/page-index.js).
The difference is that my component with `useSWRPages` is using dynamic route, which loads user's feed, and the route looks like this: `/user/[username]`.

The problem appears when I switch user feed with loaded pages. In more details:
1. I, as a user, open user-A's feed, then load a number of pages (let's say 4). 
2. I change feed to user-B's.
3. User-B's feed loads 4 pages, instead of 1.

I'd tried to pass username in to `useSWRPages`'s deps, and use it in key, but it didn't helped me. Is there any workarounds? Maybe I should pass something else in to the deps?

There's is an issue that seems like similar to mine: #126.

martrik:
I'm experiencing the same issue. On top of that, one loader is displayed for each page that's being loaded, which doesn't make a great UI.

tem-tem:
can `offset` variable cause any issues?

as far as I understand its value is equal to `null`, on first load:
https://github.com/zeit/swr/blob/f01de5a9d7de59e9075b02497355593a6b6cb6e2/src/use-swr-pages.tsx#L103-L105

but it also equals to `null` when there's no next page:
https://github.com/zeit/swr/blob/f01de5a9d7de59e9075b02497355593a6b6cb6e2/src/use-swr-pages.tsx#L130

https://github.com/zeit/swr/blob/master/examples/pagination/pages/page-index.js#L36

quietshu:
Currently you can try 
```js
mutate('_swr_page_count_' + pageKey, 0)
mutate('_swr_page_offset_' + pageKey, 0)
```

as a temporary work around.

I will work on simplifying the `useSWRPages` API next week since it's tricky to get it right (conditional and dependent hooks) and we have several bugs/feature requests around it. 🙏

tem-tem:
thank you @quietshu, but calling `mutate` didn't helped :(
however, i'm using [useSWR with axios](https://github.com/zeit/swr/blob/master/examples/axios-typescript/libs/useRequest.ts) with global fetch, so maybe I messed up somewhere.

[...]

PR #221, opened 2020-01-06
Moving page cache to use a cache map

aequasi:
Currently, changing the pageKey doesn't reset the pages cache, and theres no other way to clear the pages cache.

Might solve #189

### DIFF
--- a/src/use-swr-pages.tsx
+++ b/src/use-swr-pages.tsx
@@ -177,7 +177,10 @@ export function useSWRPages<OffsetType = any, Data = any, Error = any>(
 
     // render each page
     const p = []
-    const pageCache = pageCacheMap.get(pageKey) || []
+    if (!pageCacheMap.has(pageKey)) {
+      pageCacheMap.set(pageKey, [])
+    }
+    const pageCache = pageCacheMap.get(pageKey)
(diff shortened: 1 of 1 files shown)


src/use-swr-pages.tsx
```diff
@@ -176,7 +177,7 @@ export function useSWRPages<OffsetType = any, Data = any, Error = any>(
 
     // render each page
     const p = []
-    const pageCache = pageCacheRef.current
+    const pageCache = pageCacheMap.get(pageKey) || []
```
quietshu:
```suggestion
    if (!pageCacheMap.has(pageKey)) {
      pageCacheMap.set(pageKey, [])
    }
    const pageCache = pageCacheMap.get(pageKey)
```

src/use-swr-pages.tsx
```diff
@@ -196,6 +197,7 @@ export function useSWRPages<OffsetType = any, Data = any, Error = any>(
           pageFn: _pageFn,
           offset: pageOffsets[i]
         }
+        pageCacheMap.set(pageKey, pageCache)
```
quietshu:
```suggestion
```

aequasi: Didn’t realize that was by reference

PR #223, opened 2020-01-07
Fix check for addEventListener existence

robinandeer:
I noticed that React Native still crashes after applying the changes from #217.

More specifically the error I see is:

```
ReferenceError: Can't find variable: addEventListener
```

If you are seeing the same issue, this PR fixes that by checking if the variable `addEventListener` is defined as suggested here: https://stackoverflow.com/a/519157/2310187


quietshu:
Thank you @robinandeer ! Can you fix this for `removeEventListener` as well (line 504)?

robinandeer:
Sure I've updated the PR now :)

PR #225, opened 2020-01-10
revalidateOnMount flag that disable automatic revalidation on mount

ciak0:
Hi there,
we're currently using useSWR but we have found conditional revalidation a bit limiting.

For example we have some occurences where we want to trigger revalidation only after a certain action is being taken (could be a button click).
With current conditional revalidation techniques we are forced to use an additional state just to prevent first validation on mount.

This PR aims to add a simple autoLoad (or whatever other name) flag in the options so that automatic revalidation on mount can be prevented and we can simply trigger it directly in a callback (for example a button onClick).

Best,
Gabriele

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ const { data, error, isValidating, mutate } = useSWR(key, fetcher, options)
 - `suspense = false`: enable React Suspense mode [(details)](#suspense-mode)
 - `fetcher = undefined`: the default fetcher function
 - `initialData`: initial data to be returned (note: This is per-hook)
-- `revalidateInitialData = true`: automatically revalidate when component is mounted
+- `revalidateOnMount = true`: automatically revalidate when component is mounted
 - `revalidateOnFocus = true`: auto revalidate when window gets focused
 - `revalidateOnReconnect = true`: automatically revalidate when the browser regains a network connection (via `navigator.onLine`)
 - `refreshInterval = 0`: polling interval (disabled by default)
@@ -239,12 +239,12 @@ const { data } = useSWR(() => shouldFetch ? '/api/data' : null, fetcher)
 const { data } = useSWR(() => '/api/data?uid=' + user.id, fetcher)
 ```
 
-If you want to use a constant `key` and prevent initial revalidation you can simply set `revalidateInitialData` option to false.
+If you want to use a constant `key` and prevent initial revalidation you can simply set `revalidateOnMount` option to false.
 
 ```js
 // disable automatic revalidation on mount
 const { data } = useSWR('/api/data', fetcher, {
-  revalidateInitialData: false
+  revalidateOnMount: false
 })
 
 // ...later you can trigger revalidation in a callback or after another event
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ const { data, error, isValidating, mutate } = useSWR(key, fetcher, options)
 - `suspense = false`: enable React Suspense mode [(details)](#suspense-mode)
 - `fetcher = undefined`: the default fetcher function
 - `initialData`: initial data to be returned (note: This is per-hook)
-- `revalidateOnMount = true`: automatically revalidate when component is mounted
+- `revalidateOnMount`: enable or disable automatic revalidation when component is mounted (by default revalidation occurs on mount when initialData is not set, use this flag to force behavior)
 - `revalidateOnFocus = true`: auto revalidate when window gets focused
 - `revalidateOnReconnect = true`: automatically revalidate when the browser regains a network connection (via `navigator.onLine`)
 - `refreshInterval = 0`: polling interval (disabled by default)
--- a/README.md
+++ b/README.md
@@ -239,18 +239,6 @@ const { data } = useSWR(() => shouldFetch ? '/api/data' : null, fetcher)
 const { data } = useSWR(() => '/api/data?uid=' + user.id, fetcher)
 ```
 
-If you want to use a constant `key` and prevent initial revalidation you can simply set `revalidateOnMount` option to false.
-
-```js
-// disable automatic revalidation on mount
-const { data } = useSWR('/api/data', fetcher, {
-  revalidateOnMount: false
-})
-
-// ...later you can trigger revalidation in a callback or after another event
-const onClick = useCallback(() => trigger('/api/data'), [])
-```
-
 ### Dependent Fetching
 
 SWR also allows you to fetch data that depends on other data. It ensures the maximum possible parallelism (avoiding waterfalls), as well as serial fetching when a piece of dynamic data is required for the next data fetch to happen.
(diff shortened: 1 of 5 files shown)


src/use-swr.ts
```diff
@@ -394,7 +394,7 @@ function useSWR<Data = any, Error = any>(
     const softRevalidate = () => revalidate({ dedupe: true })
 
     // trigger a revalidation
-    if (!config.initialData) {
+    if (!config.initialData || !config.autoLoad) {
```
```suggestion
    if (!config.initialData && config.autoLoad) {
```

According to your use case, we should require `autoLoad` to be `true` to start revalidation at the beginning.

test/use-swr.test.tsx
```diff
@@ -80,6 +80,18 @@ describe('useSWR', () => {
     )
   })
 
+  it('should not call fetch function when autoLoad is false', async () => {
+    const fetch = jest.fn(() => 'SWR')
+
+    function Page() {
+      const { data } = useSWR('constant-3', fetch)
```
```suggestion
      const { data } = useSWR('auto-load', fetch, { autoLoad: false })
```

And for the test, the `autoLoad` option should be disabled (also use a new key here).

src/use-swr.ts
ciak0: yep sorry, I inverted condition

test/use-swr.test.tsx
ciak0: Ok great, will do

sergiodxa:
Could I propose rename from `autoLoad` to `revalidateInitialData`?

ciak0:
Ok, will do it!

ciak0:
> Could I propose rename from `autoLoad` to `revalidateInitialData`?

Just pushed option with new name and updated documentation README that I previously forgot.
Let me know what you think about it @sergiodxa @quietshu!

Cheers

README.md
```diff
@@ -109,6 +109,7 @@ const { data, error, isValidating, revalidate } = useSWR(key, fetcher, options)
 - `suspense = false`: enable React Suspense mode [(details)](#suspense-mode)
 - `fetcher = undefined`: the default fetcher function
 - `initialData`: initial data to be returned (note: This is per-hook)
+- `revalidateInitialData = true`: automatically revalidate when component is mounted
```
sergiodxa:
```suggestion
- `revalidateInitialData = true`: automatically revalidate when component is mounted (if `initialData` is provided)
```

src/use-swr.ts
```diff
@@ -394,7 +394,7 @@ function useSWR<Data = any, Error = any>(
     const softRevalidate = () => revalidate({ dedupe: true })
 
     // trigger a revalidation
-    if (!config.initialData) {
+    if (!config.initialData && config.revalidateInitialData) {
```
sergiodxa: Shouldn't this be or?
```suggestion
    if (!config.initialData || config.revalidateInitialData) {
```
If there is not initial data or want to revalidate the initial data, then trigger a revalidation

ciak0:
Uhmm...yes the purpose is to avoid revalidate on mount, ignoring
initialData, so I guess the latter name is better.

Regarding behavior:
if the flag is set to false I expect that revalidate doesn't occur at all
so that condition is probably an "or". I'll double check tomorrow!

Cheers

Il giorno mar 28 gen 2020 alle 22:57 Sergio Xalambrí <
notifications@github.com> ha scritto:

ciak0:
@sergiodxa @quietshu sorry for the delay, just updated PR with suggested hints and made it more clear. Let me know if you think is good or you have any other comment!

Best,
Gabriele

src/use-swr.ts
```diff
@@ -393,7 +393,7 @@ function useSWR<Data = any, Error = any>(
     const softRevalidate = () => revalidate({ dedupe: true })
 
     // trigger a revalidation
-    if (!config.initialData) {
+    if (!config.initialData && config.revalidateOnMount) {
```
ingoclaro, line 396: I think this should be an `||` so that if revalidateOnMount is set to true, it always revalidates regardless if there is initialData.
In that sense perhaps the default value of revalidateOnMount should be undefined, so that:
if value is undefined (ie not explicitly set) -> preserve existing behavior
if value is true -> always revalidate
if value is false -> don't revalidate

thoughts?

ciak0: Hi there,
if I understood well, current behavior is that whenever `initialData` is not set revalidation occurs (so that's why `true` is default value).

So if I change that condition with an `||`, I don't set `initialData` and I set `revalidateOnMount` to `false` that condition would be equal to:
`if (!undefined || false)` which is equal to `if (true || false)` and so it would trigger revalidation, ignoring the fact that `revalidateOnMount` flag is set to false.

ingoclaro: yeah, I didn't explain myself well enough, if you agree on the behavior for the value (undefined, true, false) then the solution isn't the || :)
Probably something like:
`if(config.revalidateOnMount === true || (config.revalidateOnMount === undefined && !config.initialData))`

ciak0: Ah gotcha, well...do you think is needed another value?
What's the downside of setting `true` as default value?

Anyway, yes, it's feasible.

ingoclaro: I just think that the variable name might be confusing, since it won't revalidate if initialData is provided, so from api usage perspective if I'm setting it as true explicitly (even with initialData), and it's not happening, it's weird, hence my suggestion to have it as undefined by default = current behavior.

ciak0:
Yeah you're right. I got your point!
Makes absolutely sense to me, I'm gonna do it asap!

Il giorno mar 7 apr 2020 alle 19:00 Ingo Claro <notifications@github.com>
ha scritto:

src/use-swr.ts
ciak0: Yeah you're right. I got your point!
I just pushed the changes with updated code, readme and tests.
Let me know what you think about it!

ciak0:
@shuding @pacocoursey @Timer any news? do I have to do something else?

sergiodxa:
@shuding is there something else @ciak0 needs to do to merge this?

shuding:
...although there’re some other things we need to figure out for `useMutation`, but that’s another topic. 

Anyway, I think generally it’s great for `useSWR` to have a lot of options and be very customizable, but it’s better to not expose them directly to the users and have some helper abstractions. An example is `$.post` ⊂ `$.ajax`.

sergiodxa:
👍 I think it makes sense, I was thinking that this would allow us to have a `useSWRLazy` which will only trigger the fetch after you run mutate, but could still read from cache if it was already cached.

ciak0:
Yeah great, makes sense!
I'll do it asap :)

Cheers,
Gab

Il giorno sab 9 mag 2020 alle 01:20 Sergio Xalambrí <
notifications@github.com> ha scritto:

sergiodxa:
@shuding I think this is ready to merge?

ciak0:
Thank you guys for this great library.
I've been using it intensively since day 1!

Glad to help.

Il giorno mar 12 mag 2020 alle 05:26 Shu Ding <notifications@github.com> ha
scritto:

whj1995:
revalidateOnMount set false, will prevent subsequent conditional fetching。 @ciak0

PR #231, opened 2020-01-14
Expose Cache

### LINKED ISSUE #158: Expose cache to users
I saw from #4 that there's a plan to expose SWR's (currently internal) cache to users at some point - is there some ETA on that?

My main use case right now would be to improve SSR with Next.js - [with the presented example code](https://github.com/zeit/swr#ssr-with-nextjs) (similar to code I have in my own Next.js app), `getInitialProps` will refetch the data *every time* before switching to the page, even if data already exists for that page in the SWR cache. Thus with exposed cache you could do something along the lines of this instead:

```typescript
import useSWR, { cacheGet, cacheSet } from "swr";

const path = "/api/data";

App.getInitialProps = async () => {
  let data;  
  if (typeof window === "undefined") {
    // server-side getInitialProps
    data = await fetcher(path); // always fetch on server
  } else {
    // client-side getInitialProps
    data = cacheGet(path); // check cache for data
    if (!data) {
      data = await fetcher(path); // only fetch if data not available in cache
      cacheSet(path, data); // update cache since we got the data for it
     }
  }
  return { data };
}

function App (props) {
  const initialData = props.data;
  const { data } = useSWR(path, fetcher, { initialData });

  return <div>{ data }</div>;
}
```

This way, you get those nice server side rendered pages while also making in-app navigation snappier for pages you've already visited.

### LINKED ISSUE #161: Have a way to clear cache
I'm writing tests for a component using `useSWR`, also with `{ suspense: true }`, and if I render a component reading the same key (e.g. the same component with different keys) I'm getting the old cache in my second test.

Not sure if this will be addressed with #158, but having a way to tell SWR to clear the cache of a certain key would be awesome for this use case.

```js
import { clearCache } from "swr";

clearCache(); // clear all cache
clearCache("/api/data"); // clear only a certain cache key
```

☝️ something like that.

What I have tried:

- `mutate(key, null, false)`
- `mutate(key, {}, false)`

Both doesn't work right since useSWR will still read that `null` or `{}` from the cache instead of triggering an immediate fetch and suspending the component.

I will try to create a CodeSandbox to replicate this soon.

### LINKED ISSUE #212: [RFC] Custom Cache Support
## Goal

Enable SWR users to control the cache.

This is based on the [Cache collection PR](https://github.com/zeit/swr/pull/92) and other issues such as #4, #16, #158, #161.

## Background

Right now there is no way to control the cache used by SWR, this means:

- We can't use another cache implementation (#16), maybe a project needs time based LRU, another could use size based LRU, another could use a completely different cache mechanism which could range from cache everything to cache nothing and request again every time, even use a SSR-friendly cache for Next.js usage or change where we keep the cached data (e.g. use localStorage to keep an offline cache)
- Data fetched on tests it's shared between tests (#161) and could not be easily cleared.
- There is not way to clear the cache or delete cache items individually (#4), this could be useful if we deleted a cached item and want to clear the cache, mutating to undefined will not work

## Proposal

I propose an API to support changing the cache through the SWRConfig component.

```js
const customConfig = {
  cache: CustomCacheImplementation
};

<SWRConfig value={customConfig}>
  <App />
</SWRConfig>;
```

With this, `App` and any component inside it calling `useSWR` will use this cache implementation.

The custom cache should only be customized with `SWRConfig` and not in a per `useSWR` instance, if you want to use a custom cache in a certain instance the parent should wrap the component in `SWRConfig`, this way the components are not aware of the cache mechanism used.

### Using multiple caches

This could also allow using multiple cache implementations in different parts of the app, you could render another `SWRConfig` deep inside the component tree of App to change the cache in that branch of the tree.

When this could be useful? Maybe we now certain parts should be cached offline (e.g with localStorage or IndexedDB) and others should be cached in-memory, we will be able to customize it.

### Default Cache

By default SWR will come with a cache implementation, e.g. it could use the one implemented in #92, this will allow most projects to don't care about the cache implementation.

### Testing

If we are testing components calling `useSWR` we could wrap the render of the component in `SWRConfig` with their own cache instance.

```js
// app.test.js
import React from "react";
import { SWRConfig } from "swr";
import { render, screen } from "@testing-library/react";
import { CacheInMemory } from "in-memory-cache-implementation";
import App from "./app";

test("should render with data", async () => {
  jest.once(JSON.stringify({ user: { username: "evilrabbit" } }));
  render(
    <SWRConfig value={{ cache: new CacheInMemory() }}>
      <App />
    </SWRConfig>
  );
  expect(await screen.findByText("evilrabbit")).toBeInTheDocument();
});
```

Now every test will use their on in-memory cache and will not share previously fetched data between them.

Using a new cache instance per test will simplify running tests in parallel without causing any issue because we clear the shared cache before another test finished first.

### Controlling the Cached Data

Since we could create the cache instance in their own module we should be able to import it anywhere and control it.

```js
// cache.js
class CustomCache {
  // implementation here
}

export default new CustomCache();
```

```js
// random-component.js
import cache from "./cache";

function RandomComponent() {
  // code here
  React.useEffect(() => {
    // clear cache key once the component unmounted
    return () => cache.del("cache-key");
  }, []);
  // code here
}
```

### Cache Interface

The cache should be an object, similar to a JS Map, including events to let instances of SWR subscribe to it. The exposed interface could be something like this:

```ts
type Key = string;
type Event = "set" | "delete";

interface Cache {
  set<Data>(key: Key, value: Data): Data; // save data to a key, create or update
  get<Data>(key: Key): Data | null; // return the data stored in the key or null
  delete(key: Key): void; // delete a single key
  has(key: Key): boolean; // check if a key is in cache
  clear(): void; // delete all key
  on(event: Event, callback: (key: Key) => void);
  off(event: Event, callback: (key: Key) => void);
}
```

Any custom cache should implement that interface, internally they could use any caching mechanism, e.g. a localStorage based cache which will store everything there forever.

```ts
// CacheEmitter could provide the on/off/emit methods
// it could be EventEmitter or something custom provided by SWR itself
class LocalStorageCache extends CacheEmitter implements Cache {
  set<Data>(key: Key, value: Data): Data {
    let stringifiedValue = value;
    if (typeof value !== "string") {
      stringifiedValue = JSON.stringify(value);
    }
    localStorage.set(key, value);
    this.emit("set", key);
  }

  get<Data>(key: Key): Data | null; {
    const stored = localStorage.getItem(key);
    if (!stored) return null;
    return JSON.parse(stored);
  }

  delete(key: Key): void {
    localStorage.removeItem(key);
    this.emit("delete", key);
  }

  has(key: Key): boolean {
    const value = this.get(key);
    return value !== null;
  }

  clear(): void {
    const keys: Key[] = Object.keys(localStorage);
    localStorage.clear();
    keys.forEach(key => this.emit("delete", key));
  }
}
```

### LINKED ISSUE #237: put initialData in cache
### description 

when populating a `useSWR` hook with initialData, that data is per-hook and not stored in the cache to be available to other hook invocations, resulting in duplicate fetches for `useSWR` invocations that don't have the SSR initialData passed to them.

for example:

0. in a next.js project a page's `getInitialProps` function is generating `initialData` as in the example.
1. the page component does `useSWR(someKey, { initialData })` and correctly SSR renders the initialData.
2. a child component also invokes `useSWR(someKey)`. the child component's `useSWR` sees an empty cache and executes fetcher, resulting in an unnecessary fetch.

### workarounds

1. prop-drill or context-pass initialData to all the components that need it
2. only use `useSWR` in the parent and prop-drill / context-pass the return value of `useSWR` to components that need it

### expectation

when initialData is provided, it will be stored in the cache under the provided key. when other `useSWR` hooks are invoked, they'll see the cached value and avoid an extra call to fetcher

sergiodxa:
This attempts to partially resolve #212 and resolves #161 and resolves #237 and resolves #158

## Implementation Notes

I updated this to only expose the global cache, I kept the cache interface, adding a way to tell the `set`, `delete` and `clear` methods if you want to notify components reading from the cache, this way `mutate` can call `cache.set` and `cache.set` can also call `mutate` as a notification method without running on a infinite loop (mutate calls `cache.set` with this flag as `false`).

I did this to come back to custom cache once I found a way to let `mutate` know which cache to call without manually passing it.

~I created a CacheInterface and make the global config create a default instance of the cache, also exported both the interface and the Cache class.~

~I also updated `useSWR` to read from the cache instance received through the context object.~

~I was not able, yet, to find a way to make `mutate` and `trigger` read from the correct cache instance, because they are global my only idea so far was to pass the cache instance as a third argument (probably optional), but that could make it hard to test components using a custom cache for tests and a default cache for the mutate call.~

## Missing things
~- `useSWRPages` is still using global cache (I haven't tried yet to use it here but I believe is the same as `mutate` and `trigger`)~
~- `mutate` is still using global cache~
~- `trigger` is still using global cache~

Any feedback is welcomed!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -9,4 +9,5 @@ export {
   responseInterface,
   CacheInterface
 } from './types'
+export { default as Cache } from './cache'
 export default useSWR
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,7 @@
 export * from './use-swr'
 import { default as useSWR } from './use-swr'
 export { useSWRPages } from './use-swr-pages'
+export { cache } from './config'
 export {
   ConfigInterface,
   revalidateType,
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,6 +6,7 @@ export {
   revalidateType,
   RevalidateOptionInterface,
   keyInterface,
-  responseInterface
+  responseInterface,
+  CacheInterface
 } from './types'
 export default useSWR
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,5 +10,4 @@ export {
   responseInterface,
   CacheInterface
 } from './types'
-export { default as Cache } from './cache'
 export default useSWR
--- a/test/use-swr.test.tsx
+++ b/test/use-swr.test.tsx
@@ -464,6 +464,47 @@ describe('useSWR - refresh', () => {
     })
     expect(container.firstChild.textContent).toMatchInlineSnapshot(`"count: 5"`)
   })
+
+  it('should allow use custom isEqual method', async () => {
+    function Page() {
+      const { data, revalidate } = useSWR(
+        'dynamic-11',
+        () => ({
+          timestamp: +new Date(),
+          version: '1.0'
+        }),
+        {
+          compare: function isEqual(a, b) {
+            if (a === b) {
+              return true
+            }
+            if (!a || !b) {
+              return false
+            }
+            return a.version === b.version
+          }
+        }
+      )
+
+      if (!data) {
+        return <div>loading</div>
+      }
+      return <button onClick={revalidate}>{data.timestamp}</button>
+    }
+
+    const { container } = render(<Page />)
+
+    expect(container.firstChild.textContent).toMatchInlineSnapshot(`"loading"`)
+    await waitForDomChange({ container })
+    const firstContent = container.firstChild.textContent
+    await act(() => {
+      // trigger revalidation
+      fireEvent.click(container.firstElementChild)
+      return new Promise(res => setTimeout(res, 1))
+    })
+    const secondContent = container.firstChild.textContent
+    expect(firstContent).toEqual(secondContent)
+  })
 })
 
 describe('useSWR - revalidate', () => {
--- a/test/use-swr.test.tsx
+++ b/test/use-swr.test.tsx
@@ -305,7 +305,7 @@ describe('useSWR', () => {
       </div>
     `)
 
-    // content should come from fetcher due lack of cached data
+    // content should be updated with fetcher results
     expect(await findByText('random message')).toMatchInlineSnapshot(`
       <div>
         random message
--- a/test/use-swr.test.tsx
+++ b/test/use-swr.test.tsx
@@ -7,7 +7,8 @@ import {
 } from '@testing-library/react'
 import React, { ReactNode, Suspense, useEffect, useState } from 'react'
 
-import useSWR, { mutate, SWRConfig, trigger, cache, Cache } from '../src'
+import useSWR, { mutate, SWRConfig, trigger, cache } from '../src'
+import Cache from '../src/cache'
 
 class ErrorBoundary extends React.Component<{ fallback: ReactNode }> {
   state = { hasError: false }
--- a/test/use-swr.test.tsx
+++ b/test/use-swr.test.tsx
@@ -7,7 +7,7 @@ import {
 } from '@testing-library/react'
 import React, { ReactNode, Suspense, useEffect, useState } from 'react'
 
-import useSWR, { mutate, SWRConfig, trigger, Cache } from '../src'
+import useSWR, { mutate, SWRConfig, trigger, cache } from '../src'
 
 class ErrorBoundary extends React.Component<{ fallback: ReactNode }> {
   state = { hasError: false }
@@ -280,6 +280,56 @@ describe('useSWR', () => {
       `"hello, SWR"`
     )
   })
+
+  it('should react to direct cache updates', async () => {
+    cache.set('custom-cache-key', 'custom cache message')
+
+    function Page() {
+      const { data } = useSWR('custom-cache-key', () => 'random message', {
+        suspense: true
+      })
+      return <div>{data}</div>
+    }
+
+    // render using custom cache
+    const { queryByText, findByText } = render(
+      <React.Suspense fallback={null}>
+        <Page />
+      </React.Suspense>
+    )
+
+    // content should come from custom cache
+    expect(queryByText('custom cache message')).toMatchInlineSnapshot(`
+      <div>
+        custom cache message
+      </div>
+    `)
+
+    // content should come from fetcher due lack of cached data
+    expect(await findByText('random message')).toMatchInlineSnapshot(`
+      <div>
+        random message
+      </div>
+    `)
+
+    act(() => cache.set('custom-cache-key', 'a different message'))
+
+    // content should be updated from new cache value
+    expect(await findByText('a different message')).toMatchInlineSnapshot(`
+      <div>
+        a different message
+      </div>
+    `)
+
+    act(() => cache.delete('custom-cache-key'))
+
+    // content should go back to be the fetched value
+    expect(await findByText('random message')).toMatchInlineSnapshot(`
+      <div>
+        random message
+      </div>
+    `)
+  })
 })
 
 describe('useSWR - loading', () => {
@@ -355,47 +405,6 @@ describe('useSWR - loading', () => {
     // it doesn't re-render, but fetch was triggered
     expect(renderCount).toEqual(1)
     expect(dataLoaded).toEqual(true)
-  })
-
-  it('should use custom cache', async () => {
-    const cache = new Cache({
-      'custom-cache-1': 'custom cache message'
-    })
-
-    function Page() {
-      const { data } = useSWR('custom-cache-1', () => 'random message', {
-        suspense: true
-      })
-      return <div>{data}</div>
-    }
-
-    // render using custom cache
-    const { queryByText, findByText, rerender } = render(
-      <SWRConfig value={{ cache }}>
-        <Page />
-      </SWRConfig>
-    )
-
-    // content should come from custom cache
-    expect(queryByText('custom cache message')).toMatchInlineSnapshot(`
-      <div>
-        custom cache message
-      </div>
-    `)
-
-    // render againt with default cache
-    rerender(
-      <React.Suspense fallback={null}>
-        <Page />
-      </React.Suspense>
-    )
-
-    // content should come from fetcher due lack of cached data
-    expect(await findByText('random message')).toMatchInlineSnapshot(`
-      <div>
-        random message
-      </div>
-    `)
   })
 })
 
--- a/test/use-swr.test.tsx
+++ b/test/use-swr.test.tsx
@@ -7,7 +7,7 @@ import {
 } from '@testing-library/react'
 import React, { ReactNode, Suspense, useEffect, useState } from 'react'
 
-import useSWR, { mutate, SWRConfig, trigger, cache } from '../src'
+import useSWR, { mutate, SWRConfig, trigger, cache, Cache } from '../src'
 
 class ErrorBoundary extends React.Component<{ fallback: ReactNode }> {
   state = { hasError: false }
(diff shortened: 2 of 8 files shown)


sergiodxa:
@quietshu I updated this PR to only expose the cache without letting the user replace it, let me know if you want me to change something, I will try again to allow to replace the cache once I think a good way to let mutate identify the correct cache.

quietshu:
Thank you @sergiodxa so much for working on this! I read the code briefly and it looks great and very clean 👍 And yeah indeed it will be a tricky problem for global mutation, as well as namespaced cache (if we're gonna do that). I will review the code again carefully and think about that too.

gabrieledarrigo:
Any news on this? 
It can be an unblocker for https://github.com/zeit/swr/issues/161 that's quite important in regards to testing.

mirshko:
Also curious, 

looking to hopefully have it fix #237

Daiz:
This should also resolve #158

andreoav:
This PR solves some problems we have with SWR and testing.
Currently we need to skip all tests for components that uses SWR.

embeddedt:
Having a way to clear cache is also useful for when one component does an API request and you need another component to update what its displaying and not show its previous data (without having to use ugly callbacks and refs).

gabrieledarrigo:
Ciao guys, any update on this?

sergiodxa:
@gabrieledarrigo I’m waiting for Shu to review it

nulladdict:
Hi, great job on this PR
However, I feel like we also need to expose `getErrorKey` and `getKeyArgs`, otherwise we wouldn't know what key to use

sergiodxa:
@nulladdict you are right, I move them to the Cache class as a single `serializeKey` method

quietshu:
@sergiodxa seems we have some conflicts now  :P 

Would love to land this in the next release!

sergiodxa:
@quietshu I fixed the conflicts and also added a `keys` method to get all cache keys as an array and a `subscribe` method to know when something changed in the cache. This method will be useful to integrate with SWR from the outside (e.g. subscribe to cache changes to sync with localStorage, then use localStorage to hydrate the cache before the first render)

test/use-swr.test.tsx
```diff
@@ -1251,3 +1250,72 @@ describe('useSWR - suspense', () => {
...
+    const listener = jest.fn()
+    const unsubscribe = tmpCache.subscribe(listener);
+    tmpCache.set("cache-2", "random message");
+
+    expect(listener).toHaveBeenCalled()
+
+    unsubscribe();
+    tmpCache.set("cache-2", "a different message");
+
+    expect(listener).toHaveBeenCalled()
```
```suggestion
    expect(listener).toHaveBeenCalledTimes(1)
```

quietshu:
BTW, @sergiodxa do you think it's a good idea to serialize the key in `set`, `get`, `has` and `delete`, if the key is not a string? So users don't need to call `serializeKey`, and all these behaviors can be unified:

```js
useSWR(['api/user', user])
mutate(['api/user', user])
cache.get(['api/user', user]) // instead of cache.get(cache.serializeKey(['api/user', user]))
```

src/index.ts
```diff
@@ -1,11 +1,14 @@
...
 export {
   ConfigInterface,
   revalidateType,
   RevalidateOptionInterface,
   keyInterface,
-  responseInterface
+  responseInterface,
+  CacheInterface
 } from './types'
+export { default as Cache } from './cache'
```
quietshu: Just thought more about this. It's still unclear to me why do we need to export the class `Cache`. Seems we can have support for multilayer cache, async cache and namespaced cache on top of the singleton `cache`.

sergiodxa: I did it because the original implementation let you create your own instance, right now it doesn't make any sense, I remove it.

sergiodxa: Done in https://github.com/zeit/swr/pull/231/commits/92d6ec748bbdab088c462a54455ba379437c1a29

sergiodxa:
@quietshu you are right, it will make more sense, and it will still work if the user pass the serialized string directly, and applied it in https://github.com/zeit/swr/pull/231/commits/b4ed2890828b44740d9f8e9ebb52b0ef5bd3699c

quietshu:
Released `swr@0.2.0-beta.0`, will fully test in prod before marking it as stable (v0.2.0)  🙏

sergiodxa:
Makes sense, I will try it in production too

embeddedt:
I can also test this with the use case I described [above](https://github.com/zeit/swr/pull/231#issuecomment-584156199). If I don't reply, assume everything worked fine. :slightly_smiling_face:

embeddedt:
I tested the feature with my application and it has solved the problem I was having before. Thanks!

nulladdict:
I've also tested this in prod, and it works pretty fine for us.
One small thing is that for my use case (clearing out *all* errors in suspense mode when retry button in my ErrorBoundary is clicked) I had to finesse the error prefix from cache.
I've used something like this:
```ts
const [_key, , _errorKey] = cache.serializeKey('key')
const errorKeyPrefix = _errorKey.replace(_key, '')
```
So I think it might've also been useful to expose the hardcoded `err@` prefix

gabrieledarrigo:
Ciao guys, big up for your work on this!
But, question, does it work with testing?
I'm trying `cache.clear` in my afterEach block, but it seems that it doesn't work:

```jsx
import { cache } from 'swr';

afterEach(() => {
  cache.clear()
});

it('should fetch on mount', async () => {
  await act(async () => {
    <Foo id="1" />
  });

  expect(fetch).toHaveBeenCalledWith('1');
});

it('should show an error when fetch fails', async () => {
  (fetch as jest.Mock).mockRejectedValue(new Error('An error occured'));

  await act(async () => {
    mount(
      <Foo id="1" />
    );
  });

  expect(fetch).toHaveBeenCalledWith('1');
  // other assertions that the error message is rendered
});
```

But the second test doesn't pass because `fetch` is never called; What am I missing?

andreoav:
@gabrieledarrigo I have the same problem, even with `cache.clear`, swr returns the result from the previous test.

@sergiodxa @quietshu Do we need to do something special to clear the cache during testing?

sergiodxa:
@andreoav @gabrieledarrigo I tried it in a test and found the issue, is because the deduping feature of SWR, since not enough time has passed since the last request SWR is reusing the fetch, you can fix it in two ways:

1. Add `dedupingInterval: 0` to the configs of the useSWR call
2. Wrap the component to test in SWRConfig and set `dedupingInterval: 0` there

I recommend to use the second option so your component will still have deduping of requests enabled while running inside your app and you can disable it only for the test.

I will send a PR with an example soon.

Using the example code above something like this should work:

```js
import { SWRConfig, cache } from 'swr';

afterEach(() => {
  cache.clear()
});

it('should fetch on mount', async () => {
  await act(async () => {
    <SWRConfig value={{ dedupingInterval: 0 }}>
      <Foo id="1" />
    </SWRConfig>
  });

  expect(fetch).toHaveBeenCalledWith('1');
});

it('should show an error when fetch fails', async () => {
  (fetch as jest.Mock).mockRejectedValue(new Error('An error occured'));

  await act(async () => {
    mount(
      <SWRConfig value={{ dedupingInterval: 0 }}>
        <Foo id="1" />
      </SWRConfig>
    );
  });

  expect(fetch).toHaveBeenCalledWith('1');
  // other assertions that the error message is rendered
});
```

andreoav:
@sergiodxa Nice!! I updated my tests and it worked.

I think we should add this to the documentation.

amannn:
Thanks a lot for your work on this, this is really helpful!

As far as I can tell, this hasn't been released to npm yet, right? Any chance to get a release? Thanks!

embeddedt:
@amannn It was released as a beta: https://github.com/zeit/swr/pull/231#issuecomment-590537273

amannn:
Ah right, thanks!

BrunoQuaresma:
> ```js
> value={{ dedupingInterval: 0 }}
> ```

Thanks so much to share this. I think would be nice to have this kind of info on docs.

p19ky:
So is there a way I can access the cache variable? I want to fetch only once and I want a condition like !cache.has(dataKey) or similar

christiaan:
> So is there a way I can access the cache variable? I want to fetch only once and I want a condition like !cache.has(dataKey) or similar

I use this way of calling swr to only have it fetch once.

```js
  return useSWR<string>(API_PICTURES_IMAGE(pictureId), {
    // Images cannot be changed so should never be revalidated
    revalidateOnFocus: false,
    revalidateOnMount: false,
    revalidateOnReconnect: false,
  });
```

p19ky:
> > So is there a way I can access the cache variable? I want to fetch only once and I want a condition like !cache.has(dataKey) or similar
> 
> @p19ky I use this way of calling swr to only have it fetch once.
> 
> ```js
>   return useSWR<string>(API_PICTURES_IMAGE(pictureId), {
>     // Images cannot be changed so should never be revalidated
>     revalidateOnFocus: false,
>     revalidateOnMount: false,
>     revalidateOnReconnect: false,
>   });
> ```

When I tried it this way, it does not load once. I does not load at all. So what I did to resolve this is the following:

```
import { cache } from "swr"; //imported cache that is used by SWR

const {
    data: categories,
    error: errorCategories,
    size: sizeCategories,
    setSize: setSizeCategories,
    isValidating: isValidatingCategories,
  } = useSWRInfinite(getKeyCategories, categoriesFetcher, {
    revalidateOnFocus: false,
    revalidateOnMount: !cache.has(() => getKeyCategories()), //only revalidate if there is nothing in cache for my categories Key
  });
```

pke:
I am running into this again and again. Why isn't `clearCache` not really clearing the deduping cache too?
Wrapping ever single test in `SWRConfig` is really a chore and makes the test less readable. 

Solutions:

1. export a function to disable deduping globally by default
2. enhance clearCache to also clear the deduping cache

PR #243, opened 2020-01-24
adding resetPages to clear cached pages on new searches 

patrickus:
Adding a Reset for pages 

reference:
https://github.com/gabrielperales/swr/commit/8ba8379d2e6c580ded234385d34ea829be996788

resolving: https://github.com/zeit/swr/issues/189

thanks https://github.com/gabrielperales

### DIFF
--- a/src/use-swr-pages.tsx
+++ b/src/use-swr-pages.tsx
@@ -109,7 +109,10 @@ export function useSWRPages(pageKey, pageFn, SWRToOffset, deps = []) {
       cacheSet(pageOffsetKey, [null]);
       return [null];
     });
-  }, []);
+  }, [pageOffsetKey, pageCountKey]);
+
+  // Reset page if the key changes
+  useEffect(() => resetPages(), [pageOffsetKey])
   // Doesn't have a next page
   const isReachingEnd = pageOffsets[pageCount] === null;
   const isLoadingMore = pageCount > pageOffsets.length;
--- a/src/use-swr-pages.tsx
+++ b/src/use-swr-pages.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useCallback, useMemo, useState, useRef } from 'react';
+import React, { useEffect, useCallback, useMemo, useState, useRef } from 'react'
 import {
   pagesResponseInterface,
   responseInterface,
@@ -6,7 +6,7 @@ import {
   pageOffsetMapperType
 } from './types'
 
-import { cacheGet, cacheSet } from './config';
+import { cacheGet, cacheSet } from './config'
 /*
 The idea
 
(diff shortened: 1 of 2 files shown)


quietshu:
Published to `swr@0.1.17-beta.1`!

quietshu:
Published `swr@0.1.17-beta.4`

innocentiv:
I am faced with issues with pagination this pull request is trying to solve, but I do not like that this pull request changes the function type signature to remove types altogether. This is a breaking change that blocks type inference for the arguments passed to the useSWRPages hook. Could we try to restore types before merging this pull request?

gabrielperales:
@patrickus looks like you have configured eslint or prettier to add semicolons to your code and is changing the code style this repo has. It won't be merged with those changes in the code style.

gabrielperales:
I need this feature to be merged. If @patrickus doesn't want to keep working on it, I can take over this branch and finish the work if the SWR team thinks that this is the right approach :)

innocentiv:
Hi, I had the same problem as @gabrielperales so I worked on a fix on my account.
Really sorry for being impolite by taking over this issue. I have created a second PR (https://github.com/zeit/swr/pull/269), but If @patrickus want to work on this, I will close my PR.

src/use-swr-pages.tsx
```diff
@@ -1,13 +1,12 @@
...
-import { cacheGet, cacheSet } from './config'
+import React, { useEffect, useCallback, useMemo, useState, useRef } from 'react'
 import {
   pagesResponseInterface,
   responseInterface,
   pageComponentType,
   pageOffsetMapperType
 } from './types'
 
+import { cacheGet, cacheSet } from './config'
```
sergiodxa: This should be
```suggestion
import { cache } from './config'
```
Then use cache.get and cache.set instead

patrickus: thanks, hadn't pulled the latest changes, will integrate

src/use-swr-pages.tsx
```diff
@@ -87,68 +87,64 @@ function App () {
...
   const isEmpty = isReachingEnd && pageCount === 1 && emptyPageRef.current
   const loadMore = useCallback(() => {
-    if (isLoadingMore || isReachingEnd) return
+    if (isLoadingMore || isReachingEnd)
+      return
     setPageCount(c => {
       cacheSet(pageCountKey, c + 1)
       return c + 1
     })
   }, [isLoadingMore || isReachingEnd])
```
gabrielperales: I think we need to add `pageCountKey` as @innocentiv has done here https://github.com/zeit/swr/pull/269/files#diff-c9d2eb3ea821da85bd697b22fe31edf4R153.

src/use-swr-pages.tsx
```diff
@@ -87,68 +87,64 @@ function App () {
...
+    setPageOffsets(() => {
+      cacheSet(pageOffsetKey, [null])
+      return [null]
+    })
+  }, [pageOffsetKey, pageCountKey])
 
+  // Reset page if the key changes
+  useEffect(() => resetPages(), [pageOffsetKey])
   // Doesn't have a next page
   const isReachingEnd = pageOffsets[pageCount] === null
```
Jack-Works, line 129: Should it be `pageOffsets[Math.max(1, pageOffsets.length - 1)] === null` ?

shuding:
I think we can close this due to the new useSWRInfinite API: https://github.com/vercel/swr/pull/435. It should have solved this problem (and other related ones). As always, thank you for your awesome contribution! 🙏

Issue #136, opened 2019-11-19
Return a tailored `mutate` function with the key already applied


pelotom:
Hi, great library!

It would be a nice convenience if, similar to `revalidate`, a `mutate` function were returned from `useSWR` which already had the key applied, e.g.

```diff
-import useSWR, { mutate } from 'swr'
+import useSWR from 'swr'

function Profile () {
- const { data } = useSWR('/api/user', fetcher)
+ const { data, mutate } = useSWR('/api/user', fetcher)

  return (
    <div>
      <h1>My name is {data.name}.</h1>
      <button onClick={async () => {
        const newName = data.name.toUpperCase()
        // send a request to the API to update the data
        await requestUpdateUsername(newName)
        // update the local data immediately and revalidate (refetch)
-       mutate('/api/user', { ...data, name: newName })
+       mutate({ ...data, name: newName }) // look ma, no key!
      }}>Uppercase my name!</button>
    </div>
  )
}
```

quietshu:
I think it will be great to merge `revalidate` and `mutate`:

`revalidate(newDataOrPromise, shouldRevalidateAfterMutation = true)`

```js
const { revalidate } = useSWR('/api/user', fetcher)

// ...
revalidate()
revalidate(newData)
revalidate(newData, false)
revalidate(post)
revalidate(post, false)
```

quietshu:
Adding support in #245.

PR #245, opened 2020-01-27
Improve mutate function

### LINKED ISSUE #253: Accessing cached data outside hooks.
I think its crucial to access cached data, at least while mutating.

lets say I have 2 endpoints - `api/users` & `api/user/<id>`

say i update `api/user/<id>`, i do not want to invalidate `api/users`, I would just like to mutate it, something like this - `mutate('api/users', [...usersWithoutModifiedUser, modifiedUser])`

the problem is.. I can't get the cached value for `api/users` at the time of the mutation.. since I only have the single user data I'm modifying.

I was thinking mutation api should look something like this

```
mutate('api/users', (cachedData) => {
  ... // using the cached data to build the new cached data.
   return [...usersWithoutModifiedUser, modifiedUser];
})
```

Hope you understand my issue, looking forward hearing from you,
Thank you for the great lib.

sergiodxa:
- If no data call trigger against the key
- If data is a function call it passing current cached value
- Return data or throw error after function
- SWR returns bound mutate function with the `key` already defined

Resolves https://github.com/zeit/swr/issues/155
Resolves #253

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ It features:
 - Fast page navigation
 - Revalidation on focus
 - Interval polling
+- Request deduplication
 - Local mutation
 - Pagination
 - TypeScript ready
--- a/README.md
+++ b/README.md
@@ -141,6 +141,7 @@ You can also use [global configuration](#global-configuration) to provide defaul
 - [Multiple Arguments](#multiple-arguments)
 - [Manually Revalidate](#manually-revalidate)
 - [Mutation and Post Request](#mutation-and-post-request)
+- [Mutate Based on Current Data](#mutate-based-on-current-data)
 - [SSR with Next.js](#ssr-with-nextjs)
 - [Suspense Mode](#suspense-mode)
 - [Error Retries](#error-retries)
@@ -354,6 +355,19 @@ Here’s an example showing the “local mutate - request - update” usage:
 mutate('/api/user', newUser, false)      // use `false` to mutate without revalidation
 mutate('/api/user', updateUser(newUser)) // `updateUser` is a Promise of the request,
                                          // which returns the updated document
+### Mutate Based on Current Data
+
+In many cases, you are receiving a single value back from your API and want to update a list of them.
+
+With `mutate`, you can pass an async function which will receive the current cached value, if any, and let you return an updated document.
+
+```js
+mutate('/api/users', async users => {
+  const user = await fetcher('/api/users/1')
+  return [user, ...users.slice(1)]
+})
+```
+
 ```
 
 ### SSR with Next.js
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ npm install swr
 ### API
 
 ```js
-const { data, error, isValidating, revalidate } = useSWR(key, fetcher, options)
+const { data, error, isValidating, mutate } = useSWR(key, fetcher, options)
 ```
 
 #### Parameters
@@ -103,7 +103,6 @@ const { data, error, isValidating, revalidate } = useSWR(key, fetcher, options)
 - `data`: data for the given key resolved by `fetcher` (or undefined if not loaded)  
 - `error`: error thrown by `fetcher` (or undefined)  
 - `isValidating`: if there's a request or revalidation loading  
-- `revalidate`: function to trigger the validation manually
 - `mutate`: function to mutate the cached data
 
 #### Options
@@ -357,6 +356,8 @@ Here’s an example showing the “local mutate - request - update” usage:
 mutate('/api/user', newUser, false)      // use `false` to mutate without revalidation
 mutate('/api/user', updateUser(newUser)) // `updateUser` is a Promise of the request,
                                          // which returns the updated document
+```
+
 ### Mutate Based on Current Data
 
 In many cases, you are receiving a single value back from your API and want to update a list of them.
--- a/README.md
+++ b/README.md
@@ -289,13 +289,13 @@ Dan Abramov explains dependencies very well in [this blog post](https://overreac
 ### Manually Revalidate
 
 You can broadcast a revalidation message globally to all SWRs with the same key by calling
-`trigger(key)`.
+`mutate(key)`.
 
 This example shows how to automatically refetch the login info (e.g.: inside `<Profile/>`) 
 when the user clicks the “Logout” button.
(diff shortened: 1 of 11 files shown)


jameslouiz:
This is exactly what I'm looking for. Great work!

omerman:
> This is exactly what I'm looking for. Great work!
> 
> In the mean time, for a bounded Mutate with ability to provide a function with access to the old data you can do this:
> 
> ```
> const useSWRWithBoundMutate = (key, fetcher, options) => {
>   const { data, error } = useSWR(key, fetcher, options);
>   const boundMutate = fn => mutate(key, fn(data));
>   return { data, error, boundMutate };
> };
> ```

It also gives you freedom of accessing the mutate from wherever you want, not only via react's scope( useSWR hook), because in many cases you would want to update just a part of a given response, and it is impossible to perform it outside react's scope with the current api.

quietshu:
Also as I commented in #262, I think we should remove the `shouldRevalidate` trick in https://github.com/zeit/swr/blob/edcfd265d3ad22901b203ecdb25a1fdc92254bf2/src/use-swr.ts#L112-L116 to avoid misunderstanding.

sergiodxa:
@quietshu I updated the README and removed the `shouldRevalidate`, always default to `true` now. Also updated the Optimistic UI example.

test/use-swr.test.tsx
```diff
@@ -847,7 +847,8 @@ describe('useSWR - local mutation', () => {
       // mutate and revalidate
```
omerman: should be just `// mutate` :P

README.md
```diff
@@ -354,6 +357,31 @@ Here’s an example showing the “local mutate - request - update” usage:
 mutate('/api/user', newUser, false)      // use `false` to mutate without revalidation
 mutate('/api/user', updateUser(newUser)) // `updateUser` is a Promise of the request,
                                          // which returns the updated document
```
quietshu: I think there're 3 missing backticks ` ``` ` after this line.

sergiodxa: Done

src/use-swr.ts
```diff
@@ -90,16 +90,30 @@ const broadcastState: broadcastStateInterface = (key, data, error) => {
...
-const mutate: mutateInterface = async (_key, _data, shouldRevalidate) => {
+const mutate: mutateInterface = async (
+  _key,
+  _data,
+  shouldRevalidate = true
+) => {
   const [key] = getKeyArgs(_key)
   if (!key) return
 
+  // if there is no new data data, call revalidate against the key
```
```suggestion
  // if there is no new data, call revalidate against the key
```

src/use-swr.ts
```diff
@@ -90,16 +90,30 @@ const broadcastState: broadcastStateInterface = (key, data, error) => {
...
+  if (!_data) return trigger(_key)
```
```suggestion
  if (typeof _data === 'undefined') return trigger(_key, shouldRevalidate)
```

test/use-swr.test.tsx
```diff
@@ -846,13 +847,82 @@ describe('useSWR - local mutation', () => {
...
     })
     await act(() => new Promise(res => setTimeout(res, 110)))
     expect(container.textContent).toMatchInlineSnapshot(`"data: 999"`)
   })
 
+  it('should trigger on mutation without data', async () => {
+    let value = 0
+
+    function Page() {
+      const { data } = useSWR('dynamic-12', () => value++, {
```
```suggestion
      const { data } = useSWR('dynamic-14', () => value++, {
```

test/use-swr.test.tsx
```diff
@@ -846,13 +847,82 @@ describe('useSWR - local mutation', () => {
...
+    }
+    const { container } = render(<Page />)
+
+    // hydration
+    expect(container.firstChild.textContent).toMatchInlineSnapshot(`"data: "`)
+    await waitForDomChange({ container }) // mount
+    expect(container.firstChild.textContent).toMatchInlineSnapshot(`"data: 0"`)
+    await act(() => {
+      // trigger revalidation
+      mutate('dynamic-12')
```
```suggestion
      mutate('dynamic-14')
```

test/use-swr.test.tsx
```diff
@@ -846,13 +847,82 @@ describe('useSWR - local mutation', () => {
...
+      return new Promise(res => setTimeout(res, 1))
+    })
+    expect(container.firstChild.textContent).toMatchInlineSnapshot(`"data: 1"`)
+  })
+
+  it('should call function as data passing current cached value', async () => {
+    // prefill cache with data
+    cacheSet('dynamic-13', 'cached data')
```
```suggestion
    cacheSet('dynamic-15', 'cached data')
```

test/use-swr.test.tsx
```diff
@@ -846,13 +847,82 @@ describe('useSWR - local mutation', () => {
...
+    const callback = jest.fn()
+    await mutate('dynamic-13', callback)
```
```suggestion
    await mutate('dynamic-15', callback)
```

test/use-swr.test.tsx
```diff
@@ -846,13 +847,82 @@ describe('useSWR - local mutation', () => {
...
+    expect(callback).toHaveBeenCalledWith('cached data')
+  })
+
+  it('should return results of the mutation', async () => {
+    // returns the data if promise resolved
+    expect(mutate('dynamic-14', Promise.resolve('data'))).resolves.toBe('data')
```
```suggestion
    expect(mutate('dynamic-16', Promise.resolve('data'))).resolves.toBe('data')
```

test/use-swr.test.tsx
```diff
@@ -846,13 +847,82 @@ describe('useSWR - local mutation', () => {
...
+
+    // throw the error if promise rejected
+    expect(
+      mutate('dynamic-14', Promise.reject(new Error('error')))
```
```suggestion
      mutate('dynamic-16', Promise.reject(new Error('error')))
```

test/use-swr.test.tsx
```diff
@@ -846,13 +847,82 @@ describe('useSWR - local mutation', () => {
...
+    ).rejects.toBeInstanceOf(Error)
+  })
+
+  it('should get bound mutate from useSWR', async () => {
+    function Page() {
+      // eslint-disable-next-line no-shadow
+      const { data, mutate: boundMutate } = useSWR(
+        'dynamic-15',
```
```suggestion
        'dynamic-17',
```

quietshu:
Thank you @sergiodxa! 

Close #262, #136.

PR #248, opened 2020-01-31
Change suspense handling to allow changes in config.suspense

slightlytyler:
Right now changing `config.suspense` from true -> false causes a hooks order error
```
Warning: React has detected a change in the order of Hooks called by OffProductNotificationLivePreviewScene. This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks: https://fb.me/rules-of-hooks

   Previous render            Next render
   ------------------------------------------------------
1. useContext                 useContext
2. useState                   useState
3. useContext                 useContext
4. useRef                     useRef
5. useRef                     useRef
6. useState                   useState
7. useCallback                useCallback
8. useRef                     useRef
9. useRef                     useRef
10. useCallback               useCallback
11. useLayoutEffect           useLayoutEffect
12. useLayoutEffect           useLayoutEffect
13. useEffect                 useMemo
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
This shouldn't be happening and we can avoid this problem and allow dynamic `config.suspense` by reordering some of the logic

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -470,7 +470,10 @@ function useSWR<Data = any, Error = any>(
 
     // set up reconnecting when the browser regains network connection
     let reconnect = null
-    if (typeof addEventListener !== 'undefined' && config.revalidateOnReconnect) {
+    if (
+      typeof addEventListener !== 'undefined' &&
+      config.revalidateOnReconnect
+    ) {
       reconnect = addEventListener('online', softRevalidate)
     }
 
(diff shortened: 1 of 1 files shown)


slightlytyler:
After reading #92 it seems like this lib is trending towards using a separate hook`useSWRSuspense`. With that in mind I think the pattern of dynamic `config.suspense` is incorrect though it still seems like the current implementation is broken in that regard. I've refactored my code to like:
```js
const { data: initialData } = useSwr(takeoverRequestHandling ? null : '/view', fetcher);
const { data, error, isValidating } = useSwr(takeoverRequestHandling ? '/live-view' : null, liveFetcher);
```

PR #270, opened 2020-02-16
Add React and Hook config for ESLint

innocentiv:
This PR adds eslint-plugin-react and eslint-plugin-react-hooks to the swr repository.

ESLint config extends plugin:react/recommended and add the react-hooks/rules-of-hooks rule as warning.

The use of useMemo after the early return in `if (config.suspense)` is a violation of the rules of hooks and even if I do not expect config.suspense to change during the normal use of the library, I moved it above that piece of code. I think it is a safer approach. We could add and an `eslint-disable-next-line` comment if you prefer.


pke:
@innocentiv could you rebase please to increase the chance of a merge?

innocentiv:
I have rebased this feature, updated the eslint plugins to the latest version, changed the rules to the recommended config from react docs (now they fixed the autofix on react-hooks/exhaustive-deps). I hope this suffices, but after seven months I have nome doubt 😄

shuding (approved):
Thank you for your great work @innocentiv!

PR #272, opened 2020-02-17
Export the existing cacheClear function (one liner)

JulianG:
I think this fixes https://github.com/zeit/swr/issues/161 by simply exposing an existing `cacheClear` function.

I've seen other PRs that address other cache related issues, but I think this is the minimum set of changes (one line) that will enable isolated deterministic tests when manipulating the "key" is not an option.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -9,3 +9,5 @@ export {
   responseInterface
 } from './types'
 export default useSWR
+
+export { cacheClear } from './config'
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -9,3 +9,5 @@ export {
   responseInterface
 } from './types'
 export default useSWR
+
+export { cacheClear } from './config'
```
JulianG: I would personally prefer to export this as a function called `clearCache`. But I'm happy either way.

quietshu:
Thank you for your contribution @JulianG! However we already have #231 in the works, which will expose more cache APIs including clearing the cache. I'm going to close this PR for now.  🙏

JulianG:
Yes. That makes sense. Cheers!

PR #277, opened 2020-02-22
add errorRetryCount config option

### LINKED ISSUE #276: Setting max retryCount on error
Is there a way to add a max retryCount on error without overriding the entire onErrorRetry? I think it'd be nice to be able to keep the very reasonable default behaviour of exponential backoff and stopping on ```!isDocumentVisible()```, but be able to tweak and limit max tries via config.

What do you think?

dominictwlee:
Add errorRetryCount config option to limit error retries.

closes #276

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -43,6 +43,10 @@ function onErrorRetry(
     return
   }
 
+  if (config.errorRetryCount && opts.retryCount > config.errorRetryCount) {
+    return
+  }
+
   // exponential backoff
   const count = Math.min(opts.retryCount || 0, 8)
   const timeout =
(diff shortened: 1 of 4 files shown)


src/config.ts
```diff
@@ -43,6 +43,10 @@ function onErrorRetry(
     return
   }
 
+  if (config.errorRetryCount && opts.retryCount > config.errorRetryCount) {
```
whj1995, line 46: errorRetryCount = 0, nerver return

dominictwlee: good catch, I'll make a PR for it later in the day, thanks.

PR #279, opened 2020-02-22
fix: disable dedupe when trigger/mutate (#271)

### LINKED ISSUE #271: [Question]: is it as designed that trigger and mutate will follow dedupingInterval
Hey,

small question: is it as designed, that a manual trigger will follow the `dedupingInterval` and don't invalidate the cached item if it was less than the `dedupingInterval`?

Example: https://codesandbox.io/s/quirky-albattani-ttfrw

Expected: On each button click the new time will be fetched

Actual: It only will be fetched if I click slower than 200ms (default `dedupingInterval`)

Version: 0.1.17

rtxu:
fix #271

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -67,7 +67,6 @@ const getKeyArgs = key => {
   return [key, args]
 }
 
-const DEDUPE = true
 const NO_DEDUPE = false
 
 const trigger: triggerInterface = (_key, shouldRevalidate = true) => {
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -67,6 +67,9 @@ const getKeyArgs = key => {
   return [key, args]
 }
 
+const DEDUPE = true
+const NO_DEDUPE = false
+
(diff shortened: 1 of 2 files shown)


rtxu:
Here is a test demo

[Edit swr_user_experience](https://codesandbox.io/s/swruserexperience-7glnd?fontsize=14&hidenavigation=1&theme=dark)

mxschmitt:
@rtxu I've duplicated two working tests for trigger and mutate and added there a `dedupingInterval` to show that the fix is working.

quietshu:
Thank you @rtxu and @mxschmitt !

Tom-Goji:
This change is exactly what we were looking for thanks 👍 

@shuding It looks like it should have made it into either the 0.5.0 or 0.5.1 tags but I can't see it in either. Has this been held back from release? https://github.com/vercel/swr/commits/0.5.1

shuding:
@Tom-Goji I don’t think this was held back, are you seeing the same bug with the latest release? If so feel free to open a new issue with reproduction, thank you!

PR #283, opened 2020-02-25
Add mutateMany function

### LINKED ISSUE #264: Mutate multiple call results
I'd like to be able to mutate the cached results for a variable number of call results. Essentially, what I'm after is a `mutate` function not with the current signature:

```ts
mutate(
  key,
  data,
  shouldRevalidate?
): void
```

..but rather with a signature something like this:

```ts
mutateMany(
  select: (key) => boolean,
  update: (result: R) => R
): void
```

or this:

```ts
type updater = (result: R) => R
mutateMany(
  selectivelyUpdate: (key) => void | updater
): void
```

In PR https://github.com/zeit/swr/pull/245 (Improve mutate function) I found a clever trick (the "bound mutator") to relatively easy work around the current non-existence of this feature. I made a little userland proof-of-concept out of the idea, which looks something like this:

```ts
import useSWR, { mutate as swrMutate } from "swr";

const BOUND_MUTATORS: Array<{ key, boundMutator }> = [];

export function useWrappedSWR(key, ...) {
  const swr = useSWR(key, ...);

    if (key) {
      const entry = {
        key,
        boundMutator(update: (result: any) => any) {
          if (swr.data) {
            swrMutate(key, update(swr.data));
          }
        }
      };

      BOUND_MUTATORS.push(entry);

      return () => {
        const i = BOUND_MUTATORS.indexOf(entry);
        if (i >= 0) {
          BOUND_MUTATORS.splice(i, 1);
        }
      };
    }
  }, [key]);

  return swr;
}

export function mutateMany(selectivelyUpdate: (key: any) => void | ((result: any) => any)) {
  BOUND_MUTATORS.forEach(({ key, boundMutator }) => {
    const update = selectivelyUpdate(key);
    if (update) {
      boundMutator(update);
    }
  });
}
```

👉 **Is this kind of feature something you are looking into, or willing to think about/support?**

I'd be very very happy, am looking forward to using this library but currently rely on this kind of (self-written) cache updating logic in a project of mine with a relatively large codebase :)

sergiodxa:
Resolves #264 

This PR adds a `mutateMany` function, it has the following definition

```ts
export type mutateManyInterface<Data = any> = (
  select: (key: keyInterface) => boolean,
  data?: Data | Promise<Data> | mutateCallback<Data>,
  shouldRevalidate?:
    | boolean
    | ((key: keyInterface, currentValue: Data) => boolean)
) => Promise<Data[]>
```

The function will be called as

```ts
mutateMany(
  key => key.startsWith("/api/users"),
  undefined
)
```

The way this function works is:

1. It calls select passing for each key in the cache, there you could check if you want to mutate it or not (returned true/false)
2. It calls `mutate` against the selected keys and pass data, this data is the same as `mutate`
3. It calls shouldRevalidate (if it's a function) passing the selected key to get if it should or not revalidate, it could also be a boolean to pass true/false to all selected keys

@quietshu if you think this is a good addition to SWR I could document it and add some tests.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -58,14 +58,14 @@ export type triggerInterface = (
   key: keyInterface,
   shouldRevalidate?: boolean
 ) => void
-type mutateCallback<Data = any> = (currentValue: Data) => Data
+type mutateCallback<Data = any> = (currentValue: Data, key: string) => Data
 export type mutateInterface<Data = any> = (
   key: keyInterface,
   data?: Data | Promise<Data> | mutateCallback<Data>,
   shouldRevalidate?: boolean
 ) => Promise<Data | undefined>
 export type mutateManyInterface<Data = any> = (
-  select: (key: keyInterface) => boolean,
+  select: (key: string) => boolean,
   data?: Data | Promise<Data> | mutateCallback<Data>,
   shouldRevalidate?:
     | boolean
--- a/src/types.ts
+++ b/src/types.ts
@@ -64,6 +64,13 @@ export type mutateInterface<Data = any> = (
   data?: Data | Promise<Data> | mutateCallback<Data>,
   shouldRevalidate?: boolean
 ) => Promise<Data | undefined>
+export type mutateManyInterface<Data = any> = (
+  select: (key: keyInterface) => boolean,
+  data?: Data | Promise<Data> | mutateCallback<Data>,
+  shouldRevalidate?:
+    | boolean
+    | ((key: keyInterface, currentValue: Data) => boolean)
+) => Promise<Data[]>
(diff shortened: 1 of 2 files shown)


quietshu:
I don't think we should add it to this lib. The goal is that we want this lib to be extremely simple and performant. IMO things like this can be implemented with existing APIs.

mxschmitt:
Looks like this is a duplicate of #268

sergiodxa:
@quietshu 👍 I agree, will close it and if someone wants to implement it could use it as an example implementation

Also if someones one to publish this as a package in npm feel free to do it.

PR #292, opened 2020-03-05
Change useRequest.js to use `mutate`

mattfwood:
Looks like `revalidate` has been changed to `mutate`, so if you try to use this example it'll say that `mutate` is not a function.


quietshu:
Thanks!

PR #297, opened 2020-03-11
Fix 'reconnect' assignment

zhengjitf:
```js
reconnect = addEventListener('online', softRevalidate)
```
`addEventListener()` return `undefined`，should change to:
```js
addEventListener('online', reconnect = softRevalidate)
```


quietshu:
Thank you!

PR #300, opened 2020-03-12
Use window as a target for online event

### LINKED ISSUE #301: Explicit target for online event
Hi, currently we subscribe and unsubscribe to `'online'` event using global `addEventListener` and `removeEventListener`

Unfortunately there seems to be a Chrome bug triggered by Sentry's SDK which leads to `TypeError`s when `this` is `undefined`. More info in [this issue](https://github.com/getsentry/sentry-javascript/issues/2074).
It seems like bug is pretty tricky to fix on Sentry's side, but it's pretty easy to fix on SWR's side. I've created a PR with a [fix](https://github.com/zeit/swr/pull/300)

Also we leak subscriptions, because the assignment to reconnect is wrong, but that seems to be addressed in [this PR](https://github.com/zeit/swr/pull/297)

nulladdict:
Also fix unsubscribe callback assignment

Closes #301

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -464,11 +464,8 @@ function useSWR<Data = any, Error = any>(
 
     // set up reconnecting when the browser regains network connection
     let reconnect = null
-    if (
-      typeof addEventListener !== 'undefined' &&
-      config.revalidateOnReconnect
-    ) {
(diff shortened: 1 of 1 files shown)


nulladdict:
Sure thing, I've resolved the conflicts

quietshu:
Thanks @nulladdict !

Issue #290, opened 2020-03-05
Warning in React 16.13.0


mikebarkmin:
Since React 16.13.0 is get this error:

```
Warning: Cannot update a component from inside the function body of a different component.
    in Unknown (created by Pagination)
    in div (created by ForwardRef(Grid))
    ...
```

It occurs when is use the `useSWRPages` hook. Everything seems to be working. But maybe this will lead to unwanted behavior.

Do you know what is causing this warning?

This is the component where is use `useSWRPages`:

```
function Pagination({
  name,
  pageSize,
  filter,
  order,
  orderBy,
  url,
  selector,
  Renderer,
  variant,
  spacing,
  placeholder
}) {
  const {
    pages,
    isEmpty,
    isLoadingMore,
    isReachingEnd,
    loadMore
  } = useSWRPages(
    name,
    ({ offset, withSWR }) => {
      const page = offset / pageSize;
      // eslint-disable-next-line react-hooks/rules-of-hooks
      const { data } = withSWR(
        useRequest(
          `${url}?${constructPagingQuery(
            page,
            pageSize,
            order,
            orderBy,
            humps.decamelizeKeys(filter)
          )}`
        )
      );

      if (!data) {
        return null;
      }

      if (variant == 'grid') {
        return selector(data).map(d => (
          <Grid item key={d.id} xs={12} sm={6} lg={3} xl={2}>
            <Renderer style={{ height: '100%' }} {...d} />
          </Grid>
        ));
      } else if (variant == 'list') {
        return selector(data).map(d => (
          <Grid item key={d.id} xs={12}>
            <Renderer style={{ height: '100%' }} {...d} />
          </Grid>
        ));
      }
    },
    (SWR, index) => {
      if (SWR.data && selector(SWR.data).length === 0) return null;
      return (index + 1) * pageSize;
    },
    [filter, order, orderBy, url]
  );

  return (
    <Box display="flex" flexDirection="column">
      {!isEmpty ? (
        <>
          <Grid
            container
            spacing={spacing}
            justify="center"
            alignItems="stretch"
          >
            {pages}
          </Grid>
          <Box
            marginTop={4}
            display="flex"
            flexDirection="column"
            alignItems="center"
          >
            <ProgressButton
              onClick={loadMore}
              disabled={isReachingEnd || isLoadingMore}
              loading={isLoadingMore}
            >
              {isLoadingMore
                ? '. . .'
                : isReachingEnd
                ? 'no more data'
                : 'load more'}
[...]

PR #317, opened 2020-03-25
Wrap side effects to fix new warnings in React 16.13

neolefty:
Fix for #290 "Warning in React 16.13.0"

[New warnings in React 16.13](https://reactjs.org/blog/2020/02/26/react-v16.13.0.html#warnings-for-some-updates-during-render) — especially triggering `setState()` from another component.

This change:
* Is minimal. It wraps `setTimeout()` around two `setState()` calls in `useSWRPages()`. There's a comment in it that we can probably delete.
* Silences the new warnings people are seeing in #290 — tested on branch [@neolefty/publish](https://github.com/neolefty/swr/tree/publish) and available to try out from [@neolefty/swr](https://www.npmjs.com/package/@neolefty/swr).
* Probably adds another rendering roundtrip by delaying `setState()`. But that's actually the point of the new warnings, I think.

### DIFF
--- a/src/use-swr-pages.tsx
+++ b/src/use-swr-pages.tsx
@@ -149,29 +149,32 @@ export function useSWRPages<OffsetType = any, Data = any, Error = any>(
         pageSWRs[id].error !== swr.error ||
         pageSWRs[id].revalidate !== swr.revalidate
       ) {
-        setPageSWRs(swrs => {
-          const _swrs = [...swrs]
-          _swrs[id] = {
-            data: swr.data,
-            error: swr.error,
-            revalidate: swr.revalidate,
-            isValidating: swr.isValidating,
-            mutate: swr.mutate
+        // hoist side effects: setPageSWRs and setPageOffsets -- https://reactjs.org/blog/2020/02/26/react-v16.13.0.html#warnings-for-some-updates-during-render
+        setTimeout(() => {
+          setPageSWRs(swrs => {
+            const _swrs = [...swrs]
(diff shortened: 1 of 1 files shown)


src/use-swr-pages.tsx
```diff
@@ -149,29 +149,32 @@ export function useSWRPages<OffsetType = any, Data = any, Error = any>(
...
       ) {
-        setPageSWRs(swrs => {
-          const _swrs = [...swrs]
-          _swrs[id] = {
-            data: swr.data,
-            error: swr.error,
-            revalidate: swr.revalidate,
-            isValidating: swr.isValidating,
-            mutate: swr.mutate
+        // hoist side effects: setPageSWRs and setPageOffsets -- https://reactjs.org/blog/2020/02/26/react-v16.13.0.html#warnings-for-some-updates-during-render
```
neolefty:
```suggestion
```
Okay to delete this comment?

src/use-swr-pages.tsx
```diff
@@ -149,29 +149,32 @@ export function useSWRPages<OffsetType = any, Data = any, Error = any>(
...
+        setTimeout(() => {
```
neolefty: Another way to hoist is `useEffect()`s, possibly with dependency arrays to take the place of the `if` clauses below.

src/use-swr-pages.tsx
quietshu: I think it's better to keep it! :)

PR #321, opened 2020-03-30
Minor documentation improvements

akx:
I found the docs for `mutate` a little lacking, so here's a couple small improvements.

Closes #322.

PR #322, opened 2020-03-30
update documentation about mutate

jan-paulus:
Includes and extends #321 (credits to @akx)

### DIFF
--- a/README.md
+++ b/README.md
@@ -341,7 +341,8 @@ function Profile () {
         // send a request to the API to update the data
         await requestUpdateUsername(newName)
         // update the local data immediately and revalidate (refetch)
-        mutate({ ...data, name: newName })
+        // NOTE: key has to be passed to mutate as it's not bound
+        mutate('/api/user', { ...data, name: newName })
       }}>Uppercase my name!</button>
     </div>
   )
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -387,6 +388,12 @@ try {
 }
 ```
 
+### Bound `mutate()`
+
+The SWR object returned by `useSWR` also contains a `mutate()` function that is pre-bound to the SWR's key.
+
+It is functionally equivalent to the global `mutate` function but does not require the `key` parameter.
+
```
pacocoursey: Maybe we can add a simple code example? Easier to parse than reading imo

```suggestion

const { mutate } = useSWR('/api/user')
```

jan-paulus:
Added a proper code example as suggested. @pacocoursey

README.md
```diff
@@ -341,7 +341,7 @@ function Profile () {
         // send a request to the API to update the data
         await requestUpdateUsername(newName)
         // update the local data immediately and revalidate (refetch)
-        mutate('/api/user', { ...data, name: newName })
+        mutate({ ...data, name: newName })
```
jasonlafferty: should this line have changed? as this isn't a bound `mutate`

quietshu: I think we should add a comment which mentions that this is the global `mutate`.

jan-paulus: Yes, my mistake. Will fix it.

quietshu:
Closed in favor of #321.

PR #326, opened 2020-04-01
Add small note on cache.clear + testing in the Readme

samselikoff:
I know this is a newer feature but it took me some digging to find it. `cache.clear()` works like a charm!

### DIFF
--- a/README.md
+++ b/README.md
@@ -524,9 +524,7 @@ Your test setup will end up looking something like this. Assume we're testing ou
 import React from "react";
 import {
   render,
-  waitForElement,
-  waitForElementToBeRemoved,
-  fireEvent
+  waitForElementToBeRemoved
 } from "@testing-library/react";
 import AppComponent from "../App";
 import { cache, SWRConfig } from "swr";
--- a/README.md
+++ b/README.md
@@ -511,17 +511,58 @@ Together with techniques like [page prefetching](https://nextjs.org/docs#prefetc
 
 ## Testing
 
-Because SWR uses a global cache, you will need to reset it in between your tests. Otherwise, cached data could leak across tests.
+Because SWR uses a global cache, you'll need to do some cleanup after each test to ensure that no state leaks across your tests.
 
-To reset the cache use the `cache` named import and call `cache.clear()` after each test:
+There are two steps to resetting the cache:
+
+1. Import `cache` from `swr` and call `cache.clear()` after each test.
+2. Set the [`dedupingInterval`](https://github.com/zeit/swr#options) config option to 0 when rendering your app, so SWR doesn't ignore requests from new test runs before the deduping interval has had a chance to pass. You can do this using the `SWRConfig` provider context directly in your tests.
+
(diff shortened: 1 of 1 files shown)


samselikoff:
@sergiodxa yep, thanks for mentioning that I actually had to go back to my app and add that as well

samselikoff:
Updated!

samselikoff:
Ran into this again! Would love to see it in the readme, anything left for me to do?

aboutlo:
Hi there, I'm facing the same issue.  

However `cache.clear()` isn't fixing the problem at least with SWR 0.2.2 
I create an example here: https://codesandbox.io/s/frosty-butterfly-y295j?file=/src/swr.spec.js

The only current workaround is using different keys between the tests. Which is actually what it's done in SWR itself: https://github.com/vercel/swr/blob/master/test/use-swr.test.tsx

Is this a known issue?

btw. I love this lib thank you for your hard work.

sergiodxa:
@aboutlo you need to also wrap your rendered component in SWRConfig and set the dedupingInterval to zero, otherwise SWR will reuse the promises of the previous test and refill the cache with the same data.

aboutlo:
@sergiodxa gotcha! this has fixed the problem and helped me to understand another bit of SWR magic 🙏 

For anyone else who could have a similar issue `dedupingInterval:0` can be passed directly to `useSWR` without having `SWRConfig` as a wrapper.

```js
const { data } = useSWR(["/something"], {
    fetcher,
    dedupingInterval:0
});
```

sergiodxa:
I don’t recommend you to pass it to useSWR directly, in production you want to have it enabled, that is why SWRConfig allow you to replace it in the test.

aboutlo:
Of course, not for production, I was stuck just with a test :)

barakyosi:
Is there a way to allow setting `dedupingInterval` globally in tests?

The example in readme won't work if `AppComponent` renders `SWRConfig` internally (which happens in the component I wish to test).

Let me know if you want me to open a new issue for it.

BTW - love your work! It's super helpful :pray:

hg-pyun (approved):
LGTM

shuding:
Thanks for all your discussions and contributions here! With 1.0 we finalized the new cache API, and a better way to test: https://swr.vercel.app/docs/advanced/cache#reset-cache-between-test-cases.

samselikoff:
Thanks for following up!

I use Cypress to do black-box testing rather than shallow mounting, so I'll still need to import + call `cache.clear()`. Should this be okay in concurrent React? Maybe worth mentioning as lots of folks use Cypress or something like it.

shuding:
Thanks for the feedback! Yeah, in that case it makes sense to clear the global cache from outside of the React tree. I think we will need a dedicated “Testing” page in the docs soon, and currently it can be achieved by this (since the exported `cache` was gone):

```js
import { SWRConfig } from 'swr'

SWRConfig.default.cache.clear()
```

We will make sure to put this in the docs (we also removed all the docs in the readme so we don’t have to maintain two places) :)

PR #335, opened 2020-04-07
Fixes related to cache clear and concurrent promises

slightlytyler:
Found a bug while trying to write some tests for components using SWR. `cache.clear()` doesn't clear the concurrent promises so further calls get stuck in the suspense fallback (not quite sure why that's the resulting behavior).  Simply iterating through the concurrent promises and deleting entries fixes the problem. 

Also had to update react testing library and JSDOM, otherwise the tests pass incorrectly (pretty sure this isn't a bug in either of those libraries). The JSDOM changes the required node version to > 10 to develop the project though

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,4 +1,5 @@
 module.exports = {
   preset: 'ts-jest',
   testRegex: '/test/.*\\.test\\.tsx$',
-};
+  testEnvironment: 'jest-environment-jsdom-sixteen'
+}
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     ]
   },
   "devDependencies": {
-    "@testing-library/react": "9.3.0",
+    "@testing-library/react": "^10.0.2",
     "@types/jest": "24.0.20",
     "@types/node": "11.12.0",
     "@types/react": "16.9.11",
@@ -49,6 +49,7 @@
     "eslint-config-prettier": "6.5.0",
     "husky": "2.4.1",
     "jest": "24.9.0",
+    "jest-environment-jsdom-sixteen": "^1.0.3",
     "lint-staged": "8.2.1",
     "prettier": "1.18.2",
     "react": "16.11.0",
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -1,6 +1,7 @@
 import { CacheInterface, keyInterface, cacheListener } from './types'
 import { mutate } from './use-swr'
 import hash from './libs/hash'
+import { CONCURRENT_PROMISES } from './config'
 
 export default class Cache implements CacheInterface {
   private __cache: Map<string, any>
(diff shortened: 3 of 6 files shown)


slightlytyler:
Something to note is that I'm aware of discussion like https://github.com/zeit/swr/pull/326  and `dedupingInterval` but from my testing that doesn't solve the problem

slightlytyler:
For reference here is an example test that was failing before this change
```js
import React from 'react';
import { cache } from 'swr';
import { render, waitFor } from '@testing-library/react';
import SomeComponent from './SomeComponent;
import mockedHttpClient from './mockedHttpClient';

describe('container error handling', () => {
  afterEach(() => {
    cache.clear();
  });

  it('should render a homepage at the root url', async () => {
    const { queryByText } = render(<SomeComponent />);
    await waitFor(() => expect(queryByText(/some title/i)).toBeVisible());
  });

  it('should render an error message when srosh is 500ing', async () => {
    mockedHttpClient.onGet('/some_path').replyOnce(500);
    const { queryByText } = render(<SomeComponent />);
    await waitFor(() => expect(queryByText(/error boundary text/i)).toBeVisible());
    expect(queryByText(/title/i)).toBeNull();
  });
});
```

shuding:
Thank you for this amazing PR @slightlytyler! I think I'll look deeper into this because from my understanding, the concurrent promises will be removed right after throwing to <Suspense>, if you have `dedupingInterval` set to 0. So this seems like another bug.

slightlytyler:
@shuding you're write I think, after looking again I found a bug with how deduping interval was handled https://github.com/zeit/swr/pull/335/files#diff-ab20307c185620c9b54994fa4dcee1b7R270

my change there fixes a part of the failing test I added but it doesn't completely pass yet

slightlytyler:
Been messing with it for a while and still can't quite narrow the source of the race conditions. Different values in the timeouts produce different failures so pretty sure it's a race condition with concurrent promises beyond the initial one I mentioned

slightlytyler:
Looking again and seems like the problem is that on initial mount `revalidate` is called twice. Once as a hard revalidation then again as a soft revalidation. The soft revalidation fails the `shouldDeduping` check since concurrent promises have already been removed from the deduping timeout. Instead the soft revalidation get's processed as a hard revalidation, and resets concurrent promises

I don't think this produces a bug in the real behavior but it totally messes with tests

slightlytyler:
Finally got something to  pass  the tests. Basic idea is that  you tracker a counter `cache.version`, clear increments it, and then in the hook we check for a mismatch in  `revalidate` and  throw away updates that fail the check. `cache.clear` also synchronously removes concurrent promises

Not sure if these are the right patterns but I think the test case is valid

Anyways, love SWR, hope I can help you get this solved 👍

alexl0gan:
My team are also having the same issue. Any progress on this pr @slightlytyler @shuding ?

slightlytyler:
@alexl0gan not working on a project that uses SWR anymore but I think react-query's `ReactQueryCacheProvider` is a better way to solve this problem then trying to manipulate the singleton cache synchronously in tests. Closing this as I'm sure it's useless to the project at this point

kristojorg:
@slightlytyler I don't think this is useless to the project... I'm still facing the same issue

kristojorg:
I've resorted to mocking SWR entirely to resolve this, for anyone else looking

PR #341, opened 2020-04-12
Add overload to improve type inference

Janpot:
Add a function overload for `useSWR` that improves [type inferencing for array keys](https://github.com/Janpot/next-rpc/blob/f677daa0db8903d2e80a2edebb445e1c6ff4a45a/examples/with-swr/pages/index.js#L8).

Unfortunately a workaround for up to 4 parameters had to be made to get around missing [variadic generics](https://github.com/microsoft/TypeScript/issues/5453)

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -117,9 +117,24 @@ const mutate: mutateInterface = async (
   return data
 }
 
-function useSWR<K extends any[], Data = any, Error = any>(
-  key: K,
-  fn: (...args: K) => Promise<Data>,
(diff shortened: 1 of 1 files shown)


shuding:
Thanks @Janpot! Now it seems that TS has launched https://github.com/microsoft/TypeScript/pull/39094 (will be included in 4.0), I think we can wait and solve this elegantly once we can use it.

vhakulinen:
Any updates on this?

huozhi:
It's already launched in swr! Close as resolved

PR #353, opened 2020-04-23
Swap out Graph.Cool API

motleydev:
Graph.Cool will be sunset soon, this endpoint mimics the same data and similar query structure.


shuding:
I think we should instead use a fake API endpoint as an example, just for showing off the idea. 😄

PR #365, opened 2020-05-01
Allow cache serialization and subscribers to know the updated key and type of event

sergiodxa:
I'm working on a SWR DevTools with @cristianbgp and being able to serialize the cache (useful to download the current version) and to subscribe to specific keys will make the development way easier.

Right now I have some workarounds, iterating the whole cache to get all the keys for both, serializing and know what changed.

The new subscription function will also help build other libraries like https://github.com/sergiodxa/swr-sync-storage which are currently iterating the whole cache on each change to serialize it.

BTW, in a future PR I would like to make useSWR use cache.subcribe internally to get the updates, this will allow us to easily count active hooks and purge keys without subscribed as attempted in #92


gaoxiaoliangz:
What's the latest update on this? Is this PR going to be merged? I really want this feature.

christian-valadez:
+1, this feature would be really great to have!

PR #366, opened 2020-05-01
Add support to detect inactive keys and purge

sergiodxa:
This is built on top of https://github.com/zeit/swr/pull/365 and implements #92

This adds support to purge inactive keys. The way it does it is by running `cache.subscribe` on the effect run in useSWR (this subscription does nothing right now), this way we can know which keys are active based on the amount of hooks subscribed to them.

When a component unmount it will unsubscribe from the cache, this will mark a key as deactivated and call `cache.purge`, this method will check if the cache size has reached the `MAX_SIZE` (customizable with an environment variable, I could change it if you don't like this approach) and if it had reached it it will delete the keys marked as inactive, clear the list of inactive keys, try to deactivate all the remaining keys (this will mark as deactivate keys that were nor marked as such before the purge) and lastly trigger an event to all wildcard the subscribers.


src/cache.ts
```diff
@@ -35,14 +45,38 @@ export default class Cache implements CacheInterface {
...
     const [_key] = this.serializeKey(key)
     if (shouldNotify) mutate(key, null, false)
     this.__cache.delete(_key)
-    this.notify()
+    this.trigger('delete', _key)
+  }
+
+  isActive(key: keyInterface) {
+    const [_key] = this.serializeKey(key)
+    return this.__inactive.has(_key)
```
ties-v, line 60:
```suggestion
    return !this.__inactive.has(_key)
```

sergiodxa: Nice catch! Thanks

sergiodxa:
@shuding where you able to check this? what do you think? is there something you would like to change?

PR #367, opened 2020-05-01
Add a default fetcher using Fetch API

sergiodxa:
Not sure if you want this, but I found most of the time a simple fetcher like `url => fetch(url).then(res => res.json())` is more than enough to use useSWR until you have reasons to build your own fetcher (like handling errors or parsing the response in a specific way).

I believe adding a default one could make using SWR a little bit faster since now you will only need to install and use.

@shuding if you prefer to don't provide a default one feel free to close this PR.


shuding:
I agree on this one, but would love to hear more discussions!

allpwrfulroot:
Love it. Would request a note about best practices for error responses; that particular default leads to a lot of vague SyntaxErrors

sergiodxa:
I think when you want to handle error errors is when you will start to use a custom build fetcher, this fetcher should allow you to start using SWR without any config or boilerplate until you need a better fetcher

sergiodxa:
But yeah we may need to better document how to handle errors, specially because Fetch does not throw in all the HTTP errors

leerob:
I really like this idea, since it makes it easier to hit the ground running with SWR.

cristianbgp:
This is great, I always need to make that simple fetcher on small demos. Also, I think this would help new people using SWR easily

shuding (approved):
Based on the feedback we should take it! :+1:

brandonpittman:
So, I tried [Paco's idea for SWR as a local cache](https://paco.sh/blog/shared-hook-state-with-swr) but this PR fills the console with localhost:3000/whatever 404 errors. 😱

sergiodxa:
@brandonpittman you will need to define a custom fetcher then, maybe you can try passing undefined as your fetcher 🤔

brandonpittman:
@sergiodxa Yeah, passing `null` in the fetcher spot fixed it. Thanks!

PR #370, opened 2020-05-03
Remove the server side suspense warning

shuding:
In concurrent mode Suspense is supported in server side, we can remove this warning.


shuding:
Yeah, we were doing the same work around but also moving to Concurrent mode, so far so good.

Issue #313, opened 2020-03-23
[Bug?] mutate() does not always re-render when it should


benallfree:
Has anyone encountered this? Here is the basic pattern:

```js
const { mutate } = useSwr(...)
mutate( data=>{...data, newElem: 'foo'} )
```
I have a case where multiple async operations fire `mutate` in an async callback, but I noticed that the final `mutate` doesn't always re-render.

sergiodxa:
```js
mutate( data=>{...data, newElem: 'foo'} )
```
That should be:
```js
mutate(data=> ({...data, newElem: 'foo'}))
```
Note sure if that may cause it. If that doesn't fix it, try creating a codesandbox to replicate it.

benallfree:
@sergiodxa Here is a sandbox demonstrating the issue https://codesandbox.io/s/swr-mutate-bug-lhq8p

Please click the 'mutate' buttons successively in any order you desire to observe the problem. You will see it does not always re-render.

It seems to have something to do with nested structures. 

Mutating `["Foo 1", "Foo 2", "Foo 3"]` works fine

Mutating the subarrays in `[["Foo 1"], ["Foo 2"], ["Foo 3"]]` does not work correctly

benallfree:
HI @sergiodxa I left a few sandbox samples above, just wanted to make sure you saw it because I've made lots of edits throughout the day :)

ties-v:
I think I've got the same issue. Created an other reproduction. Maybe it helps.

https://codesandbox.io/s/happy-bird-4t2fg

benallfree:
@ties-v You need to `return {...list, items: [...list.items]};` from `bm()`, then you will have the same issue :)

Right now, you are doing `return list` which is not correct, you should always [return a new object reference](https://stackoverflow.com/questions/43638938/updating-an-object-with-setstate-in-react).

ties-v:
Ah yes, Thanks for the catch @benallfree! But indeed, when applying your fix it renders sometimes.

ties-v:
So I was digging a bit further and noticed that it works as expected when using [immer](https://github.com/immerjs/immer). See my new [sandbox](https://codesandbox.io/s/magical-napier-3wem3).

Maybe this helps tracking down the bug and provides a workaround for now

embeddedt:
I have also encountered this problem. I haven't completely fleshed out the server side of my app yet so I'm not using revalidation.

[...]

PR #373, opened 2020-05-06
enhance: clone object when retrieve data from cache

huozhi:
Try to fix https://github.com/zeit/swr/issues/313

## How
Not sure if this is what you want. To make sure the data manipulation in `boundMutate((data) => { data.xxx = ... })` won't effect the data in cache, `cache.set` will always receive a clone mirror of origin data.

This is just a workaround, welcome more discussions here

## Defects
Shapes of data could be changed. like `{get val() { return 1}}` will become `{val: 1}` in the cache

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -20,12 +20,12 @@ export default class Cache implements CacheInterface {
 
   get(key: keyInterface): any {
     const [_key] = this.serializeKey(key)
-    return clone(this.__cache.get(_key))
+    return this.__cache.get(_key)
   }
 
   set(key: keyInterface, value: any, shouldNotify = true): any {
     const [_key] = this.serializeKey(key)
-    this.__cache.set(_key, value)
+    this.__cache.set(_key, clone(value))
     if (shouldNotify) mutate(key, value, false)
     this.notify()
   }
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -2,6 +2,13 @@ import { CacheInterface, keyInterface, cacheListener } from './types'
 import { mutate } from './use-swr'
 import hash from './libs/hash'
 
+const clone = (ref: any): any => {
+  if (ref !== null && typeof ref === 'object') {
+    return Object.assign(Object.create(Object.getPrototypeOf(ref)), ref)
+  }
+  return ref
+}
+
 export default class Cache implements CacheInterface {
   private __cache: Map<string, any>
   private __listeners: cacheListener[]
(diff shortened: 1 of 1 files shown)


sergiodxa:
I’m not sure if mutating data instead of returning a new copy is something to encourage, I do it using immer to ensure I always return a new object, returning the same object all the time is what let SWR avoid unnecessary re-renders, and in mutate you could even apply logic to return the original object based on a condition and still get this benefit.

However if the idea is to always clone the object I think doing in it in set is a better idea, this way if call cache.set you will always update the cache with a new object, even if it has the same keys and values, and cache.get could still return the same reference until a set is called.

sergiodxa:
Another idea could be to integrate immer in mutate when passing a function, this way the function will always receive a draft of the current cached value and you can mutate it directly there and it will ensure you have a new copy of the object, not sure @shuding what you think about that, Immer will add a few extra kb.

huozhi:
@sergiodxa thanks for pointing out the performance issue of cloning when get from cache.

the thing I'm not sure is will cloning object break logic of caching actual class instance? i.e.

```js
class A {
  #v = 1;
  getV() { return this.#v; }
}

a = new A()
b = clone(a)

b.getV() // Error
```

would like to hear more thoughts from you. @sergiodxa @shuding

sergiodxa:
Are you caching instances of classes? That definitely works but I’m not sure about that, what’s your use case for that? Usually you will use SWR to get data from an API so most of the data tends to be serializable JSON objects.

huozhi:
I didn't have any use case I'm just guessing the edge cases 😂if it's not popular to cache instances that would be nice. totally agree that most use cases would be playing with pure objects.

sergiodxa:
@huozhi what do you think about supporting Immer produce function in `mutate` when passing a function?

This way you will be able to do:
```ts
mutate(key, draftData => {
  draftData.key = "new value"
});
```
Draft data is what Immer pass you which will receive your changes and create a new object in a safely way, you could also return undefined or the same object without changes to "discard" the mutation.

Do you think that will be easy to use and solves the problem?

sergiodxa:
I submitted an example using Immer to mutate the data here https://github.com/zeit/swr/pull/374, if SWR adds support for this the produce function would be unnecessary and you will always receive the `draftData`

huozhi:
@sergiodxa I just feel immer is a little bit heavy for this scenario. since `mutate` is only one of the functionalities of swr and callback is one of the supported argument types. personally I prefer to make it simple, and maybe not involve other libraries. but would like to hear more voice about how community want to handle it. : )

shuding:
I don't think this is a real issue, and I commented [here](https://github.com/vercel/swr/issues/313#issuecomment-638099853).

Regarding Immer, I think @sergiodxa's example is good enough and we don't have to have built-in support (of immutable data structures). Developers can choose their own.

Issue #230, opened 2020-01-14
0.1.15-beta.0 regression in ssr with initialData: staleness thing doesnt refetch. works fine in 0.1.15


iamstarkov:
as described, will add a demo asap

iamstarkov:
the problem for us is that in `0.1.16` `initialData` is behaving more or `defaultData` or `fallbackData`

* PR with a code https://github.com/iamstarkov/swr/pull/2
* notice how Pokemon profile doesnt change https://zeit-swr-230-demo.iamstarkov.now.sh/demo
* with a workaround it does change https://zeit-swr-230-demo.iamstarkov.now.sh/workaround 

workaround in the meantime:

```js
const useFixedSWR = (key, fetcher, {initialData, ...config} = {}) => {
  const initial = React.useRef(initialData);
  React.useEffect(() => {
    if(key){
      initial.current = undefined;
    }
  }, [key]);
  return useSWR(key, fetcher, {...config, initialData: initial.current});
}
```

iamstarkov:
the problem doesnt exist in `0.1.15`

* PR with a code https://github.com/iamstarkov/swr/pull/3
* notice how Pokemon profile **does change** https://zeit-swr-230-demo-0115.iamstarkov.now.sh/demo

iamstarkov:
Looks like regression was introduced in `0.1.15-beta.0`

* PR with a code https://github.com/iamstarkov/swr/pull/4
* notice how Pokemon profile does not change anymore zeit-swr-230-demo-0115.iamstarkov.now.sh/demo

iamstarkov:
this is the diff https://github.com/zeit/swr/compare/0.1.15...v0.1.15-beta.0

iamstarkov:
#211 is also in the diff, so maybe maybe

iamstarkov:
#211 with #186 are both in the diff, so i dunno

sambokai:
I think this PR is issue is not well formatted and described. It might make sense to open a new one with a concise description and not clutter it with references.

iamstarkov:
I dont agree that is "not well formatted and described". Following sentence describes the problem perfectly:

> the problem is that in 0.1.16 initialData is behaving more or defaultData or fallbackData

in regards to reference, I dont control em

huozhi:
could you try the solution posted in https://github.com/vercel/swr/issues/284#issuecomment-869174803 ?
[...]

PR #388, opened 2020-05-12
fix: cache initial data

### LINKED ISSUE #308: Interaction of mutate & initialData appears to be incorrect
Hi folks! Lovely module here. Ran into a bit of an issue.

If you use initialData in your initial call to useSWR,

```js
const { data } = useSWR(key, loadData, {
  initialData: initialData
})
```

And then someone calls `mutate` with a callback function on the page, then the `mutate` callback gets _undefined_ as its input, not the `initialData` value. Mutate [pulls data from the cache](https://github.com/zeit/swr/blob/master/src/use-swr.ts#L87), but [initialData is not set in the cache](https://github.com/zeit/swr/blob/master/src/use-swr.ts#L170).

```
mutate(key, data => ({ ...data, foo: 1 }))
```

The end result is that if someone clicks a button that triggers a callback-based mutation _before_ the initial data is revalidated, then the page likely crashes, because you just set the data to `undefined`.

Unfortunately, its not easy to work around this issue: I'd be okay with delaying the UI's appearance until there's an item in the cache, but the cache is internal. If I hid the UI based on isValidating, then it would blip whenever there are mutations. The mutate callback can't choose to return `undefined` for 'no effect'.

Ideally, mutate would be able to mutate the data in initialData, but I am probably missing something in the philosophy of swr that makes that incorrect.

aj-may:
It appears some time ago the `initialData` configuration was used as a fallback.  When #211 was merged, this behavior changed to be used with SSR like in the next.js example in the README.  Issue #230 explains this was the expectation. I'm using SSR, so im fine with the new behavior.

Since `initialData` is now not quickly revalidated, another issue (#308) has been raised.  Since `initialData` is not cached, and the mutate w/ callback grabs the curerent data from the cache, when `initialData` is used, `undefined` is returned.

fixes #308

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -167,6 +167,10 @@ function useSWR<Data = any, Error = any>(
     fn = config.fetcher
   }
 
+  if (config.initialData && !cache.has(key) && !IS_SERVER) {
+    cache.set(key, config.initialData, false)
+  }
+
   const initialData = cache.get(key) || config.initialData
   const initialError = cache.get(keyErr)
 
@@ -366,7 +370,7 @@ function useSWR<Data = any, Error = any>(
     // and trigger a revalidation
 
     const currentHookData = stateRef.current.data
-    const latestKeyedData = cache.get(key) || config.initialData
+    const latestKeyedData = cache.get(key)
 
     // update the state if the key changed (not the inital render) or cache updated
     if (


aj-may:
@Timer @shuding @pacocoursey Anything else I need to do to get this reviewed?  I wasn't able to find any contributing docs, so I'm not sure how your process works. 

Let me know if there's anything I can do to move this along.  Thanks!

src/use-swr.ts
```diff
@@ -167,7 +167,9 @@ function useSWR<Data = any, Error = any>(
     fn = config.fetcher
   }
 
-  const initialData = cache.get(key) || config.initialData
+  if (config.initialData) cache.set(key, config.initialData, false)
```
sergiodxa, line 170: What happens when the component is updated and you still have initialData in the config?

sergiodxa: Maybe this should do something like:
```ts
let initialData = cache.get(key) ?? config.initialData
if (!cache.has(key)) cache.set(key, config.initialData, false)
```
If the key is already in cache ready it, if not use the config.initialData and update the cache.

aj-may: Sorry for the delay, took me some time to figure out how to test this in my own project.  Seems to be working fine as is without these proposed changes.

Can you clarify what you mean by "when the hook is updated".  I don't fully understand the comment.

aj-may: @sergiodxa I think I understand your point now.

If I initialize the hook with key `foo` in one part of my application with initial data `['1', '2']`, but in another part of the application init the hook, again with key `foo`, but with initial data `['3', '4']` what is the desired behavior?

@sergiodxa @pacocoursey should the first or last hook take precedence?

sergiodxa: Yeh if there is initialData defined you are first overwriting the cache and then reading from it. I think if already have some cached value the expected result is to get that value and then wait for a revalidation, so the second usage of the same key should receive the cached value even if that cached value is the initial value of another call of SWR.

pacocoursey: I agree, which means the first hook should take precedence.

Hook called with initial data: `a` → `a` cached
Hook called somewhere else with different initial data → return `a`

This also means that:

Hook called without initial data but returns data: `b` → `b` cached
Hook called somewhere else with initial data → return `b` (ignore initialdata)

aj-may:
@sergiodxa @pacocoursey Updated based on the feedback. ☺️

mrmntte:
When usSWR is used with SSR, a unique global instance is created to store the cache.
How could I prevent the initialData data from being stored in the server cache?

sergiodxa:
Ohh right, this will start filling the cache server side too.

You can use cache.clear() after or before the render to clear the cache. Maybe this change should also check if it’s running server side and don’t save in cache the initial data.

aj-may:
The current version of SWR will also fill the serve-side cache if not passed `initialData`.

What's the concern with using the cache server-side? memory consumption?

If we don’t want to use the cache server-side,  then what would the desired behavior be for `useSWR` server-side?

I don't have a good reason why, but I feel like adding a `typeof window === 'undefined'` in the library is not a great option.

sergiodxa:
I'm pretty sure right now SWR is never filling the cache server-side, because the fetch always happen client-side (effects only run client-side), even if you are using Suspense it will not because current (not experimental) version of React doesn't support Suspense for SSR so you need to avoid rendering the component doing the fetch at all.

The main concern with using the cache server-side is leaking data, imagine _user A_ access `/dashboard` in Vercel, the data for that user will be stored in a cache, then _user B_ access the same page, because the required data is already in cache we will SSR the dashboard using _user A_ data instead of fetching _user B_ data, then client-side we will revalidate the data and correctly fetch _user B_ data updating the UI, while the happens the second user could access the first user dashboard and see its deployments, projects, etc.

This could be somehow avoided in a Serverless paradigm because each request should have its own Node.js instance with its own memory, nevertheless even in Serverless you usually want to keep a function warm if you are receiving multiple requests to avoid the cold start time, so you could still leak data between users.

oran1248:
@aj-may So when the fix will be released ?

aj-may:
@sergiodxa Ahh, your right!  I just tested it again and master doesn't fill the cache server side.  I push an update that will check the `IS_SERVER` value before caching.

@oran1248 I'm not a maintainer so I can't answer that question, but I just pushed an update to address comments.

aj-may:
@Timer @shuding @pacocoursey can I get a re-review on this?

src/use-swr.ts
```diff
@@ -366,7 +370,7 @@ function useSWR<Data = any, Error = any>(
     // and trigger a revalidation
 
     const currentHookData = stateRef.current.data
-    const latestKeyedData = cache.get(key) || config.initialData
+    const latestKeyedData = cache.get(key)
```
sergiodxa, line 373: This should fallback to `config.initialData` as before for the case where SWR is running server-side

shuding:
TBH I don't think it's a good idea to directly write to cache, _in the data read phrase_. Let me try to explain my concerns.

#### Correctness
Usually _cache read_ is always safe because it doesn't change the data. However mixing cache read with write will cause unexpect issues, for example it causes a _side effect_:

```js
useSWR('data', { initialData: 1 }) // → 1
useSWR('data')                     // → 1

// changing the order will cause a different result
useSWR('data')                     // → undefined
useSWR('data', { initialData: 1 }) // → 1
```

If you have a complex component tree, with Suspense or Concurrent mode enabled in the future, it will be impossible to track these side effects.

The data flow of SWR is a loop, which makes things easier to maintain:
```
                 +--> read cache ----+
                 |                   |
  render         |                   v
+--------------->+                 fetch
  mutate         |                   |
                 |                   |
                 +--- write cache <--+
```

As you can see, _cache read_ happens on render, which is the same as React itself: `render` is pure, and doesn't have any side effect. Then _fetch_ and _cache write_ are wrapped with `useEffect`, which can cause side effects.

Generally SWR obeys the rules of React, which ensures its correctness.

#### Performance
We plan to support different cache backend in the future. So it can be a Map (memory), localStorage (disk, sync), IndexedDB (disk, async) or even a remote storage (network). That means cache write can somehow be very expensive, comparing to read.

#### API Design
Regarding this issue (#230), this is one of my regrets in SWR's current API. `initialData` is per hook, and it's treated as a "fallback", rather than a "cache preset" (which is what this PR trying to do):

```js
const { data } = useSWR('data', { initialData: 1 })

// the mental model is more like
const data = useSWR('data').data || 1
```

There're many issues with that API design. It's a wrong direction and we shouldn't have gone with it. Instead, the correct way of doing "presets" is context (`SWRConfig`). Imagine this API:

```js
<SWRConfig value={{ initialData: { data: 1 } }}>
  <Component/>
</SWRConfig>
```

The context provides a default cache object, that all its children can read from or override to. The beauty is it doesn't add any extra logic to the SWR hook.

With the new cache interface (thanks @sergiodxa), I think we should even simplify it to:

```
<SWRConfig value={{ cache: initialCache }}>
  <Component/>
</SWRConfig>
```

So we can have namespaced cache and initial data at the same time, and it will be extremely friendly for SSR/SSG as well.

yuritoledo (approved):

nandorojo:
@shuding That makes sense to me. However, I'm not sure how it will fit into Next's SSG patterns. Since static generation is per-page, but the `SWRConfig` goes in the root of the app, how would statically-generated data get passed to `SWRConfig`?

I want to share a single page's static data across my Next site, similar to Gatsby's `useStaticQuery`.

Should I call `cache.set('page-data', staticData)` on the page's mount in order to share the static data globally, and forget `initialData` altogether?

Develliot:
I have the same issue, 

I have fundraising page and which uses initial data and I have an other end point for adding donations. when I add a donation I want to update the running total on the fundraising page. I can't just revalidate the page because I have no guarantee that the page endpoint has updated in time I have to mutate it locally. 

I can mutate the page by key when it doesn't exist in cache because I need to know the old page data before sending it modified page data. I'm forced to do something like this: 

```
const FundraisingPage = ({ pageData }: Props) => {
  //  figure out unique page url for page cacheId
  const router = useRouter();
  const { slug } = router.query;
  const slugString = queryAsString(slug);
  const url = `${createFundraisingPageUrl({ slug: slugString })}`;

  // page cache
  const opts = { initialData: pageData };
  const { data, error, mutate } = useSwrGet(url?.length ? url : null, opts);
  // manually setting cache as this doesn't seem to happen with swr for ssr only on the browser
  useEffect(() => {
    if (data) {
      cache.set(url || "", data);
    }
  }, []); 

.....
```

nandorojo:
I might be reading this incorrectly, but 2 thoughts:

1) If you're using query parameters in `router.query`, rather than dynamic routes, they won't exist on the server. This is why it's not working for SSR. You should be using get server side props from Next.js if you want the initial server load to always be up to date (I think, I've only used `getStaticProps`.) Also, use dynamic routes instead of query parameters.

2) Your server should return to you only after the money has posted to your database. I personally wouldn't optimistically render for something as sensitive as dollar amounts, but that's your call. In any case, revalidating should be safe by the time you've POSTed your data.

A pattern I like: POST data to the server, it returns the mutated data, and then inject that into SWR's cache.

mthines:
@aj-may 

I ended up checking if an identical URL and request body already has been sent, then settings `revalidateOnMount` to false.

Then it fetches only onces, and then only use the cache if it exists.

```tsx
  const response = useSWR<Data, Error>(
    apiUrl && token ? [apiUrl, 'GET', reqData] : null,
    {
      // If we already have cached results, we don't want to call the endpoint again.
      // If we don't have cached results, then call the end point
      revalidateOnMount: !cache.keys().some((k) => apiUrl && k.includes(apiUrl) && reqData && k.includes(reqData)),
      revalidateOnFocus: false,
      revalidateOnReconnect: false,
      refreshWhenOffline: false,
      refreshWhenHidden: false,
    },
  );
```

Not sure if it fixes your request, but it worked for me.

Consider looking at [`swr-sync-storage`](https://github.com/sergiodxa/swr-sync-storage) if you want to persist the data on page refresh :)

shuding:
#1017 is going to cover this very same use case in a more general way, which we are aiming to launch soon. 🎉

shuding:
Closing this PR since #1017 was landed already. Docs: https://swr.vercel.app/advanced/cache.

PR #391, opened 2020-05-13
Make trigger return a promise resolved to the new cache value

sergiodxa:
Right now `mutate` is typed to return a Promise with the updated value, but if you don't pass a second argument `mutate` calls `trigger` that was not retuning anything.

This PR changes `trigger` to return a promise always, if it didn't received a key it's an immediately resolved promise, if it has key and there are updaters for that key, it uses Promise.all to await for all of them to be resolved and then returns the new cached value, lastly If there are no updated it will return an immediately resolved promise with the current cached value.

This should allow us to do:

```ts
import { mutate } from 'swr'

// do something
await mutate(key)
// do more things after the cache key was revalidated
```

This fix https://github.com/zeit/swr/issues/355

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -58,7 +58,7 @@ export type updaterInterface<Data = any, Error = any> = (
 export type triggerInterface = (
   key: keyInterface,
   shouldRevalidate?: boolean
-) => void
+) => Promise<any>
 type mutateCallback<Data = any> = (currentValue: Data) => Promise<Data>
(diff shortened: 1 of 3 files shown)


outaTiME:
ETA?

shuding:
@sergiodxa thank you for this PR! I think we should do the same here (inside `mutate`): https://github.com/vercel/swr/blob/master/src/use-swr.ts#L124

PR #404, opened 2020-05-19
Refresh state when change pageKey

### LINKED ISSUE #396: useSwrPages doesn't work with filters (dynamic URLs)
If you have a table with filters and your query depends on the filters, useSwrPages doesn't perform well and starts to duplicate results.

Here's the use case:
1 - Setup a useSwrPages to fetch from an api that returns the next offset number (skip in mongo terms) and "null" if it knows there are no more skips (end of data).
2 - Make your page's SWR depend on some filters that are added as a query string to the URL and add the filters to the dependencies.
3 - Load 3 or 4 pages without filters
4 - Apply filters (which will change the query string).
5 - Each SWR will fetch again with the filters and when there are not enough results, the offsetter function will return null (because API returned null). However, because there are 4 pages already, SWR will run them anyway with null as offset and retrieve the same set of results duplicated in the 4 pages.

Trying to pass null to the swr so that it doesn't fetch again won't work because "null" is the initial value of the offset and the page has no way of knowing if it's the first or not.

A way to overcome this is to not restrict us with null values or expose a method `cleanup` to remove all the pages. We can call `cleanup` after `filters` has changed to make useSwrPages restart from page 0.

ykzts:
Even if the pageKey is changed, it does not work properly because it is using at the previous state. Refreshes the state when the pageKey changes.

Resolve #396

### DIFF
--- a/src/use-swr-pages.tsx
+++ b/src/use-swr-pages.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useState, useRef } from 'react'
+import React, { useCallback, useMemo, useState, useRef, useEffect } from 'react'
 
 import { cache } from './config'
 import {
@@ -98,6 +98,7 @@ export function useSWRPages<OffsetType = any, Data = any, Error = any>(
 ): pagesResponseInterface<Data, Error> {
   const pageCountKey = `_swr_page_count_` + pageKey
   const pageOffsetKey = `_swr_page_offset_` + pageKey
+  const pageSWRsKey = `_swr_page_swrs_` + pageKey
 
   const [pageCount, setPageCount] = useState<number>(
     cache.get(pageCountKey) || 1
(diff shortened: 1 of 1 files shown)


CurtisBaldwinson:
This PR works a bit better than https://github.com/vercel/swr/pull/269

The other PR has an occasional data duplication stutter still, and this PR accomplishes the full fix with just a few lines. Awesome job! I compiled and tried both PRs and this one is a keeper!

src/use-swr-pages.tsx
```diff
@@ -98,14 +98,27 @@ export function useSWRPages<OffsetType = any, Data = any, Error = any>(
...
 
   const [pageCount, setPageCount] = useState<number>(
     cache.get(pageCountKey) || 1
   )
   const [pageOffsets, setPageOffsets] = useState<OffsetType[]>(
     cache.get(pageOffsetKey) || [null]
   )
-  const [pageSWRs, setPageSWRs] = useState<responseInterface<Data, Error>[]>([])
+  const [pageSWRs, setPageSWRs] = useState<responseInterface<Data, Error>[]>(
+    cache.get(pageSWRsKey) || []
```
neilgupta, line 110: pageSWR's aren't cached currently. We could either just always use an empty array, or cache pageSWRs on line 176.

I tried caching pageSWRs, but this seems to cause a significant delay when loading a pageKey that is cached. On the other hand, if we don't cache pageSWRs, then we lose the scroll position on a page when going back. I haven't found a good solution for the hanging yet.

src/use-swr-pages.tsx
```diff
@@ -98,14 +98,27 @@ export function useSWRPages<OffsetType = any, Data = any, Error = any>(
...
+
+  useEffect(() => {
+    setPageCount(cache.get(pageCountKey) || 1)
+  }, [pageCountKey])
+  useEffect(() => {
+    setPageOffsets(cache.get(pageOffsetKey) || [null])
+  }, [pageOffsetKey])
+  useEffect(() => {
+    setPageSWRs(cache.get(pageSWRsKey) || [])
+  }, [pageSWRsKey])
```
neilgupta, line 121: Is there an advantage to creating multiple effect hooks, versus a single effect that resets all 3 states and depends on pageKey?

Also should this be a layoutEffect to prevent multiple updates when pageKey changes?

shuding:
Hi all, thank you for your effort on this! We've collected a lot of feedback on the current pages API, and we've decided to redesign it from scratch.

It's almost finished in #435. Please let us know how do you feel about the new API, so we can have better cover of all these scenarios.

ykzts:
@shuding thx! I would like to check the RFC.

PR #427, opened 2020-06-05
Fix mutate not disposing older async mutations

### LINKED ISSUE #425: Mutating toggle on/off optimistic causes flash of old data
Imagine the following scenario where a user can optimisticly toggle a checkbox on and off:
```javascript
const { data, mutate } = useSWR("/api/toggle", fetch);

const toggle = (e) => {
   const isToggled = e.target.checked:

   // mutate optimisticly without revalidation
   mutate(isToggled , false);

   // execute POST and update without revalidation
   mutate(
        fetch("/api/toggle", {
            method: "POST",
            headers: {
                "Content-Type": "application/json",
            },
            body: JSON.stringify({isToggled}),
        }),
        false
    );
}

return <Checkbox onChange={toggle} checked={data} />
```

The problem is that when a user clicks the checkbox (or switch) twice in rapid succession, he'll see a flash of stale data:
checkbox

Starting state: checkbox is toggled off
1. User clicks checkbox 
    - checkbox is toggled on
    - POST request to toggle on is fired
1. User clicks checkbox again
    - checkbox is toggled off
    - POST request to toggle off is fired
1. POST request of step 1 returns
   - checkbox is incorrectly turned on
1. POST request of step 2 returns
   - checkbox is turned off

The problem is at step 3, where the first POST request returns with stale data. I'd expect SWR to cancel any old ongoing mutates when calling new mutates?

TheHolyWaffle:
Fixes #425

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -81,6 +81,10 @@ const mutate: mutateInterface = async (
 
   let data, error
 
+  // Keep track of timestamps before await asynchronously
+  const beforeMutationTs = MUTATION_TS[key]
+  const beforeConcurrentPromisesTs = CONCURRENT_PROMISES_TS[key]
+
   if (_data && typeof _data === 'function') {
     // `_data` is a function, call it passing current cache value
     try {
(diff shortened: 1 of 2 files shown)


TheHolyWaffle:
Ready for review @shuding

Issue #434, opened 2020-06-08
`deps` in `useSWRPages` doesn't seem to do anything


timwachsener:
I have a paginated list that can be filtered. So I thought putting the filter string in `deps` would help to get the pages refetched when it changes - but it doesn't. Also including the filter in the `pageKey` didn't help.

```
useSWRPages(
  `items-${id}`,
  ({ offset, withSWR }) => { ... },
  ({ data }) => { ... },
  [filter]
)
```

shuding:
The new pagination API has launched: https://swr.vercel.app/docs/pagination. Feel free to reopen if you are still encountering an issue!

Issue #399, opened 2020-05-16
How to use useSWRPages with graphql?


jackyang1984:
Anyone knows how to use useSWRPages with graphql, the examples just use rest api.

w3bdesign:
Here's a simple example. 

It's not the best code or the most functional code, but it should at least get you started. 

Obviously you need to change GRAPHQL_ENDPOINT_URL.

I am using graphql-request for GraphQL.

```
const App = () => {
  const FETCH_ALL_CATEGORIES_QUERY = `
query MyQuery {
  productCategories      {
    nodes {
      name
    }
  }
}
`;

  const { data, error } = useSWR(FETCH_ALL_CATEGORIES_QUERY, (query) =>
    request('GRAPHQL_ENDPOINT_URL', query)
  );

  console.log(data); // TODO pass this to a component or do something with it
  console.log(error); // TODO do error handling

  return <p>Example</p>;
};
```

jackyang1984:
@w3bdesign  thanks a lot, i want to know how to use useSWRPages , not useSWR.

shuding:
Hi @jackyang1984 please take a look at our new API in #435

shuding:
The new pagination API has launched: https://swr.vercel.app/docs/pagination. Feel free to reopen if you are still encountering an issue!

Issue #384, opened 2020-05-10
Dependent fetching with paging not working


GustavPS:
```javascript
const { pages, error } = useSWR(validated ? `http://${server.server_ip}:4000/api/movies/list` : null, fetcher)
````
The above line work, the request is being run after validated get the value true.

The block below does not work, the request is never runned, even when validated gets equal to true
```javascript
   const { pages, isLoadingMore, loadMore } = useSWRPages(
        "movie-list",
        ({ offset, withSWR }) => {
          const url = offset || `http://${server.server_ip}:4000/api/movies/list`;
          const { data } = withSWR(useSWR(validated ? url : null, fetcher));
          console.log(data);
          if (!data) return null;
    
          const  results = data;
          console.log("results");
          console.log(results);
          return results.data.map(result => (
            <h1>Test</h1>
          ));
        },
        SWR => SWR.data.next,
        []
      ); 
```

illuminist:
Sounds like you need to put `validated` into deps array as well.

GustavPS:
> Sounds like you need to put `validated` into deps array as well.
> 
> Edit: Addition to need deps array, I dug into source code a bit, I just found something wrong in useSWRPages, will investigate further.

Thanks, but what exactly is the deps array?

illuminist:
> Thanks, but what exactly is the deps array?

It's the same thing as in useEffect, useMemo, useCallback, etc hook. It's use for the hook to reevaluate its data or run it's callback when some value in dependency array is difference from the last render.

shuding:
The new pagination API has launched: https://swr.vercel.app/docs/pagination. Feel free to reopen if you are still encountering an issue!

PR #435, opened 2020-06-08
New infinite loading API

shuding:
This PR adds a new `useSWRInfinite` API for infinite loading, which is far more stable and easier to understand and use:

```js
const { data, error, mutate, size, setSize } = useSWRInfinite(getKey, fetcher?, options?)
```

The new API signature is very similar to `useSWR`. One difference is that the first parameter `getKey` is a function that returns the key:

```js
(currentPageIndex, previousPageData) => keyOfThisPage
```

### API
The returned object includes `size` and `setSize`, they are returned by an internal state inside the hook.

- `data`: an array of paginated fetch response values.
- `mutate`: same as useSWR's bound mutate function[1]
- `options`: all current SWR options are supported, with 2 new options:
  - `initialSize`: number of how many pages should be loaded initially, default to `1`
  - `revalidateAll`: should we always try to revalidate all pages[2], default to `false`

[1]: note that the global `mutate` won't work with this new API
[2]: `useSWRInfinite` supports all the revalidation options of SWR too (`refreshInterval`, `revalidateOnFocus`, `revalidateOnReconnect`...). However by default it will only revalidate the **first page**. You can change the behavior to revalidate all pages by setting `revalidateAll` to true.

### Why Renaming
There're a lot of confusion about `useSWRPages` that people are using it for *normal* pagination use cases. However that's not necessary and you should implement it with [page components](https://twitter.com/shuding_/status/1250912534745341952/photo/1) or dynamic SWR keys.

`useSWRInfinite` should only be used when you need dynamic number of fetch requests (a list that you can load more and append data to), and the key of each request depends on the previous data.

### Example 1: Paginated API

```
GET /users?page=0&limit=10
[
  { name: 'Alice' },
  { name: 'Bob' },
  { name: 'Cathy' },
  ...
]
```

With `useSWRInfinite`:
```js
const { data } = useSWRInfinite(
  (index, previousPageData) => {
    // reached the end
    if (previousPageData && previousPageData.length === 0) return null

    // API key
    return `/users?page=${index}&limit=10`
  },
  fetcher, // same as useSWR
  opts     // same as useSWR
)

// render
if (!data) return 'loading'
return <>
  {data.map((users, index) => {
    // render each page
    return users.map(user => <div>{user.name}</div>)
  })}
</>
```

### Example 2: Cursor Based Paginated API

```
GET /users?cursor=123&limit=10
{
  data: [
    { name: 'Alice' },
    { name: 'Bob' },
    { name: 'Cathy' },
    ...
  ],
  nextCursor: 456
}
```

With `useSWRInfinite`:
```js
const { data } = useSWRInfinite(
  (index, previousPageData) => {
    // reached the end
    if (previousPageData && previousPageData.data.length === 0) return null

    // first page, previousPageData is null
    if (index === 0) return `/users?limit=10`

    return `/users?cursor=${previousPageData.nextCursor}&limit=10`
  },
  fetcher, // same as useSWR
  opts.    // same as useSWR
)
```

### Example 3: Advanced Features

Here's an example showing how you can use this API to implement:
- loadMore
- isLoadingMore
- isLoadingInitialData
- isEmpty
- isReachingEnd
- change data source
- refresh the entire list

https://codesandbox.io/s/swr-infinite-z6r0r?file=/src/App.js

### Checklist
- [x] cache page count
- [x] scroll position restoration
- [x] #391
- [x] make `setSize` return promise
- [ ] update documentation
- [ ] create examples
- [x] deprecate `useSWRPages`

Latest release of this PR: `swr@0.3.0-beta.6`.

---

Related to #21, #70, #126, #133, #178, #189, #202, #205, #243, #246, #269, #289, #309, #311, #343, #347, #384, #396, #399, #434 (almost all our existing issues are about the undocumented pagination API 🙊).


timwachsener:
I tried this as a replacement for `useSWRPages` in a Next.js application and found some things:
- when loading more pages every previous page is always revalidated
- when navigating to a different Next.js page (but using the same component, for example another category on a blog) the page count gets mixed up: load up to page 4 in category A, switch to category B and load page 2, switch back to category A and try to load page 5 => page 3 is loaded on category A
- when the key changes the page count doesn't get reset. For example when changing a filter variable my key gets recomputed but the page count stays the same resulting in duplicated list items

I use `useSWRInfinite` as follows with a GraphQL server:
```
export const usePages = (key, query, variables, config) => {
  return useSWRInfinite(
    (index, previousPageData) => {
      const hashed = toString([key, print(query), variables]);

      if (index === 0) {
        return [hashed];
      }

      if (!previousPageData.items?.pageInfo?.hasNextPage) {
        return null;
      }

      return [hashed, previousPageData.items.pageInfo.endCursor];
    },
    (_, cursor) => {
      return request(query, { ...variables, cursor });
    },
    {
      initialData: config.initialData.items ? [config.initialData] : null,
      revalidateOnFocus: false,
    }
  );
};
```

shuding:
Thanks for the comment @timwachsener.
1) Should be a bug, will get it fixed.
2) I will think more about that scenario, but using the exact same key for both lists (inside difference components) will mix up the cache for sure. Maybe you can use an extra identifier `['extra-id', api]` as the key, so they will be considered as different data sources.
3) Will be fixed as well. Currently page count is just a state and not cached. It should be cached and bound to the key.

timwachsener:
> 2. I will think more about that scenario, but using the exact same key for both lists (inside difference components) will mix up the cache for sure. Maybe you can use an extra identifier `['extra-id', api]` as the key, so they will be considered as different data sources.

That's exactly what I'm doing. It's the other way round: I'm using different keys as you suggested when rendering the same component with different props.
I guess this may be resolved with 3. when the page count is bound to the key.

src/use-swr-infinite.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+      // return the data
+      return data
+    },
+    extraConfig
+  )
+
+  // extend the SWR API
+  const mutate = swr.mutate
+  swr.mutate = (data, shouldRevalidate = true) => {
```
sergiodxa, line 159: Should this be memoized?

src/use-swr-infinite.tsx
```diff
@@ -0,0 +1,185 @@
...
+      cache.set(contextCacheKey, { originalData, force: false })
+    } else if (shouldRevalidate) {
+      // calling `mutate()`, we revalidate all pages
+      cache.set(contextCacheKey, { force: true })
+    }
+
+    return mutate(data, shouldRevalidate)
+  }
+  swr.page = page
+  swr.setPage = arg => {
```
sergiodxa, line 172: Should this be memoized

test/use-swr-infinite.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+  render,
+  waitForDomChange,
+  fireEvent,
+  act
+} from '@testing-library/react'
+
+import { useSWRInfinite } from '../src'
+
+describe('useSWRInfinite', () => {
+  afterEach(cleanup)
```
sergiodxa, line 13: RTL setup this automatically, it should not be required

src/use-swr-infinite.tsx
shuding: Thanks! Should be fixed

shuding:
@timwachsener the issues addressed (1. and 3.) should be all fixed now in `swr@0.3.0-beta.1`.

Jack-Works:
Hi, the new API is cool and easy to use. But I want to request some more features so this hook can easily integrated with the infinite scrolling library in the React ecosystem.

For example https://github.com/ankeetmaini/react-infinite-scroll-component require a "load more" function that return a Promise. So if the infinite scrolling is near to the end, the framework will call the load more function to load more pages.

shuding:
Hi @Jack-Works! Yeah that's a good idea.

It can be easily supported once we ship https://github.com/vercel/swr/pull/391 (thanks to @sergiodxa):

```js
setPage(page + 1).then(...)
```

Jack-Works:
If this hook can return a wrapped function to do setPage(...) with the proper page, it will easier to integrate with those infinite scrolling libraries.

shuding:
@Jack-Works do you have a more derailed example? Currently you can do this with the lib I think:

```js
const { setPage } = useSWRInfinite(...)
const loadNext = () => setPage(x => x + 1)

return <InfiniteScroll next={loadNext}>...</InfiniteScroll>
```

Jack-Works:
https://github.com/vercel/swr/pull/438
@shuding I made an example for this hook, but my custom fetcher seems never to be called.

https://github.com/vercel/swr/pull/438/files#diff-a684bd3ae2b27ac9dcc73d7b0aeea3e7R9

src/types.ts
```diff
@@ -47,8 +47,9 @@ export interface RevalidateOptionInterface {
   dedupe?: boolean
 }
 
-type keyFunction = () => string | any[] | null
-export type keyInterface = keyFunction | string | any[] | null
+export type keyType = string | any[] | null
```
Jack-Works, line 50:
```suggestion
export type keyType = string | number | string[] | number[] | null
```

shuding: I think this is okay. A cache **key** should be required to be a string, but if you pass a number it's okay and will be converted.

Jack-Works:
it seems lacking a "isReachingEnd"

src/types.ts
Jack-Works: I'm having bugs with `key === 0` in my previous example 🙈 maybe swr is using `!` to check?

shuding:
@Jack-Works did you check the 3rd example (https://codesandbox.io/s/swr-infinite-z6r0r?file=/src/App.js)?

src/types.ts
shuding: Yep SWR is using `!` and `0` is falsy.

timwachsener:
> @timwachsener the issues addressed (1. and 3.) should be all fixed now in `swr@0.3.0-beta.2`.

Looks great on first try. I will check this out in detail later.

TodorTotev:
Works great for me

Jack-Works:
> @Jack-Works did you check the 3rd example (https://codesandbox.io/s/swr-infinite-z6r0r?file=/src/App.js)?

Cool, but if we can calculate those `isLoadingInitialData`, `isLoadingMore`, `isEmpty`, `isReachingEnd` for developers that will be grate!

shuding:
@Jack-Works yeah that would be ideal, but it's impossible to implement internally, because we don't know the shape of the returned data. It can be an array (so `isEmpty` is true when the array is empty), but it can be an object too. So overall I think we should provide basic examples and let users to implement these themselves.

src/use-swr-infinite.tsx
```diff
@@ -0,0 +1,200 @@
...
+): ExtendedResponseInterface<Data, Error>
+function useSWRInfinite<Data = any, Error = any>(
+  getKey: KeyLoader,
+  config?: ExtendedConfigInterface<Data, Error>
+): ExtendedResponseInterface<Data, Error>
+function useSWRInfinite<Data = any, Error = any>(
+  getKey: KeyLoader,
+  fn?: fetcherFn<Data>,
+  config?: ExtendedConfigInterface<Data, Error>
+): ExtendedResponseInterface<Data, Error>
```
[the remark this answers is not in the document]
shuding: Good catch, thank you!

TheHolyWaffle:
How would the cursor based pagination example work? `useSWRInfinite` only exposes an `setPage` function, which is supposed to be called with an integer as far as I can see.

The use-case I have in mind is a list of blog posts, where a user can use the buttons "Next" and "Back" to navigate through the blogs. Regular index based pagination doesn't work, since new blog posts can be added or removed causing inconsistencies in the lists. So going back and forward is relative to the current list of blogs visible to the user.

I'd expect `useSWRInfinite` to expose something like `setCursorNext` and `setCursorBack`. That cursor can then be used in the `getKey` to build a key such as `/api/blogs?next=${cursorNext}` and `/api/blogs?next=${cursorBack}`

Right now, it appears that the `getKey` function only accepts currentPageIndex number.

shuding:
@TheHolyWaffle If i understand correctly, you need the normal pagination implementation (because you have "Next" and "Back" actions), not **infinite loading**. So you might not need `useSWRInfinite` at all.

I'd do something like this:

```js
function App () {
  const [previous, setPrevious] = useState(null)
  const [next, setNext] = useState(null)
  const { data } = useSWR(`/api/blogs?${
    previous ? `previous=${previous}` :  next ? `next=${next}` : ''
  }`)

  if (!data) return 'loading'

  return <>
    {data.map(post => <a key={post.id}>{post.title}</a>)}
    <button onClick={() => {
      setNext(null)
      setPrevious(data[0].cursor)
    }}>Back</button>
    <button onClick={() => {
      setPrevious(null)
      setNext(data[data.length - 1].cursor)
    }}>Next</button>
  </>
}
```

Ligerx:
Is the `getKey` function only meant to return a string? I attempted to return an array and it wouldn't work properly.

o-alexandrov:
Great work, comparing to the `useSWRPages` the new hook, `useSWRInfinite` doesn't require much thinking after getting used to `useSWR`.

Btw, you are probably aware, but just fyi, the bounded `mutate` doesn't work in the provided example:
> @Jack-Works did you check the 3rd example (https://codesandbox.io/s/swr-infinite-z6r0r?file=/src/App.js)?

`mutate` doesn't seem to work at all there, even when passing the first argument `[]` to reset the pages.

shuding:
@Ligerx Yes it supports returning an array, what issue did you have and can you share me a reproduction?

shuding:
@o-alexandrov in the CodeSandbox demo `mutate` does work for me: click "Refresh" and you can see in the network tab that the revalidation requests are actually fired.

screenshot-wiZ1IxAa@2x

Ligerx:
@shuding Oh turns out I was using the fetcher api incorrectly. I expected it to return me an array instead if destructuring the array into individual parameters.

martonlanga:
Shouldn't `page` default to `0` here? `const { data, error, page /* === 1 */, setPage } = useSWRInfinite`

The `index` in the `getKey` starts with `0`, so it was a bit confusing for me.

Otherwise it's pretty straightforward and easy to use!

jaybytez:
We were having the same issue as #243 for swr pages, so we are looking to swr infinite.  It looks like the swr infinite implementation resolves this but it might be because we are unable to get the scroll position restoration working.  Does this scroll position restoration work the same in swr infinite?

shuding:
Thanks for the feedback @martonlanga! `page` means the number of pages being fetched (the length of the pages array), so it's different from the `index`. 

I see the confusion here. I think renaming them to `length` and `setLength` might be a better idea.

shuding:
@jaybytez scroll position restoration does work the same. Did you change anything else?

neevannv:
I've implemented useSWRInfinite with revalidateAllPages to fetch more data when scrolled, everything looks good but previously generated requests on scroll process are also revalidating, thought it would revalidate all when navigate to other page and come back, any thoughts on how it works.

Here is the code used
const { data, error, page, setPage } = useSWRInfinite(
    (index, prevData) => {
      let url = null;
      let offset = 0;
      // If no more results to fetch stop callling request
      if (prevData && !!prevData.hasMore === false) {
        return url;
      }
      // Set offset value based on previous response
      if (prevData && !!prevData.hasMore) {
        offset = prevData.offset + prevData.count;
      }
      // Set search value condition
      let searchFilterCond = '';
      if (searchValue) {
        searchFilterCond = `&default=${encodeURIComponent(searchValue)}`;
      }
      // API
      url = `${constants.api}/items?offset=${offset}&limit=${limit}&links=none&fields=all&q=${searchFilterCond}`;
      return url;
    },

Attached screenshot shows requests generated when scrolling
swr-infinity-issue

shuding:
@neevannv I think something else is wrong. Can you share the code that you are using to load new pages when scrolling?

By default SWRInfinite only revalidates the **first** page (e.g.: when focusing or remounting (back from another page)). When loading new pages (e.g.: when `setPage(n => n + 1)`) it only revalidates the **new** page.

neevannv:
@shuding 
  const handleScroll = () => {
    const scrollPos = document.body.scrollTop || document.documentElement.scrollTop;
    const scrollPercent = (100 * scrollPos) / (document.body.scrollHeight - window.innerHeight);
    // Fetch more stories if scroll reached morethan 50% of page height
    // console.log(isLoadingMore);
    if (scrollPercent > 50 && !isLoadingMore && scrollPos > lastScrollPos) {
      setPage(page + 1);
    }
    lastScrollPos = scrollPos;
    window.sessionStorage.setItem(scrollKey, lastScrollPos);
  };

neevannv:
@shuding 
Tried setPage like below also, but no use
setPage((p) => p + 1);

shuding:
@neevannv I'm not able to reproduce still. Check [this example](https://codesandbox.io/s/swr-infinite-z6r0r?file=/src/App.js), it only fetches new pages. Do you mind creating a reproducible demo on CodeSandbox?

src/use-swr-infinite.ts
```diff
@@ -0,0 +1,200 @@
...
+    contextCacheKey = 'context@' + firstPageKey
+  }
+
+  // page count is cached as well, so when navigating the list can be restored
+  let pageCountCacheKey: string | null = null
+  if (firstPageKey) {
+    pageCountCacheKey = 'size@' + firstPageKey
+    initialSize = cache.get(pageCountCacheKey) || initialSize
+  }
+  const pageCountRef = useRef<number>(initialSize)
```
neilgupta, line 105: It'd be nice if the page count updated automatically when the key changes, rather than requiring the user to reset the count to 0.

```javascript
if (firstPageKey) {
  pageCountCacheKey = 'size@' + firstPageKey
}
const pageCountRef = useRef<number>(initialSize)
useEffect(() => {
  pageCountRef.current = cache.get(pageCountCacheKey) || initialSize
}, [pageCountCacheKey, initialSize])
```

shuding:
@martonlanga I updated the API to use `size` and `setSize`, which I think are better than "length" or "page".

src/use-swr-infinite.ts
shuding: Thanks for the feedback @neilgupta! Yeah I agree that in some cases it helps, but I don't think it should be the default behavior. Consider this scenario:

Group 1

Say I have a list with a "show advanced" checkbox. When enabled the API should have a `?advanced=1` query attached. So when toggling the checkbox, I don't want the page count to be reset to 0. 

Therefore, if we make it the default behavior, it's gonna be hard for the user to bypass it. They'll have to memorize the page count before changing the key. So I think it's acceptable to manually reset the page size if we need to change the data source.

neevannv:
@shuding I create project on codesandbox.io for some reason swr not installed properly.
https://codesandbox.io/s/nextjs-revalidate-all-pages-issue-t998w?file=/pages/index.js
Attached complete code can reproduce issue.
[swr-scroll-issue.zip](https://github.com/vercel/swr/files/4813480/swr-scroll-issue.zip)

src/use-swr-infinite.ts
balthazar: An option might be nice for this, as I'm not sure loading more than one page when the query changes would be the behavior people expect, even if doing a `setSize` in an `useEffect` isn't that big of a deal, it could be a slight improvement

shuding:
@munone-bot Yes you can still use `mutate` with the data array. You can modify any item in the array, and the hook will shallowly diff it with the original data array and revalidate the indexes that has been updated.

neevannv:
> @shuding I create project on codesandbox.io for some reason swr not installed properly.
> https://codesandbox.io/s/nextjs-revalidate-all-pages-issue-t998w?file=/pages/index.js
> Attached complete code, can reproduce the issue.
> [swr-scroll-issue.zip](https://github.com/vercel/swr/files/4813480/swr-scroll-issue.zip)

Any luck?

neevannv:
Is there any way to get initial data as previous data in first request call when we define initial data.

o-alexandrov:
@munone-bot you probably want to push to the end of the array your `newData` object to the `updatedData` array, or to insert specifying an index `updatedData.splice(indexWhereToPut, 0, newData)`

neevannv:
[swr-scroll-issue.zip](https://github.com/vercel/swr/files/4831998/swr-scroll-issue.zip)
@shuding I create project on codesandbox.io for some reason swr not installed properly.
https://codesandbox.io/s/nextjs-revalidate-all-pages-issue-t998w?file=/pages/index.js
Attached complete code, can reproduce the issue.
swr-scroll-issue.zip

shuding:
Hi @neevannv I just ran your code locally, and updated SWR to the latest beta `swr@0.3.0-beta.6`. `page` and `setPage` are now `size` and `setSize`.

> previously generated requests on scroll process are also revalidating

The reason of this is that you have `revalidateAllPages: true`. So every time you update the page count, previous pages will be revalidated. It's the documented behavior.

shuding:
> @munone-bot you probably want to push to the end of the array your `newData` object to the `updatedData` array, or to insert specifying an index `updatedData.splice(indexWhereToPut, 0, newData)`

Yep, and you also need to update the page size via `setSize(n => n + 1)`.

hems:
Is there an option to provide the `initialData`?

I can't find an example of the new API where i can provide the `initialData` which i currently use since the first page on my infinite loading is SSR ?

hems:
Also i can't find anything on the README about the new API?

timwachsener:
> Is there an option to provide the `initialData`?
> 
> I can't find an example of the new API where i can provide the `initialData` which i currently use since the first page on my infinite loading is SSR ?

You can simply pass initialData as you would with just `useSWR`. But make sure to wrap it in an array so it resembles the value `useSWRInfinite` would return.

shuding:
@hems We will add docs of this new API on the README once it's finalized and released in the new major version. :+1:

src/use-swr-infinite.ts
shuding: Thanks again for the suggestion @neilgupta and @balthazar!

I thought more about this and decided to make it a default behavior but also provide an option. That said, page size will be reset automatically and you can also make it persistent. PR: https://github.com/vercel/swr/pull/490.

hoangvvo:
I followed Example 2 but could not get my code to work. Can anyone help me debug if they have the time.

Here is the relevant code: https://github.com/hoangvvo/nextjs-mongodb-app/blob/8f189ce19ebecdc5eef7a9df0254c80078cf3a45/components/post/posts.jsx#L48

```
```

Basically my cursor would be `createdAt` value of the oldest posts to use as `from` parameter. However, when I try to `console.log`, `index` is always `0` and `previousPageData` is always `null` even after calling `setSize(size + 1)`

jackguoAtJogg:
Is there a way that I can mutate the data in different component? I tried used the same way as `useSwr` but doesn't seem to be working.

sergiodxa:
@jackguoAtJogg you can do
```js
import { mutate } from "swr"
mutate(key, newData)
```

Develliot:
I have this working with mutations quite nicely and the advanced example helped a lot.
I didn't use the code for 'isReachingEnd' because there are some scenarios where it's not going to work

`
  const isReachingEnd =
    isEmpty || (data && data[data.length - 1]?.length < PAGE_SIZE);
`
It's possible to be on the last page and the number of page items on that page could be equal to the page size. You wouldn't want to use this to hide the load more button because it might never disappear but each click would keep bumping up the setSize value. 

I don't know what the solution is because we don't know what the 'total items' number is. 
My rest API luckily containers meta data that I can use but this might not be the case for everyone.

Develliot:
This uses a bound mutate function, but if I wanted to mutate this from the outside using 
`import { mutate } from "swr"`

how do I know that the key is for the the useSWRInfinite instance?

shuding:
@Develliot Yep I think that's too API specific. We can't have a built-in solution for everyone. As for the "last page" case, what we are doing is to always fetch 1 more item, but strip it when returning the data.

hems:
> how do I know that the key is for the the useSWRInfinite instance?

IIRC useSWR `key` is an array with all the parameters that were passed to the `fetcher`.

For ihstance if your useSWR is calling fetcher with: `['/api/fetchItems', 0, 100] ` then your key will actually be `['/api/fetchItems', 0, 100]`

shuding:
Right now, the key of the `useSWRInfinite` hook is `['many', keyOfFirstPage]`. The data is an array.

But keep in mind that it's an internal API and might change in the future.

hems:
> Right now, the key of the `useSWRInfinite` hook is `['many', keyOfFirstPage]`. The data is an array.
> 
> But keep in mind that it's an internal API and might change in the future.

shouldn't it be consistent to the simple `useSWR` and basically always be an array with the same parameters that are being passed to the `fetcher`? which i believe is the result of your `getKey` function ?

shuding:
@hems However `useSWRInfinite` is not just **one** request. It contains multiple request calls to `fetcher`, which one should we use as its key?

Also, people can use the keys of each page of `useSWRInfinite` in normal `useSWR` hooks as well, so there must be some special “namespace” to avoid conflicts.

Develliot:
@shuding One solution might be to allow the user to pass a custom 'listName' and this gets used as the key and it falls back to ['many', keyOfFirstPage] if not specified. 

```
useSWRInfinite(
    url,
    fetcher,
    ops,
    listName,
  )
```

Functions with more than 3 params makes me shudder, may be it could live in ops. 

The main benefit is that the user doesn't have to know what the code looks like. I can think of quite a few use cases where this would be useful

hems:
> @hems However `useSWRInfinite` is not just **one** request. It contains multiple request calls to `fetcher`, which one should we use as its key?

Every request from the useSWRInfinite would be unique because the "offset" and "perPage" parameter has to change for each request? Let's suppose:

````js
// fetcher.js

module.exports = (url, page, perPage) => request.get(`${url}?page=${page}&perPage=${perPage}`)
````

If you use `useSWR` under the hood the keys would be:

 - first page `['https://api', 0, 100]`
 - second page `['https://api', 1, 100]`
 - third page `['https://api', 2, 100]`

At least that's my understanding of how `useSWR` works and how I deal with mutating keys on the project i'm currently working.

But please let me know if I am missing something?

**Comment by hems** [comment]
@shuding I'm not 100% sure if I understand your concern.

Are trying to have 1 unique key for the whole bundle of requests?

I think it's more transparent / easier to have one key per page, like if you simply have done several `useSWR` manually yourself?

Then you could simply use the regular `mutate` function, using the regular `cache` from `swr`, etcs?


````js
import { mutate } from 'swr'

const key = ["my_api_address", page, limit]

mutate(key...)
````

Since `useSWR` cache is available via [config.js](https://github.com/vercel/swr/blob/master/src/cache.ts) users are able check which pages are already cached, also would make `useSWRInfinite` feel very native to `useSWR`?


----

Currently, I'm using `useSWRPages` and what it basically does isL for every page send an array of parameters to my `fetcher` function ( i believe because that's how `useSWR` works? ), so if I would update my application to `useSWRInifnite` the `getKey` function would have to return an `Array` ( as opposed to a simple string ) and that's what would be sent to the `fetcher`.

For instance, on my currently Application my `fetcher` actually fetch the data using WebSockets and it's working perfectly with `useSWRPages` the way i just described.

**Comment by Develliot** [comment]
Personally I think when mutating data I don't want to get the cache page by page, that would be a massive ball ache, I think it's much easier to have a mutation that works on the the whole group of pages.

**Comment by shuding** [comment]
I think it's better to pack the resource as a hook:
```js
function useDataInfinite () {
  return useSWRInfinite(getKey, fetcher, opts)
}
```

When rendering the data in Foo.js:
```js
function Foo () {
  const { data } = useDataInfinite()
  // ...
}
```

Mutating the data in another component:
```js
function Bar () {
  const { mutate } = useDataInfinite()

  return <button onClick={() => {
    mutate(data => newData)
  }}>
    mutate
  </button>
}
```

That way you don't need to know what the key is.

Develliot:
That is an interesting solution

Develliot:
That works, it's definitely more lines of code than passing an ID in one place and importing mutate, but it is a clever solution, I like it.

ssurana2:
Hi, was trying to use `useSWRInfinite` but not sure if this is intentional or a bug. The issue is, on focusing the page again, revalidation call only goes for the first page instead of all the pages that have loaded. But when calling `mutate` manually revalidates all the loaded page. It would be great if revalidation on focus also revalidates all the loaded pages instead of only the first page.

pacocoursey:
@ssurana2 From the PR description:

> [2]: useSWRInfinite supports all the revalidation options of SWR too (refreshInterval, revalidateOnFocus, revalidateOnReconnect...). However by default it will **only revalidate the first page**. You can change the behavior to revalidate all pages by setting revalidateAll to true.

Develliot:
Is that so you can revalidate specific pages, why would you only want to revalidate the first page? I guess may be if you add and item and it appears as the first in the list may be.

shuding:
> why would you only want to revalidate the first page?

@Develliot It's a tradeoff. We don't think it's a good practice in the infinite scrolling scenario. It will be expensive to revalidate all pages, and sometimes unnecessary from the UX aspect. We feel that in _most_ of the infinite scrolling cases, such as the Facebook timeline, what will likely happen is that the user focuses on the window, and get a new item jumping in from the top. 

However, we also provide you an option to always revalidate all.

hems:
> I think it's better to pack the resource as a hook:

that sounds good.

Question: will you use useSWR under the hood still? Because if you do then we also get the ability to mutate / interact with the data the "standard way" by using the useSWR cache itself, meaning we would have best of both worlds?

lucabarbosa:
Hey guys, couldn't I use multiple arguments when returning the getKey function? Just like useSWR, **return [key, query];**

```js
 const { data, error, size, setSize } = useSWRInfinite(
    (index) => {
      const query = {
        category: id,
        page: index,
        order,
        orderby: orderBy,
        status: 'publish',
      };

      const key = queryable(query);
      return [key, query];
    },
    fetcher
  );
```

shuding:
@lucabarbosa Yeah you should be able to do that! If you’re seeing it not working, that’s a bug and please feel free to share with us.

lucabarbosa:
@shuding I got it! It wasn’t working because I was putting an object in the return and it was rendering endlessly.

I have one more question. In my case, I am creating a product listing page with a load more button with useSWRInfinite. When I enter the page of a product and then return to the list of products, I am losing all the loaded indexes and having to start loading with the first index. Is it correct?

shuding:
@lucabarbosa Nope, if you are doing a client side transition the indexes will keep, because its cached.

lucabarbosa:
@shuding I couldn't do that. Maybe because I'm using NextJS? I'll do some tests later. I'm fetching data in client side with Next.

ourfamilygithub:
> > @hems However `useSWRInfinite` is not just **one** request. It contains multiple request calls to `fetcher`, which one should we use as its key?
> 
> Every request from the useSWRInfinite would be unique because the "offset" and "perPage" parameter has to change for each request? Let's suppose:
> 
> ```js
> // fetcher.js
> 
> module.exports = (url, page, perPage) => request.get(`${url}?page=${page}&perPage=${perPage}`)
> ```
> 
> If you use `useSWR` under the hood the keys would be:
> 
> * first page `['https://api', 0, 100]`
> * second page `['https://api', 1, 100]`
> * third page `['https://api', 2, 100]`
> 
> At least that's my understanding of how `useSWR` works and how I deal with mutating keys on the project i'm currently working.
> 
> But please let me know if I am missing something?

@hems, @shuding  , did you figure out how to use `useSWRInfinite` in such cases?
 i really didn't understand about pack it as a hook. my problem is how the `fetcher url` should be when you are returning multiple key like this:

```
const fetcher = () =>
    axiosInstance.get(`/user/comments?paginate=true`).then((response) => response.data);

  const { data, error, size, setSize } = useSWRInfinite(
    (index, prevPageData) => {

      if (prevPageData && prevPageData.data.length === 0) return null

      if(index === 0) return '/user/comments?paginate=true';

      return `/user/comments?paginate=true&page=${prevPageData.nextPage}`;
    },
    fetcher
  );

 <button onClick={() => setSize(size + 1)}>catch next page data</button>

```

and it does not work, it only works on `(index === 0)` becasue have same key with fetcher

shuding:
@ourfamilygithub 

You can change your `fetcher` to accept a URL instead of using a fixed value, like this:

```diff
- const fetcher = () =>
-    axiosInstance.get(`/user/comments?paginate=true`).then((response) => response.data);
+ const fetcher = url =>
+    axiosInstance.get(url).then((response) => response.data);

  const { data, error, size, setSize } = useSWRInfinite(
    (index, prevPageData) => {

      if (prevPageData && prevPageData.data.length === 0) return null

      if(index === 0) return '/user/comments?paginate=true';

      return `/user/comments?paginate=true&page=${prevPageData.nextPage}`;
    },
    fetcher
  );

 <button onClick={() => setSize(size + 1)}>catch next page data</button>
```

ourfamilygithub:
@shuding , Thanks it works fine now
i have found something new with `SWRInfinite` , when you are increasing `size` it make a new request to fetch new page data and when you decreasing it read from cache , i guess the problem here is , for instance when u reach to page 9 and decrease size to 8 then u decide to go 9 again it will fetch it again!

shuding:
@ourfamilygithub Yes, I think that's the expected behavior, because SWR will always try to load the fresh data while showing the cache, for the eventual consistency.

derekang:
> Thanks for the comment @timwachsener.
> 
> 1. Should be a bug, will get it fixed.
> 2. I will think more about that scenario, but using the exact same key for both lists (inside difference components) will mix up the cache for sure. Maybe you can use an extra identifier `['extra-id', api]` as the key, so they will be considered as different data sources.
> 3. Will be fixed as well. Currently page count is just a state and not cached. It should be cached and bound to the key.

Hello, 

Just wondering if #1 has been fixed as I am also hitting on the same issue.

hems:
> @hems, @shuding , did you figure out how to use `useSWRInfinite` in such cases?

I still did not update my project, but I'm hoping it will be solved by returning an array on the `getKey` function which them will be spread as parameters of my `fetcher` just like it used to be ( :

Graveheart:
Thank you for implementing this, it looks very promising! I'm trying to replace my current implementation of `useSwr`, however my issue is that my API response is of the following format: 
```
interface ServerResponse<T> {
  data: T[];
  meta: any;
}
```
where I need the meta object as it contains the total number of rows.
I see that `useSWRInfinite` right now works only with an array response. Is there any way that the meta can be included as part of the response?

loonfly:
Hi guys, i follow the example. i do setSize(size + 1) but the size keep getting back to 1 (original size). from console log i can see the size got plus 1 then next render the size got back to 1. bug or my problem?

styxlab:
@hems: I solved the problem with the missing key array by defining a higher order key function:
```
const getKeyHoc = (siteId: string) => (pageIndex: number): keyType => {
  return [siteId, `${pageIndex}`]
}
const { data, mutate, size, setSize } = useSWRInfinite(getKeyHoc(siteId), (siteId: string, pageIndex: string) =>
   request('/api/graphql', AssetsQuery, { siteId, index: Number.parseInt(pageIndex), limit: 10 })
)
```
That's a good alternative to the other proposals.

hems:
> keyType



> @hems: I solved the problem with the missing key array by defining a higher order key function:
> 
> ```
> const getKeyHoc = (siteId: string) => (pageIndex: number): keyType => {
>   return [siteId, pageIndex]
> }
> 
> const { data, mutate, size, setSize } = useSWRInfinite(getKeyHoc(siteId), (siteId: string, pageIndex: number) =>
>   request('/api/graphql', AssetsQuery, { siteId, index: pageIndex, limit: 10 })
> )
> ```
> 
> That's a good alternative to the other proposals.

Ouch... Still very weird :( 

How did you manage to import `keyType` ? the swr package isn't exporting it?

styxlab:
> How did you manage to import `keyType` ? the swr package isn't exporting it?

import { keyType } from 'swr/dist/types'

hems:
> import { keyType } from 'swr/dist/types'

It seems i can get away with:

````tsx
  const PAGE_SIZE = 20
  const initialData = [...]

  const getKey = (pageIndex, previousPageData): KeyLoader => {
    if (previousPageData && !previousPageData.length) return null // reached the end

    return [pageIndex, PAGE_SIZE, searchTerm] // SWR key
  }

  const options = { initialData }

  // // const { data, error, isValidating, mutate, size, setSize } = useSWRInfinite(
  const { data } = useSWRInfinite(getKey, getMarketCaps, options)
```

But i just started trying out.. will report if it works as expected

arielnmagno:
Hi, how do I rewrite something like this to useSWRInfinite? I've just started migrating from useSWRPages. 

```
function useRecentDevices(  
  deviceId: string | undefined,
  dateFromUtc: Date | undefined,
  dateToUtc: Date,
) {
  return useSWRPages<Pick<SummaryList, 'skip' | 'take'> | null, SummaryList>(
    'devices',
    ({ offset, withSWR }) => {
      const { data } = withSWR(
        useSWR(
          ['devices/recent', deviceId, offset?.skip, offset?.take ?? DEFAULT_TAKE],
          async (
            path: string,
            deviceId: string | undefined,    
            skip: number | undefined,
            take: number,
          ) => {
            const response = await get<SummaryList>(path, {
              deviceId,
              dateFromUtc: dateFromUtc?.toISOString(),
              dateToUtc: dateToUtc.toISOString(),
              skip,
              take,
            });
            if (!response.success) {
              throw new Error('unable to fetch data');
            }
            return response.data;
          },
          { suspense: !offset },
        ),
      );

      const devices = data?.devices;
      return devices ? devices.map(order => <DeviceLine key={order.salesOrderId} {...order} />) : null;
    },
    ({ data }) => {
      const { hasMore, skip, take } = data!;

      if (hasMore) {
        return { skip: skip + take, take };
      }

      return null;
    },
    [deviceId, dateFromUtc, dateToUtc],
  );
}

```

Luncher:
useSWRInfinite `getKey` indicates both the conditional fetch and the end of paging. How can I distinguish between these two situations.

PR #438, opened 2020-06-09
docs(useSWRInfinite): add an example for useSWRInfinite

Jack-Works:


examples/infinite-scrolling/react-window-infinite-loader.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <div className="ListItem" style={style}>
+      Item {data?.id} Data: {data?.data}
+    </div>
+  )
+}
+
+export default function App() {
+  const {isValidating, data, error, page, setPage} = useSWRInfinite((index, previousPageData) => {
+    if (previousPageData && previousPageData.length === 0) return null
+   return previousPageData.last()?.id
```
shuding, line 22: When `index` is 0, which means that's the first page, you should return `0` here because there's no previous page data.

Jack-Works: Umm I learn it from your PR

shuding: Check the second example in #435:

```js
  (index, previousPageData) => {
    // reached the end
    if (previousPageData && previousPageData.data.length === 0) return null

    // first page, previousPageData is null
    if (index === 0) return `/users?limit=10`

    return `/users?cursor=${previousPageData.nextCursor}&limit=10`
  },
```

Jack-Works: oh, I find it, cause it seems like the key `0` is considered as falsy. I suggest using `=== null` check

shuding: Yeah and you should return a string :)

Jack-Works:
sadly I can't continue on this until #391 is done and available in #435

Jack-Works:
😥

bravokiloecho:
Hello. It looks this #391 and #435 have now been merged. Would it be possible to finish this now? It would be super helpful!

shuding:
Sorry @Jack-Works it's an accident, I didn't intend to close this PR, feel free to reopen 🙏

Jack-Works:
That's okay, I'm sorry for not touching this pr so long. Does the new API have its examples now?

Issue #443, opened 2020-06-10
Missing `'browser'` field of `package.json`


heejongahn:
When target is `web`, webpack use `['browser', 'module', 'main']`as default value for [`resolve.mainFields`](https://webpack.js.org/configuration/resolve/#resolvemainfields).

As `swr`'s `package.json` currently does not provide `'browser'` field, `'module'` entry point is used and thus the bundle breaks with syntax error in some old browsers. (e.g. IE11)

Sure, users can always override `resolve.mainFields` setting by themself, but wouldn't it would be nice if `swr` provided `"browser": "./dist/index.js",` so it works out of the box for old browsers? If core team agrees, I could come up with the PR (though it would be one-liner.)

huozhi:
@heejongahn not sure which error did you receive but I think this issue https://github.com/vercel/swr/issues/278 is probably related. it would be good if you can provide more information about the error here for us to track.

usually the `"main"` field and `"module"` are pretty enough for bundler to consume, since they're targeting CommonJS and ESModule format of assets.

If you need to convert your application assets to other formats by webpack (like UMD or just IIFE), it's up to application layer. or you can also use babel to transform swr dist by babel to the format you need.

huozhi:
for some reason I saw the ESModule build is using ESNext syntax 

`build:esm": "tsc --target ESNext --module ES6 --outDir esm`

@shuding is it ok to use ES5 or it's specified to ESNext to due to some special reason?

heejongahn:
@huozhi  For me, it was `class` syntax in `esm/cache.js` which caused the error.

huozhi:
I think for temporary work around you can include swr in the babel transform list, based on https://github.com/vercel/swr/issues/278

```js
{
  test: /\.(js|jsx)?$/,
  exclude: /node_modules\/(?!swr)/,
  use: {
    loader: 'babel-loader',
  },
},
```

shuding:
@huozhi Yeah I think it's a mistake to use `ESNext`, totally okay to change it to `ES5`.

shuding:
Fixed in `swr@0.2.3`.

heejongahn:
Thanks!

PR #444, opened 2020-06-10
build: use es5 syntax for esm build (#443)

huozhi:
output es5 synatx (configured in tsconfig.json) for both esm and cjs build

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "dev": "now dev",
     "build": "npm run build:esm && npm run build:cjs",
     "build:cjs": "ncc build src/index.ts -o dist -m -e react",
-    "build:esm": "tsc --target ESNext --module ES6 --outDir esm",
+    "build:esm": "tsc --module ES6 --outDir esm",
     "watch": "tsc --watch",
     "types:check": "tsc --noEmit",
     "format": "prettier --write \"{src,test,examples}/**/*.{ts,tsx}\"",


shuding:
Do you think if we can keep ES6 for `"module"` and add a ES5 build for `"browser"`?

huozhi:
based on the spec from rollup https://github.com/rollup/rollup/wiki/pkg.module 
> `pkg.module` will point to a module that has ES2015 module syntax but otherwise only syntax features that the target environments support.

as for `pkg.browser` I'm not sure should we support it since browser field didn't really indicate a clear module type (cjs or esm).

huozhi:
> the new target makes the package 40% larger...
on library layer, I think we don't need to worry about the size. anyway it will be adopt by app layer first and compress later (after been bundled).

I was also wondering we should remove the minify for cjs build as well 🤔

shuding:
> on library layer, I think we don't need to worry about the size. anyway it will be adopt by app layer first and compress later (after been bundled).

Yes, but the number as a data point, is very useful for the user and package maintainer. More lightweight usually means faster and more reliable, and it's a good constraint for not adding things.

For example the biggest selling point of Preact is 

> Fast 3kB React alternative with the same modern API.

shuding:
Thanks for this PR @huozhi! I'm going to add microbundle and remove ncc in another PR to clean up the build scripts.

huozhi:
image

I checked an old version of preact, the dist provide both normal dist file and a minified one. the minified one is closed to the size it mentioned (after gzip I think?). the default entry is pointing to the unminified version.

microbundle will compress by default. maybe we don't really need it? and can still create an optional minified file for users who want to use it.

PR #445, opened 2020-06-10
Add microbundle

shuding:
Add microbundle and remove ncc.

Note that this PR is created on top of #435, because that PR removes TSX code from the source.


shuding:
Will open a new PR for this.

PR #457, opened 2020-06-13
Add subscription support to useSWR

sergiodxa:
This PR adds subscription support as proposed in https://github.com/vercel/swr/discussions/437#discussioncomment-25793

The way it works uses the same API

```js
const { data } = useSWR("/api/messages", fetcher, {
  subscribe(key, mutate) {
    return subscribeToSomething(key, mutate);
  },
});
```

### API

Subscribe is a function receiving `key` and `mutate`.

- `key` the unserialized key, what useSWR receives, it could be a string, array or function.
- `mutate` is the same useSWR returns, already bound to the key

The subscribe function must return an unsubscribe function, this unsubscribe function will be called in the case:

1. The component using this useSWR unmount
2. The subscribe fn is created again (it must be wrapped in useCallback to avoid this)
3. The key changed

A global subcribe function could be also used by setting it in SWRConfig.

#### WebSockets

The most basic example is to use it go subscribe via WebSockets and get new data, e.g. in a chat application.

```js
const { data } = useSWR("/api/chat-messages", fetcher, {
  subscribe(key, mutate) {
    // start connection to the WebSocket API
    const ws = new WebSocket("wss://my.app/api/chat-messages");
    ws.onmessage = (event) => {
      mutate((messages) => [...messages, JSON.parse(event.data)], false);
    };
    return () => ws.close();
  },
});
```

#### Geolocation API

We could use it to get new data from any type of stream, in this case Geolocation changes

```js
function fetcher() {
  return new Promise((resolve, reject) =>
    navigator.geolocation.getCurrentPosition(resolve, reject)
  );
}

const { data } = useSWR("geolocation", fetcher, {
  subscribe(_, mutate) {
    const id = navigator.geolocation.watchPosition((data) =>
      mutate(data, false)
    );
    return () => navigator.geolocation.clearWatch(id);
  },
});
```

#### Firebase (and other real-time databases)

We could use it in combination with real-time databases like Firebase to get the current value in the fetcher and subscribe to updates for that same value.

```js
async function fetcher() {
  const ref = db.collection("cities").doc("SF");
  const doc = await ref.get();
  return doc.data();
}

function subscribe(_, mutate) {
  const ref = db.collection("cities").doc("SF");
  ref.on("value", (snapshot) => {
    mutate(snapshot.val(), false);
  });
  return ref.off;
}

const { data } = useSWR("db/cities/SF", fetcher, { subscribe });
```

#### GraphQL

We could use it with GraphQL to start a subscription in the same useSWR call.

```js
const { data } = useSWR(
  [gql`SOME GRAPHQL QUERY`, variables],
  (query, variables) => client.fetch(gql, { variables }),
  {
    subscribe([_, variables], mutate) {
      return client.subscribe(gql`SOME GRAPHQL SUBSCRIPTION QUERY`, {
        variables,
      });
    },
  }
);
```

#### LocalStorage

It's also possible to use it to listen to global events like the storage one and get updates when another tab change the value in storage.

```js
const { data } = useSWR("storage-key", key => localStorage.get(key), {
  subscribe(key, mutate) {
    const listener = event => {
      if (event.key !== key) return;
      mutate(localStorage.get(key), false)
    });
    window.addEventListener("storage", listener);
    return () => window.removeEventListener("storage", listener);
  }
});
```

### Checklist

- [x] Add documentation
- [x] Add examples to the folder


sergiodxa:
@shuding @pacocoursey @Timer this is ready for review, I have added two examples to the repo and the documentation

shuding:
@sergiodxa I have a concern regarding this. Normally with just `fetch`, the request will be **deduped**. Considering:

```js
const useUser = () => useSWR('/api/user', fetcher)

function Foo () {
  return useUser().data
}

function App() {
  return <>
    <Foo/>
    <Foo/>
  </>
}
```

only 1 request will be sent. 

I believe most cases when people use a subscription, it should be globally shared (rather than per hook).

So it would be really great to also have the same ability with subscriptions, so people can truly forget about setting them up or tearing them down. We can use a counter to track it.

sergiodxa:
That makes sense. What do you think if we keep a Set of the current subscribe functions, if the subscribe change (it’s created dynamically based on a prop or state) we subscribe again, so the same key could have more than on subscribed function.

shuding:
> the same key could have more than on subscribed function

Why would that be helpful / what's the use case? I still like the idea of "key as resource identifier". 🤔

sergiodxa:
The subscribe could use a prop or state internally, I think that use case should not be recommended, but it can happen, another option is to don’t support it and if someone already started a subscription it will ignore new ones.

sergiodxa:
@shuding I added the feature to only have one subscriber per key, I also added a test for this

petrbrzek:
Any chance this would get merged any time soon? Looks very useful.

c10b10:
How would caching work for subscriptions that are updated? 

Take the case of pagination. Imagine being on page one and then navigating on page two which will tear down the original subscription and create a new one with new parameters. What happens if the user comes back to the first page? Will the cache of the original subscription be involved in any way?

sergiodxa:
@c10b10 the subscription will be kept if there is at least one hook still subscribed, once all the hooks are unmounted the unsubscribe function will be called, if the hook is mounted again in the future a new subscription will be started.

c10b10:
@shuding @huozhi @Kexin-Li Could you help unblock this by reviewing it?

src/use-swr.ts
```diff
@@ -601,6 +604,42 @@ function useSWR<Data = any, Error = any>(
     revalidate
   ])
 
+  // setup subscription
+  useEffect(() => {
```
shuding, line 608:
```suggestion
  useIsomorphicLayoutEffect(() => {
```

src/use-swr.ts
```diff
@@ -601,6 +604,42 @@ function useSWR<Data = any, Error = any>(
...
+      SUBSCRIBERS[key].count -= 1
+      // and if the count is now zero we will unsubscribe
+      if (SUBSCRIBERS[key].count === 0) {
+        SUBSCRIBERS[key].unsubscribe()
+      }
+    }
+    // we use key as dependency here but use _key internally to avoid this
+    // effect to clear and run again on every render if it's an array or
+    // function
+  }, [config.subscribe, key, boundMutate])
```
shuding, line 641: Do you think if we can directly use `configRef` instead of passing it as a dependency: https://github.com/vercel/swr/blob/master/src/use-swr.ts#L229-L232?

src/use-swr.ts
```diff
@@ -601,6 +604,42 @@ function useSWR<Data = any, Error = any>(
...
+    if (!SUBSCRIBERS[key]) {
+      SUBSCRIBERS[key] = { count: 0, unsubscribe: () => {} }
+    }
+
+    // If at the moment we run this there are no other subscribers for this key
+    if (SUBSCRIBERS[key].count === 0) {
+      // we will increment it by one
+      SUBSCRIBERS[key].count += 1
+      // and create the subscription storing the unsubscribe key
+      SUBSCRIBERS[key].unsubscribe = config.subscribe(_key, boundMutate)
```
shuding, line 622: Why `_key`? If we are using a function key (`useSWR(() => throwOrReturnKey)`) I feel it's better to have the key value (with args) here, wdyt?

huozhi:
I feel subscription is sth can warp on top of useSWR, personally. my gut feeling is the data flow for subscription case and useSWR hook case are different:

useSWR hook: passively request to endpoint with requests from hook internally, if there's any depeendency changed or any required events get triggered.
```
endpoint <=== useSWR.fetcher
```

subscription mode: initiative feed data from outside to the hook.

```
observable data ===> useSWR.subscribe
```

once we added subscription here the data flow will mixed 2 things together, not sure if it's the future pattern that we expected?

```
endpoint        <=== useSWR.fetcher
observable data ===> useSWR.subscribe
```

from the implementation perspective, the dependencies of subscription effects are mostly from the userland: config.subscribe, key, boundMutate( => mutate + key). it's an enhancement could be done outside of useSWR.

if the data comes from fetcher and observale endpoint simultaneously, useSWR and extra useEffect could help; if the subscription only consumes the data from the observable endpoint, I think `mutate` API + `useEffect` are enough to implement the subscription mode.

```js
const {data} = useSWR(key, fetcher) // optional if you wanna leverage data from fetcher
useEffect(() => {
  const handle = subscribe(..., () => {
    // do mutation where you like, mutate(key, ...)
  })
  
  return () => {
     handle.unsubscribe()
  }
}, [key, subscribe])
```

in the long term, I'd love to see subscription can be implemented as a plugin or middleware of swr that we can keep the core of swr as small as possible, but strongly extendable upon useSWR hook.

shuding:
> A global subcribe function could be also used by setting it in SWRConfig.

I don't think it's a good idea to allow this option in `SWRConfig`, doesn't feel like an inheritable option.

followbl:
anything I can help with to get this merged in? It would be an awesome addition.

dfee:
This feels like the sort of PR that I'll check in on a couple years down the road and find out it's stale.

I think it'd be helpful if the maintainers shared their thoughts:
1. `swr`, as is today, is feature complete, or
2. `swr` is interested in generally growing the codebase & adopting broader functionality, or
2. (same as above), but only interested in _minor_ features

My hunch is that the split is on [2] or [3], as this feature feels like a [2]. But the first step is having this discussion.

JollyGrin:
I just discovered that my websocket app that worked on now no longer works on vercel and have been looking for a quick fix that doesn't require setting up a pusher account. Is this still unsolved by swr? If so I need to discover another option.

sergiodxa:
@JollyGrin SWR will not solve your WebSocket server doesn't working on Vercel, I will still need a new hosting for your server, this PR aims to provide an integration between SWR and a subscription-based data source as it could be some WebAPIs or WebSocket.

tettoffensive:
@sergiodxa I found this thread trying to get graphql subscriptions to work with `nextjs`. I am using `swr` for my other graphql requests, but am I understanding this correctly that this will not work with `nextjs`/`swr` right now?

Is there currently anyway to get subscriptions working with `nextjs` running an `apollo-server-micro`?

lcdproductions:
@shuding @Timer is Vercel planning on maintaining SWR long term? It seems that there's a lot of rough around the edges features that would move SWR towards a 1.0 release - but given the lack of movement/communication in issues and features in the repo - it seems that it's not top of the priority for maintainers. If it's not a priority - is it possible to add/introduce maintainers outside of Vercel that might be interested in moving the project forward?

shuding:
@lcdproductions My apologies for delaying the review, I've been busy with my full-time job and other errands. And yes we are maintaining it, and will keep doing it in the future. It's an important lib that we're using in our production too.

Also I personally will devote more time cleaning up issues and PRs this weekend, as well as adding other contributors/maintainers. However I have to say that big changes and new features might still take a long time to get merged (or rejected) in open source projects.

Regarding this specific feature, I still have some concerns (but it's good to have it in beta versions). I'll review it carefully. 🙏

shuding:
@dfee Thanks for asking! One of the maintainers here.

> swr, as is today, is feature complete, or
> swr is interested in generally growing the codebase & adopting broader functionality, or
> (same as above), but only interested in minor features

First of all, SWR is not feature complete. Features like this are always great to have. 

Second, we want to support these new features in a more flexible and careful way. Most users only use a small subset of SWR features, and one important goal of SWR is lightweight and performant. That said, for features that are "pluggable", we want to explore a "plugin system" instead of adding them to the main code base (imagine `use=[swrLocalStorage, swrLogger]`).

For features that are good to have, but must be coupled to the core, we are open to discuss and accept PRs.

I'll add the guide of contributions soon.

shuding (changes requested):
Unfortunately, there're 2 things blocking this PR from landing:

1. As I commented [here](https://github.com/vercel/swr/pull/457/files#r464987273), I think we should take the following args instead of passing `_key`, the original key:

```js
config.subscribe(fnArgs === null ? key : fnArgs, boundMutate)
```

2. Tests are not passing, let's rebase master and check again.

Once these 2 are done, let's ship this feature.

huozhi:
from the reactions we know this feature is expected from community, since the conversation in this PR is kind of stale and we're not certainly sure about the what final API will look like. I created a RFC https://github.com/vercel/swr/issues/728 here for subscription mode related discussion. feel free to comment your ideas here.

bartenra:
Another use case for this is grpc-web.

followbl:
@sergiodxa can I help on this PR - would love to get this merged (thank you for your work) - been following your discussion on subscription support and in the recent RFC. It looks as if it has @shuding blessing to get this merged -  I think it'd be a huge win for the lib.

sergiodxa:
@followbl I'm not using SWR anymore so feel free to fork this branch (or copy the code to a new fork of SWR, that would probably be easier to merge with master) and send a PR.

followbl:
@sergiodxa thanks - will do. Have to ask - of course understand if don't want to answer - you using something else or a reason you're not using it now?

sergiodxa:
@followbl I'm using React Query, I still like SWR a lot, I just wanted the DevTools of RQ and the mutations hook

huozhi:
carry on swr subscription support on PR #913, feel free to comment on. 
if you have any new idea / use cases wanna provide, you can post into RFC #728

PR #462, opened 2020-06-14
Remove extraneous IS_SERVER from useIsomorphicLayoutEffect

morgs32:
We know we're not on the server in `useIsomorphicLayoutEffect` right?
https://github.com/vercel/swr/blob/new-pagination-api/src/use-swr.ts#L40


morgs32:
Eh what does it matter.

PR #464, opened 2020-06-14
Do not rerender if key changes but not data

morgs32:
This is so that you do not get an unnecessary rerender when the key has changed but the cached data is the same. In my case I'm conditionally fetching and when the key becomes valid I get an extra render while data is still `undefined`.

A few changes are from a linter that I assume ran on commit? Let me know if it's uncouth to have left them there.


shuding:
Thank you @morgs32! Can you solve the conflict? And we will take this PR 👍

morgs32:
Sorry for the delay. Rebased and left out the lint fixes.

shuding (approved):

PR #476, opened 2020-06-19
enhance: move online listeners to config (#474)

### LINKED ISSUE #474: Move the `online` event listener to config
# Bug report

SWR registers multiple `online` event listeners:

image

## Expected Behavior

There should be only 1 listener, just like `visibilitychange`.

## Additional Context

https://github.com/vercel/swr/blob/03c7b8320dde30fab24c9706cd33c5eb27d0552e/src/use-swr.ts#L494-L497

needs to be moved to

https://github.com/vercel/swr/blob/03c7b8320dde30fab24c9706cd33c5eb27d0552e/src/config.ts#L87-L88

huozhi:
## Changes

Resolves #474 , attach online event listeners in config 

## Test Plan

* build swr locally and yarn link to the examples/focus-revalidate 
* add one more swr hook there
* disconnect and reconnect the network in devtools
* expected: the requests get re-fetched

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,6 +1,5 @@
 import deepEqual from 'fast-deep-equal'
 import isDocumentVisible from './libs/is-document-visible'
-import isOnline from './libs/is-online'
 import {
   ConfigInterface,
   RevalidateOptionInterface,
@@ -74,31 +73,6 @@ const defaultConfig: ConfigInterface = {
   shouldRetryOnError: true,
   suspense: false,
   compare: deepEqual
-}
-
-// setup DOM events listeners for `focus` and `reconnect` actions
-if (typeof window !== 'undefined' && window.addEventListener) {
-  const revalidate = revalidators => {
-    if (!isDocumentVisible() || !isOnline()) return
-
-    for (const key in revalidators) {
-      if (revalidators[key][0]) revalidators[key][0]()
-    }
-  }
-
-  // focus revalidate
-  window.addEventListener(
-    'visibilitychange',
-    () => revalidate(FOCUS_REVALIDATORS),
-    false
-  )
-  window.addEventListener('focus', () => revalidate(FOCUS_REVALIDATORS), false)
-  // reconnect revalidate
-  window.addEventListener(
-    'online',
-    () => revalidate(RECONNECT_REVALIDATORS),
-    false
-  )
 }
 
 export {
--- a/src/config.ts
+++ b/src/config.ts
@@ -15,6 +15,7 @@ const cache = new Cache()
 const CONCURRENT_PROMISES = {}
 const CONCURRENT_PROMISES_TS = {}
 const FOCUS_REVALIDATORS = {}
+const RECONNECT_REVALIDATORS = {}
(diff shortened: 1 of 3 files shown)


src/config.ts
```diff
@@ -75,25 +76,44 @@ const defaultConfig: ConfigInterface = {
...
+    window.addEventListener('visibilitychange', revalidate, false)
+    window.addEventListener('focus', revalidate, false)
+
+    focusEventsBinded = true
+  }
+
+  // set up reconnecting when the browser regains network connection
+  if (defaultConfig.revalidateOnReconnect && !onlineEventsBinded) {
+    const revalidate = () => {
+      if (!isOnline()) return
```
huozhi, line 100: I don't know if we need `isDocumentVisible` condition here, so I left a comment for checking.

src/use-swr.ts
```diff
@@ -513,6 +518,14 @@ function useSWR<Data = any, Error = any>(
           revalidators.pop()
         }
       }
+      if (onReconnect && RECONNECT_REVALIDATORS[key]) {
```
huozhi, line 521: the recycle revalidator part is kind of repetitive, we can re-abstract them but I feel it would be good in the separate PR

src/config.ts
```diff
@@ -75,25 +76,44 @@ const defaultConfig: ConfigInterface = {
...
 
 // Focus revalidate
-let eventsBinded = false
-if (typeof window !== 'undefined' && window.addEventListener && !eventsBinded) {
-  const revalidate = () => {
-    if (!isDocumentVisible() || !isOnline()) return
+if (typeof window !== 'undefined' && window.addEventListener) {
+  // only bind the events once
+  let focusEventsBinded = false
+  let onlineEventsBinded = false
```
shuding, line 82: This logic can be removed.

huozhi:
> I think we don't need the extra `RECONNECT_REVALIDATORS` 🤔. Can we use the same `FOCUS_REVALIDATORS` (and rename it to just `REVALIDATORS`?

we still have `CACHE_REVALIDATORS` and it works in a different way, i guess they might not able to merge 3 of them together? if revalidators for "focus" and "reconnect" are megerable, do you feel ok to call it "ONLINE_FOCUS_REVALIDATORS"? or any other better ideas about it?

huozhi:
oops, looks like if we still have to separate reconnect and focus revalidators? when you disable `revalidateOnFocus`, first callback in `FOCUS_REVALITORS` still get executed even it's actually the one for reconnect

src/use-swr.ts
```diff
@@ -507,41 +559,16 @@ function useSWR<Data = any, Error = any>(
       CACHE_REVALIDATORS[key].push(onUpdate)
     }
```
shuding, line 560: We can move https://github.com/vercel/swr/pull/476/files#diff-ab20307c185620c9b54994fa4dcee1b7R512-R513 here and change https://github.com/vercel/swr/pull/476/files#diff-ab20307c185620c9b54994fa4dcee1b7R556-R560 to

```js
addRevalidator(RECONNECT_REVALIDATORS, onUpdate)
```

PR #480, opened 2020-06-19
Update keyErr on the cache when mutating

### LINKED ISSUE #477: Error should be cleared from the cache when mutating a key
# Bug report

We only update the data in the cache when mutating:

https://github.com/vercel/swr/blob/03c7b8320dde30fab24c9706cd33c5eb27d0552e/src/use-swr.ts#L122-L125

## Expected Behavior

Error should also be updated.

## Additional Context

Thanks to @lucleray for bringing up this issue.

cristianbgp:
Resolves #477 ❤️

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -122,8 +122,8 @@ const mutate: mutateInterface = async (
   if (typeof data !== 'undefined') {
     // update cached data, avoid notifying from the cache
     cache.set(key, data, false)
-    cache.set(keyErr, error, false)
   }
+  cache.set(keyErr, error, false)
 
   // update existing SWR Hooks' state
   const updaters = CACHE_REVALIDATORS[key]
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -77,7 +77,7 @@ const mutate: mutateInterface = async (
   _data,
   shouldRevalidate = true
 ) => {
-  const [key] = cache.serializeKey(_key)
+  const [key, , keyErr] = cache.serializeKey(_key)
   if (!key) return
 
   // if there is no new data, call revalidate against the key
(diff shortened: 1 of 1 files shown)


src/use-swr.ts
```diff
@@ -122,6 +122,7 @@ const mutate: mutateInterface = async (
   if (typeof data !== 'undefined') {
     // update cached data, avoid notifying from the cache
     cache.set(key, data, false)
+    cache.set(keyErr, error, false)
```
huozhi, line 125: if data is undefined but error is not?

```js
if (typeof error !== 'undefined') {
   cache.set(keyErr, error, false)
}
```

cristianbgp: I don't think it might be necessary since this will update `keyErr` to undefined to clear the key unless there was an error, in that case should update to that new error 🤔

huozhi: yea we might not need to have this condition inside `if` scope, but I think we might want to move cache set outside of condition `if (typeof data !== 'undefined')`?

if the data is undefined but error is not, then we still didn't update the `keyError` in cache.

I created a test case in #481 , it won't pass if we update error cache when data is available

shuding: Yes I agree @huozhi. Consider the following case:

```js
mutate(key, post(API))
```

Where post is an async function and it throws. In that case, we will expect the cache to be updated:
- data → cleared
- error → set

cristianbgp: It should be like this right?
```javascript
if (typeof data !== 'undefined') {
  // update cached data, avoid notifying from the cache
  cache.set(key, data, false)
}
cache.set(keyErr, error, false)
```

shuding: Yep!

cristianbgp: @shuding thanks! @huozhi Thanks for creating a test case!

PR #481, opened 2020-06-20
test: add test case for error updating in cache when mutate

huozhi:
add test case for #480

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -118,9 +118,8 @@ const mutate: mutateInterface = async (
     if (error) throw error
     return data
   }
-
+  // update cached data and error, avoid notifying from the cache
   if (typeof data !== 'undefined') {
-    // update cached data, avoid notifying from the cache
     cache.set(key, data, false)
   }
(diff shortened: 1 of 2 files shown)


src/use-swr.ts
```diff
@@ -118,11 +118,13 @@ const mutate: mutateInterface = async (
...
   }
-
+  // update cached data and error, avoid notifying from the cache
   if (typeof data !== 'undefined') {
-    // update cached data, avoid notifying from the cache
     cache.set(key, data, false)
   }
+  if (typeof error !== 'undefined') {
+    cache.set(keyErr, error, false)
+  }
```
shuding, line 127: I merged #480. I think here the error cache should be cleared even `error` is undefined.

Can you resolve the conflicts so we can merge this PR? Thanks for adding a test case! 😃

huozhi: rebased the upstream and resolved the conflicts! 🙌

shuding: Thanks!!

PR #487, opened 2020-06-26
Add microbundle

shuding:
Add microbundle and remove ncc.


ykzts:
This pull request breaks the server-side app (e.g. Next.js), and the This is This is because there is a circular dependency in swr that prevents exporting to commonjs using `module.exports`  when converting to commonjs because of the circular dependency of swr.

shuding:
Thanks for the notice @ykzts! We will work on a solution before publishing a new version, or revert this PR.

PR #490, opened 2020-06-29
Add persistSize option to useSWRInfinite

shuding:
### API

```js
useSWRInfinite(getKey, fetcher, { persistSize: true })
```

The boolean value will be `false` by default. When enabling it, the page size will not be reset to 1 (or `initialSize` if set) when the first page's key changes.

### Related

https://github.com/vercel/swr/pull/435#discussion_r443202280


shuding:
Thanks @lucleray!

PR #496, opened 2020-07-01
Add microbundle

shuding:
Reverts vercel/swr#492

Since #493 is merged, we don't have circular dependency anymore.


shuding:
Related: #500.

PR #500, opened 2020-07-03
build: use tsc for cjs build

huozhi:
## Changes

Simplify cjs build process, use tsc to compile without bundling the assets, aligning with esm build process.

## Purpose

* for library, bundling and code minification can be left to application side, let its bundler and minifier to consume the library code.
* we tried to leverage microbundle for assets generation. but unfortunately commonjs assets generated by microbundle doesn't work since it disabled rollup `output.esModule`, then the output cjs code won't work with interpolate import since it doesn't have `__esModule` property. (import deafult will not work when consum cjs bundle)

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "scripts": {
     "dev": "vercel dev",
     "build": "npm run build:esm && npm run build:cjs",
-    "build:cjs": "ncc build src/index.ts -o dist -m -e react",
+    "build:cjs": "tsc --outDir dist",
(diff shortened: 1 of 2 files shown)


huozhi:
also found a modern mode of microbundle https://github.com/developit/microbundle#modern which use the syntaxes that modern browsers support. should be helpful for calculate the (smaller) size of swr.

huozhi:
close this PR as we want to move to some matured bundlers

PR #506, opened 2020-07-08
don't retry on error if errorRetryCount is set to 0

dominictwlee:
thanks @whj1995 for spotting out the bug. This should handle the edge case where user decides to set errorRetryCount to 0 instead of using the shouldRetryOnError boolean flag.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -34,7 +34,7 @@ function onErrorRetry(
   }
 
   if (
-    config.errorRetryCount != null &&
+    typeof config.errorRetryCount === 'number' &&
     opts.retryCount > config.errorRetryCount
   ) {
     return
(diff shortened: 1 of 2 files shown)


src/config.ts
```diff
@@ -33,7 +33,10 @@ function onErrorRetry(
     return
   }
 
-  if (config.errorRetryCount && opts.retryCount > config.errorRetryCount) {
+  if (
+    config.errorRetryCount != null &&
```
shuding, line 37:
```suggestion
    typeof config.errorRetryCount === 'number' &&
```

We should have that check no matter the option is set to `null` or `undefined`, so I suggest to test if it's a number.

shuding:
Good catch @dominictwlee @whj1995, and thanks for opening a PR!

PR #510, opened 2020-07-09
fix: check if mounted when trigger rerender

### LINKED ISSUE #494: React experimental concurrent issue
https://github.com/facebook/react/pull/18838

I tried `swr` with suspense mode, with a recent `0.0.0-experimental-7f28234f8` react build, and I get errors `Can't perform a React state update on a component that hasn't mounted yet` for each `useSWR` call.

> This indicates that there's a side-effect spawned during render, that then asynchronously tries to set state when the response comes back. E.g. kick off a request and then setState with the result. It's a common mistake when implementing Suspense.

> This should be done in useEffect.

Does this make sense to you?

Unfortunately I can't help you a lot more than that, it doesn't help to where the error comes exactly.

huozhi:
### fix

Resolve #494 

react exp version will warn setState calls when component is not mounted. on suspense mode, swr trigger revalidate and lead to call `rerender` when the component is not mounted yet. 

image

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -224,6 +224,7 @@ function useSWR<Data = any, Error = any>(
       }
     }
     if (shouldUpdateState || config.suspense) {
+      if (unmountedRef.current) return
       rerender({})
     }
   }, [])


samselikoff:
would love to get a release with this fix in it! 🙏

PR #511, opened 2020-07-09
Export interfaces used by useSWRInfinite

lolPants:
Rationale: Extending these interfaces in a TypeScript project when composing your own hooks using SWR.
You can already do this with config/response interfaces used by `useSWR` as these are already exposed.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,10 @@
 export * from './use-swr'
 import { default as useSWR } from './use-swr'
-export { useSWRInfinite } from './use-swr-infinite'
+export {
+  useSWRInfinite,
+  ExtendedConfigInterface,
+  ExtendedResponseInterface
+} from './use-swr-infinite'
 export { cache } from './config'
 export {
   ConfigInterface,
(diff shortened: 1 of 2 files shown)


lolPants:
@shuding No problem! Love the library and am glad to contribute 😄 

Would you prefer the interfaces to be renamed internally as well, or just renamed while being exported?

huozhi:
> Thanks for this PR! It would be great to rename them to `SWRInfiniteConfigInterface` and `SWRInfiniteResponseInterface`, so it's more clear to the users.

for other types in `types.ts`, do we need to have the prefix `SWR` as well? what do you think?

pacocoursey:
> Would you prefer the interfaces to be renamed internally as well, or just renamed while being exported?

Let's rename them internally as well, no harm in being consistent.

> for other types in types.ts, do we need to have the prefix SWR as well?

Not in this PR, at least, since this would be a breaking change for anyone relying on those types. If you'd like, feel free to open a PR with these changes and we can discuss there.

lolPants:
Sounds good to me 👍

Issue #508, opened 2020-07-08
onError callback not reflecting component state changes


diogofscmariano:
# Bug report
onError callback not reflecting component state changes

## Description / Observed Behavior
In a component using useState and useSwr, the state is not correctly propagated to the onError callback

## Expected Behavior
The useState should be correctly accessible (read/write) inside the useSwr callbacks

## Repro Steps / Code Example

```
const [onRefresh, setOnRefresh] = useState(false)
const [snackbarOpen, setSnackbarOpen] = useState(false)
const [errorOpen, setErrorOpen] = useState(false)

console.log(onRefresh, snackbarOpen, errorOpen)

const onSuccessCallback = useCallback(() => {
  setOnRefresh(true)
}, [])
const onErrorCallback = useCallback(() => {
  console.log(onRefresh)
  onRefresh ? setSnackbarOpen(true) : setErrorOpen(true)
}, [onRefresh])

const { data: availableSolutions, error } = useSWR<Solution[]>(
  "api/solutions", getAvailableSolutions, {
    refreshInterval: 2000,
    onSuccess: onSuccessCallback,
    onError: onErrorCallback,
  }
)
```
The objective here is to conditionally show a different component if the error is in the first call or in the refresh mechanism, but in this case, the onRefresh is always false (initial state value) and does not get updated.
In the example i'm running, the first `console.log` shows the `onRefresh` as true, but inside the callbacks its value is always false.

lintuming:
Yeah, the callback called by the `useSWR` always the callback created in the  initial render of your component,it's the problem of the stale closure.

> 
![image](https://user-images.githubusercontent.com/37504587/86995127-c88df800-c1da-11ea-9ba0-33c36bcf2c31.png)

You can use `useRef` as a workaround
```js
const onRefresh = useRef(false);
const onSuccessCallback = useCallback(() => {
  onRefresh.current = true
}, [])
const onErrorCallback = useCallback(() => {
   onRefresh.current ? setSnackbarOpen(true) : setErrorOpen(true)
}, [onRefresh])

```

Issue #515, opened 2020-07-13
[enhance] use the latest version of the config callbacks


lintuming:
Related issue: #508 

Currently,the config callbacks captured by the [`eventsRef` ](https://github.com/vercel/swr/blob/master/src/use-swr.ts#L237) is the callbacks created in the initial render of the component. 
If the config callbacks is pure, everything is fine.
but if the config callbacks rely on some variables of the components , it would cause some bugs related to the stale closure.

# solution
I think if the `useSWR` hook read the config callbacks via `useRef` would be great, but this may bring some bugs if some APP rely on the current behavior of the config callbacks.
```
  const eventsRef = useRef({
    emit: (event, ...params) => {
      if (unmountedRef.current) return
      configRef.current[event](...params)
    }
  })
```
what do you think?

lintuming:
cc @shuding

sergiodxa:
This looks like a good idea, would you like to send a PR with the change? It would even great to run the whole test suite against it and probably add a few tests to ensure it works as you expect

lintuming:
@sergiodxa 
sure, would love to take it.

akafaneh:
I've been struggling with an issue because of this, i'm relying on some redux state that gets updated a lot in the onError callback, and the fact that it only capture and and save the callback on first render made it impossible to get around the issue with a clean solution, so this fix would save a lot for me.

thanks

PR #521, opened 2020-07-14
fix: `eventsRef` should trigger the lastest version of config callbacks

lintuming:
# fix

Resolve: #508 #515 
 
Now the `eventsRef` read the config via the `ref`,which always capture the lastest version of the config callbacks.

This fix the problem of the stale closure related to the config callbacks.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -193,7 +193,9 @@ function useSWR<Data = any, Error = any>(
   )
 
   const configRef = useRef(config)
-  configRef.current = config
+  useIsomorphicLayoutEffect(() => {
+    configRef.current = config
+  })
 
   if (typeof fn === 'undefined') {
     // use the global fetcher
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -192,10 +192,8 @@ function useSWR<Data = any, Error = any>(
     config
   )
 
-  const lastestConfig = useRef(config)
-  useIsomorphicLayoutEffect(() => {
-    lastestConfig.current = config
-  })
+  const configRef = useRef(config)
+  configRef.current = config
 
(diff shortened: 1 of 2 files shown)


lintuming:
cc @sergiodxa

lintuming:
Also, i've noticed some code paths that using the `config` may have some similar stale closure problem.
Should we consider the scenario the config passed to the `useSWR` hook may change along the component lifecycle?

src/use-swr.ts
```diff
@@ -192,6 +192,11 @@ function useSWR<Data = any, Error = any>(
     config
   )
 
+  const lastestConfig = useRef(config)
+  useIsomorphicLayoutEffect(() => {
```
huozhi, line 196: I'm not sure if we need to have this useEffect here, could it just be updated in each rendering directly?

```js
const configRef = useRef(config)
configRef.current = config
```

shuding: @huozhi I feel that we still need to wrap it with `useEffect` (`useIsomorphicLayoutEffect` here). Updating refs directly in render can be dangerous in Concurrent Mode (not 100% about this specific usage here). 

I'll approve this for now, but I suggest adding a comment to clarify :+1:

huozhi: sure, good to know there's side effect on concurrent mode! 🙏 sorry for my unreasonable comment above

lintuming: @shuding thanks to point this out,do i need to change this back?

huozhi: yep, I think you can just simply revert the last commit. the rest looks good 👍

nicholaschiang:
@shuding @huozhi @lintuming could we merge this?

Issue #517, opened 2020-07-13
using suspense mode with nullable ReturnType<fetcher> will cause a loop


MHA15:
# Bug report
I am using swr to get a value which can be object or null
but when null received from server my component will stuck in a loop and will show loading and request many times to server

## Expected Behavior

call getValue once and use it's falsy value as resolved data

## Repro Steps / Code Example
https://codesandbox.io/s/swr-problem-with-suspense-with-nullable-fetcher-zfhc2

huozhi:
@MHA15 you can provide your useSWR hook with option `initialData` and disable `revalidateOnMount`. Once swr detect there's any fallback value (initialData) or fulfilled value in cache. swr will stop revalidating in suspense mode. pick and to disable other options like `revalidateOnFocus` / `revalidateOnReconnect` if it's needed for you. this [issue](https://github.com/vercel/swr/issues/450#issuecomment-642218441) might be helpful 

```js
useSWR(key, fetcher, {
  initialData: /*<the default data you'd like to use>*/,
  revalidateOnMount: false,
}
```

and you might want to wrap your suspendable component with `React.Suspense`. I tried your example and modify it into the following pattern. 

```js
import * as React from "react";
import useSWR from "swr";

function getValue() {
  console.log("requested again");
  return Promise.resolve(null /* also for other falsy values like "" */);
}

function Main() {
  useSWR("nullable value", getValue, {
    suspense: true,
    initialData: "",
    revalidateOnMount: false
  });

  return (
    <div className="App">
      <h1>Hello CodeSandbox</h1>
      <h2>Start editing to see some magic happen!</h2>
    </div>
  );
}

export default function App() {
  return (
    <React.Suspense fallback={<span>fallback</span>}>
      <Main />
    </React.Suspense>
  );
}
```

MHA15:
> @MHA15 you can provide your useSWR hook with option initialData and disable revalidateOnMount. Once swr detect there's any fallback value (initialData) or fulfilled value in cache. swr will stop revalidating in suspense mode. pick and to disable other options like revalidateOnFocus / revalidateOnReconnect if it's needed for you. this issue might be helpful

I did try turning off all revalidates and it doesn't help and also this problem is only when falsy value returned from fetcher function and it will work just fine in other cases 
and also I didn't want to set a defaultValue, why should I?

muqg:
[...]

PR #524, opened 2020-07-15
fix problem of unnecessary revalidates when resolved data is falsy

MHA15:
we can use ?? operator of typescript 3.8 in future
this will fix bug of infinite loading when using suspense and a fetcher which resolved a falsy value

close issue #517

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -583,8 +583,15 @@ function useSWR<Data = any, Error = any>(
     // (it should be suspended)
 
     // try to get data and error from cache
-    let latestData = cache.get(key) || initialData
-    let latestError = cache.get(keyErr) || initialError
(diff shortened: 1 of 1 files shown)


shuding:
@MHA15 Thank you! Do you mind upgrading TS to the latest version and use `??` in this PR?

MHA15:
> @MHA15 Thank you! Do you mind upgrading TS to the latest version and use `??` in this PR?

thanks for your attention, about ?? operator I think that we shouldn't use it because we don't want to change received null with initial value 
maybe below code might be better 
```js
   cache.has(key) ? cache.get(key) : initialValue
```
but still these approaches will not let us having `undefined` as resolved value

shuding:
Makes sense. Then I think we can merge this for now :+1:

Issue #549, opened 2020-07-23
isValidating change only triggers first hook in component tree to render


blackarctic:
# isValidating change only triggers first hook in component tree to render

## Description / Observed Behavior

It seems that if multiple components use `useSWR(someKey)`, a change to `isValidating` only triggers the first of these components in the tree to rerender. 

## Expected Behavior / Repro Steps / Code Example

![2020-07-22 23 22 25](https://user-images.githubusercontent.com/10765285/88257916-b64c9900-cc73-11ea-9348-062558f2ca10.gif)

In this example, I have 3 components that all use `useSWR('/api/things')`: `ThingsList`, `ThingsListError`, and `ThingsRefreshButton`, but `ThingsRefreshButton` comes first in the component tree and is the only one receiving the re-render when the value changes. The expected behavior here should be that if the value of `isValidating` changes, all these components should receive a re-render trigger.

Relevant code is provided below:

![Screen Shot 2020-07-22 at 11 29 47 PM](https://user-images.githubusercontent.com/10765285/88257940-c49ab500-cc73-11ea-8e83-600bfb9fde68.png)

![Screen Shot 2020-07-22 at 11 29 58 PM](https://user-images.githubusercontent.com/10765285/88257945-c8c6d280-cc73-11ea-8b6d-8c506d369c38.png)

![Screen Shot 2020-07-22 at 11 30 06 PM](https://user-images.githubusercontent.com/10765285/88257947-cb292c80-cc73-11ea-8656-fc023fe54234.png)

![Screen Shot 2020-07-22 at 11 30 16 PM](https://user-images.githubusercontent.com/10765285/88257954-cebcb380-cc73-11ea-8e1d-45866a7932c5.png)

![Screen Shot 2020-07-22 at 11 30 24 PM](https://user-images.githubusercontent.com/10765285/88257960-d11f0d80-cc73-11ea-820a-06858b8e1711.png)

## Additional Context

SWR version: "0.2.3"

blackarctic:
Note that if `ThingsList` is moved up in the component tree, it is the only component that receives the re-render trigger:

![2020-07-22 23 39 41](https://user-images.githubusercontent.com/10765285/88258348-cf097e80-cc74-11ea-8d17-fd115532828a.gif)

blackarctic:
Anyone have any idea what's going on here?

huozhi:
[...]

PR #569, opened 2020-07-29
Promote isValidating to the global level of data and error

blackarctic:
This fixes https://github.com/vercel/swr/issues/549 by caching and broadcasting `isValidating` in the same way that `data` and `error` are cached and broadcasted.

**Before**

before

**After**

after

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -75,7 +75,7 @@ if (!IS_SERVER && window.addEventListener) {
 const trigger: triggerInterface = (_key, shouldRevalidate = true) => {
   // we are ignoring the second argument which correspond to the arguments
   // the fetcher will receive when key is an array
-  const [key, , keyErr] = cache.serializeKey(_key)
+  const [key, , keyErr, keyValidating] = cache.serializeKey(_key)
   if (!key) return Promise.resolve()
 
   const updaters = CACHE_REVALIDATORS[key]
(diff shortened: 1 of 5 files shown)


blackarctic:
Anything I can do to speed up a review?

blackarctic:
Bump

shuding (approved):
Code looks good to me! It would be great to have a test for this I think.

Thank you @blackarctic!

blackarctic:
@shuding Test has been added: **useSWR › should broadcast isValidating**

**Test on master**

before

**Test on fix-is-validating**

after

src/use-swr.ts
```diff
@@ -232,6 +246,7 @@ function useSWR<Data = any, Error = any>(
 
   const initialData = cache.get(key) || config.initialData
   const initialError = cache.get(keyErr)
+  const initialIsValidating = cache.get(keyValidating)
```
huozhi, line 249: small question, not blocking, will the `initialIsValidating` become undefined at the first time? do we want to forcedly convert it to a boolean value here like `Boolean(cache.get(keyValidating))` or `!!cache.get(keyValidating)`?

blackarctic: We can convert it to a boolean. Probably a good practice.

blackarctic: Done

huozhi (approved):

blackarctic:
Bump

blackarctic:
Bump

kristojorg:
Bump

shuding:
Thank you all for helping!

Issue #331, opened 2020-04-06
[TypeScript] Fetcher function doesn't respect array of keys types


enzoferey:
With a piece of code like:

```ts
const key = `some graphql query with variables`;
const entityId = 1;

const fetcher = (key: string, entityId: number) => { ... }

const { data } = useSWR([key, entityId], fetcher);
```

If you now inverse the arguments of `fetcher`, TypeScript doesn't report a type error even though it would be receiving the wrong parameters in runtime.

crutch12:
Bump
Any progress on this?

shuding:
Fixed with #1477 🎉

PR #572, opened 2020-07-30
Type fetcher function arguments based on array key (#331)

TLadd:
Changes in tests provide and example of what this looks like:

```
const { data: v1 } = useSWR(
  ['args-1', obj, arr] as const,
  (a, b, c) => a + b.v + c[0]
)
```

Need to do `as const` so Typescript infers the tuple type instead of something like `(string | object | string[])[]` (in this case). Now, a, b, and c have the proper types and there are type errors if I misuse one of the variables or have the incorrect number of arguments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "react": "16.11.0",
     "react-dom": "16.11.0",
     "ts-jest": "24.1.0",
-    "typescript": "3.6.4"
+    "typescript": "3.9.7"
   },
   "dependencies": {
     "fast-deep-equal": "2.0.1"
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,4 +1,6 @@
-export type fetcherFn<Data> = (...args: any) => Data | Promise<Data>
+export type fetcherFn<Data, TKey extends readonly any[] = any[]> = (
+  ...args: TKey
+) => Data | Promise<Data>
 export interface ConfigInterface<
   Data = any,
   Error = any,
@@ -47,9 +49,11 @@ export interface RevalidateOptionInterface {
   dedupe?: boolean
 }
 
-export type keyType = string | any[] | null
-type keyFunction = () => keyType
-export type keyInterface = keyFunction | keyType
+export type keyType<TKey extends readonly any[] = any[]> = string | TKey | null
+type keyFunction<TKey extends readonly any[] = any[]> = () => keyType<TKey>
+export type keyInterface<TKey extends readonly any[] = any[]> =
+  | keyFunction<TKey>
+  | keyType<TKey>
 export type updaterInterface<Data = any, Error = any> = (
   shouldRevalidate?: boolean,
   data?: Data,
--- a/src/types.ts
+++ b/src/types.ts
@@ -49,11 +49,9 @@ export interface RevalidateOptionInterface {
   dedupe?: boolean
 }
 
-export type keyType<TKey extends readonly any[] = any[]> = string | TKey | null
-type keyFunction<TKey extends readonly any[] = any[]> = () => keyType<TKey>
-export type keyInterface<TKey extends readonly any[] = any[]> =
-  | keyFunction<TKey>
-  | keyType<TKey>
+export type keyType<TKey = any> = string | TKey | null
+type keyFunction<TKey> = () => keyType<TKey>
+export type keyInterface<TKey = any> = keyFunction<TKey> | keyType<TKey>
 export type updaterInterface<Data = any, Error = any> = (
   shouldRevalidate?: boolean,
   data?: Data,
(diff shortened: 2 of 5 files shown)


shuding:
Thanks for bringing this up! Do you think if there're any other work arounds that don't require `as const`? It feels a little bit overhead to me to have that.

TLadd:
I would have thought the `as const` was necessary for TypeScript to infer the correct type for the arguments tuple. By default, TypeScript sees something like

```
let args = ["hello", 42]
```

and infers a type of `(string | number)[]`. It does this because arrays are mutable and it allows me to call `.push` or `.remove` to mutate the items in the array. When I instead do

```
let args = ["hello", 42] as const
```

The `as const` tells TypeScript to make it a type error to mutate the array and thus it is safe to assign the tuple type `readonly ["hello", 42]` to args.

I looked into what react-query was doing and noticed https://github.com/tannerlinsley/react-query/blob/master/types/index.d.ts#L202. I don't really understand why, but this apparently forces TypeScript to infer the tuple when the array is passed as the argument directly (not assigned to a variable first). I copied that same type over and it works, except for the case when the key is a function where `as const` is still necessary.

shuding:
cc @promer94 would like to hear your thoughts about this

promer94:
TypeScript V4 made some great improvements on tuple type inferences. I create a [simplified useSWR in typescript playgroud](https://www.typescriptlang.org/play?ts=4.0.5#code/K4Zwpgyg6gSgFAcgPQEMAOBLJowCcEA0ABHMLgDYCURAvAHxEDeAxgPYB2Ir5YAdOawDmpCpQC+lAFCSc0eHGr0iydFhz5iIqrQYsOXHvyFbxUmeDlwA2iszZwGogEYATAGYAuppTEARot1JIiI2Tm4+AWEUXgAXVgAZVmYUHkSAdzwAYRRwBSlg0IMI419Y1gAxDAAPMAATPMkJaVlYOAUdIhtUO3VCZVc3BC8SHyJ-DsYgkP1woyiyxOTU1gzcbNzKfOmww0i4UrjKmvrNxrMW+QDO2zUHPqsB4ncBj2G4UfGlSYKZ3eNouKLFJgdJZHJgBo-HbFYS+B4eKRNcyQVrtJRdVT2PB9RgoABc-T6vjxLAJCEGYjEbw+V2+2yKc3eCySwNBa3BkPpsz2pVKzERZmaFlRVwxPTuxFxZKcRJJzDJFLExAez2GUuc7ip3j8xGYxFqtKmhW5-2ZSxBKzBGy2xr+sN4fJtvDQoAAFnAnFtarwUGchSjLh0xbdsZL8YS-HKFQhKcqBk41eGBlSiDl6TFtWNdfrDVCGXsAQkWctVusIac8yb7Y6jc63R6vT6-dIYgBPNBgIgQGK4DDsQQAaTArdoRBAPb7giIAB8iOxgORyJI2x2iAAVYBoHhDkc0Ii4MAoWoccgjqzj3v94i8G-AdgAa3YK3YVley-bnfK7B3o7RDG7l6DsOM7rpu27Du+q4-nuAGTj+s4bluYDwUQX47i2H5EDAYAgAuMQADxrqOd6Ps+DB7kRs4AAq4KwAC2GDgIRDAYauABmYAxMwrpfvhAAiKAxCgxEPk+aTsMQACCuCCCARBgFUMRgOwtRydBRA7uRRBTNJsnyYpymqfuh7Huwp6dBek7XjefYcbgGkeEQAD82nBMEbQ3j6MkgHi54TleRCeQOCIdNhuHkARAlCXQ1AElMbnBLpckKUpKlyW0VwHkeJ5npZAWebZeAOdQLnxQl7lwJ5KDeb5eWCNZvDBVcYV4fxgkoDFRBxeVPVJfpqVGXVIHzouzmuT1PXvDVY7+YIzU4a1UUdV140Tb13n9YZ6V-jNgHDQu2ilWtx1TbJBJ1fN4WRe1DDdcd92VTe1VnRZs0NSg7Ctq+IVKC1EVtdF0iSGxd7MDEGAcEQFwA8Je4kWJEl9SlW0acBe6aXA97DgSSXEGx7AEhxXE8ewMNSd5MV0jErq0Wkc5gHTACiuC0bgiAAHKsDERAYHRSF0cpSm1AgiKSEAA). 



In this way, user dont have to use ***'as const'*** to force typescript to infer tuples.

However, we have to solve some problems before we bring this feature into swr.

### Stable api design
Since we are working on next version fo swr, some of apis may change. 

### Upgrade TypeScript Version
1. upgrade TypeScript version
2. upgrade eslint and prettier (so they could parse and format typescript correctly)
3. refactor useSWR type (all the related types like cache interface also need refactor)
4. may need to upgrade ts-jest and @testing-library/react

### Backward compatibility
Some users may not be able to upgrade the typescript in their project. I don't have much experiences about this part.

In general, i think it would be great if swr could have this ability. I would love to do more researches and experiments on this.

promer94:
cc @shuding @huozhi

vhakulinen:
Any updates on this?

shuding:
A solution without `as const` has been landed in #1477, thank y'all for the contributions and discussions here! 🎉

PR #578, opened 2020-08-01
Add useDebugValue to show the data label in the React DevTools

### LINKED ISSUE #574: Better React DevTools support
# Bug report

Currently in the React DevTools, a SWR hook looks like this:

When unfolded:

ss-YfSimTHR@2x

The actual data is located at "SWR → Ref → Data", which is not friendly when debugging.

## Expected Behavior

By using `useDebugValue(data)` (https://reactjs.org/docs/hooks-reference.html#usedebugvalue) we can directly show the data next to the `▶ SWR` toggle. It makes debugging easier:

Kexin-Li:
close #574 

With this PR, it will look like this in the React DevTools:

image

I'm not sure if we need to show `undefined` when the data is not returned as expected? like this:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,6 @@
   "homepage": "https://swr.now.sh",
   "license": "MIT",
   "scripts": {
-    "dev": "vercel dev",
     "build": "npm run build:esm && npm run build:cjs",
     "build:cjs": "ncc build src/index.ts -o dist -m -e react",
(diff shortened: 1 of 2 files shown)


shuding:
Thank you for working on this! I think it's fine to show `"undefined"` there when the data is not ready or errored.

PR #604, opened 2020-08-17
Remove Isomorphic Fetch from Axios example

jmichaliga:
Cuts down on confusion and a dependency.

### DIFF
--- a/examples/axios/pages/api/data.js
+++ b/examples/axios/pages/api/data.js
@@ -1,4 +1,4 @@
-import fetch from 'isomorphic-unfetch'
+import axios from 'axios'
 
 const projects = [
   'facebook/flipper', 'vuejs/vuepress', 'rust-lang/rust', 'zeit/next.js'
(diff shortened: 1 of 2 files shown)


examples/axios/pages/api/data.js
```diff
@@ -7,7 +7,7 @@ const projects = [
 export default (req, res) => {
   if (req.query.id) {
     // a slow endpoint for getting repo data
-    fetch(`https://api.github.com/repos/${req.query.id}`)
+    axios(`https://api.github.com/repos/${req.query.id}`)
       .then(resp => resp.json())
```
shuding, line 11: Does axios support this?

jmichaliga: Yes, for a standard 'GET' request. I'll update the Typescript example as well.

shuding: It seems `axios` doesn't support `.json()`:

image

https://runkit.com/shuding/5f3e51784380ec001acd0501

jmichaliga: You're correct. The response requires the `.data` object for axios. I've updated the other example in a separate PR and can close this one.

PR #607, opened 2020-08-19
Remove Isomorphic Fetch from Axios Typescript example

jmichaliga:
Update to remove dependencies and confusion. Proper `data` objects handle through axios.


shuding (approved):
Thanks!

PR #619, opened 2020-08-27
fix: keep synchronized mutation executed in order

### LINKED ISSUE #556: Mutate swallows errors when using with useSWR hook.
# Bug report

Errors thrown in a promise passed to `mutate` should bubble up but are not when using in conjunction with `useSWR`.

## Expected Behavior

Here's what the docs say:

```js
try {
  const user = await mutate('/api/user', updateUser(newUser))
} catch (error) {
  // Handle an error while updating the user here
}
```

That is here: https://github.com/vercel/swr#returned-data-from-mutate.

## Repro Steps / Code Example

Example stops working from docs if you do this:

```js
useSWR('/api/user');

...

try {
  const user = await mutate('/api/user', updateUser(newUser))
} catch (error) {
  // Handle an error while updating the user here
}
```

Here's a sandbox example: https://codesandbox.io/s/old-microservice-ind5c?file=/src/App.js

## Additional Context

I'm using the latest version of SWR.

Please see the [discussion here](https://github.com/vercel/swr/discussions/547) for more details.

### LINKED ISSUE #566: Race condition when mutate is called multiple times in short succession with the same key
# Bug report

First off, thank you for your awesome work on this library 😊  I really enjoy using it and it's tremendously helpful.

Calling mutate multiple times with the same key within a very short amount of time leads to a race condition. When I pass a callback function to mutate, if I update the value on call #1 of mutate, call #2 of mutate doesn't get the updated value.

This leads to weird unexpected bugs since it's not clear that the library will behave this way.

It seems like the bug is here: https://github.com/vercel/swr/blob/master/src/use-swr.ts#L131. If mutate is called multiple times in short succession, cache.get(key) will not be up to date.

## Expected Behavior

I expect no race condition. When I update the value on call #1 of mutate, I expect the value to be updated on call #2 of mutate.

## Repro Steps / Code Example

https://codesandbox.io/s/swr-mutate-race-condition-gnrs3

## Additional Context

SWR version: 0.2.3

huozhi:
Fixes #566

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -138,12 +138,14 @@ const mutate: mutateInterface = async (
   if (_data && typeof _data === 'function') {
     // `_data` is a function, call it passing current cache value
     try {
-      data = await _data(cache.get(key))
+      _data = _data(cache.get(key))
     } catch (err) {
       error = err
     }
-  } else if (_data && typeof _data.then === 'function') {
-    // `_data` is a promise
+  }
+  // if `_data` is promise originally or the returned value from `_data()` is promise
+  // keep resolving it
+  if (_data && typeof _data.then === 'function') {
     try {
       data = await _data
     } catch (err) {


ranisalt:
This has flipped the issue for me.

Before the patch, after every request is finished and a call to `mutate` is issued, it would overwrite everything that was mutated before it (the parameter to the callback passed to `mutate` is **the original value fetched**, not **the latest state**), so the last call to mutate would contain the original state mutated by the last call to mutate *only*.

After this patch, the intermediate state is wrong, as every request finished suffers from the same issue of only mutating from the original state, but the final result is the correct, up to date state. See the reproduction:

https://imgur.com/qtw94UA

huozhi:
hi @ranisalt , sorry for my limited understanding about your problem here. would you mind giving more explanation of the incorrect behavior of this PR you mentioned? like some code samples will be super helpful, thanks

huozhi:
close this in favor of #735

PR #621, opened 2020-08-28
chore: optimization dispatch

anc95:
when dispatch, just add a compare with current state to prevent unnecessary rerender


shuding (approved):

shuding:
Thanks!

PR #622, opened 2020-08-28
chore: fix error when data tobe 0 null

anc95:
if the cache data is 0 or null, now the initialData will be from config. it is not expected

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -236,11 +236,11 @@ function useSWR<Data = any, Error = any>(
   }
 
   const resolveData = () => {
-    const cachedData = cache.get(key);
+    const cachedData = cache.get(key)
     return typeof cachedData === 'undefined' ? config.initialData : cachedData
   }
 
-  const initialData = resolveData();
+  const initialData = resolveData()
   const initialError = cache.get(keyErr)
 
   // if a state is accessed (data, error or isValidating),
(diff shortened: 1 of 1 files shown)


src/use-swr.ts
```diff
@@ -235,7 +235,7 @@ function useSWR<Data = any, Error = any>(
     fn = config.fetcher
   }
 
-  const initialData = cache.get(key) || config.initialData
+  const initialData = cache.get(key) === undefined ? config.initialData : cache.get(key)
```
shuding, line 238:
```suggestion
  const cachedData = cache.get(key)
  const initialData = typeof cachedData === 'undefined' ? config.initialData : cachedData
```

src/use-swr.ts
```diff
@@ -479,7 +479,7 @@ function useSWR<Data = any, Error = any>(
     // and trigger a revalidation
 
     const currentHookData = stateRef.current.data
-    const latestKeyedData = cache.get(key) || config.initialData
+    const latestKeyedData = cache.get(key) === undefined ? config.initialData : cache.get(key)
```
shuding, line 482:
```suggestion
    const cachedData = cache.get(key)
    const latestKeyedData = typeof cachedData === 'undefined' ? config.initialData : cachedData
```

anc95:
I just updated code as your suggest

shuding (approved):
Thanks!

PR #627, opened 2020-08-29
Add passKeyToFetcher option

santiagopoli:
This PR adds a new **'passKeyToFetcher'** option (which defaults to true). When set to false, fetcher functions will not receive the key as the first argument. For example:

```javascript
import useSWR from "swr";
import useSesion from "./useSession";

// if passKeyToFetcher was true (which is the default), getUser would receive (key, userId, token).
const getUser = (userId, token) => {
  return fetch(`/api/users/${userId}`, { headers: { authorization: `Bearer ${token}` } })
    .then((response) => response.json());
};

export const UserContext = React.createContext({});
export default function UserProvider({ children }) {
  const { session } = useSession();

  const { data: user } = useSWR(
    () => ["getUser", session.userId, session.token], 
    getUser,  
    { passKeyToFetcher: false }
  );

  const value = { user };

  return <UserContext.Provider value={value}>{children}</UserContext.Provider>;
```

Why? If my fetcher function is generic and not dependant of SWR, I don't want to receive (and ignore) a key which is only used by SWR.

I'm obviously open to suggestions, but I hope you find this useful :)

Thanks!

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -66,6 +66,7 @@ const defaultConfig: ConfigInterface = {
   refreshWhenOffline: false,
   shouldRetryOnError: true,
   suspense: false,
+  passKeyToFetcher: true,
   compare: deepEqual
 }
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -16,6 +16,7 @@ export interface ConfigInterface<
   revalidateOnMount?: boolean
   revalidateOnReconnect?: boolean
   shouldRetryOnError?: boolean
+  passKeyToFetcher?: boolean
   fetcher?: Fn
   suspense?: boolean
   initialData?: Data
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -260,7 +260,7 @@ function useSWR<Data = any, Error = any>(
     let shouldUpdateState = false
     for (let k in payload) {
       if (stateRef.current[k] === payload[k]) {
-        continue;
+        continue
       }
 
       stateRef.current[k] = payload[k]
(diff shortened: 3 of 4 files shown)


huozhi:
when your `key` is an array or a function returning array value, swr will pass the array value as arguments to fetcher. so that you could know when `passKeyToFetcher` is true/false outside of swr.
in the provided case, I think you receive the key because you pass `"getUser"` in the returned array so it becomes the first argument of fetcher.

you can remove that string in return value of key function or you can skip it in the fetcher to achieve it

santiagopoli:
The first argument in the array is always the key (and it needs to be a string) so it can be avoided as far as I know. If you pass an object for example, it will be converted into a string.

huozhi:
```js
function fetcher(...args) {
    console.log('fetcher args:', ...args)
    return Promise.resolve(123);
}
const { data } = useSWR(() => ['getUser', 4, 6], fetcher); 
```

in the above snippet, the arguments passed into fetcher are string `'getUser'` and numbers `4`, `6`. if we remove the `'getUser'`, the arguments will become `4` and `6` only. I guess it's similar to your example that passing `session.userId` and `session.token` to the fetcher.

santiagopoli:
@huozhi As said above, the first argument in the array has an special treatment, as it is the key. Look at this example:

https://codesandbox.io/s/exciting-beaver-mkin5?file=/src/App.js

If I ignore the "getUser" parameter it will work as far as there aren't other swr's that use the same array. In the above example, If i call both getUser and getOrder with the same params ([1]), the results will be faulty. The only way to make it work is to add "getUser" and "getOrder" as the first parameter of the array and ignoring it afterwards. It is important that the key is unique.

santiagopoli:
@huozhi I was looking at the code in https://github.com/vercel/swr/blob/9b7d7ad55489ec56fd11c2b48e7615c4b518d320/src/libs/hash.ts#L11 and now I can understand that the key is not computed only by using the first argument, but by hashing the entire array. But what seems odd is that the key must be unique, so you need to append something unique to it to be hashed properly (it doesn't need to be the first argument, but any of them). In a 'real' application using UUIDs, it might be impossible to get a collision of two different functions (getUser and getOrders) with the exact same params, however that can happen.

This renders this PR 'useless' but opens a different discussion: should we be using four parameters instead of three? I mean (key, params, fetcher, options) instead of (key, fetcher, options).

Thanks!

huozhi:
https://swr.vercel.app/docs/middleware#serialize-object-keys could be a solution to this requirement for keeping more a generic fetcher

PR #638, opened 2020-09-07
try fix cannot stop polling by setting refreshInterval (#632)

### LINKED ISSUE #632: Cannot stop polling by setting refreshInterval 0
SWR version 0.3.1

I have a demo reproducing this issue [here](https://codesandbox.io/s/upbeat-fermi-fi5i0?file=/src/App.js).

In this demo I have a flag variable controlling whether swr should poll by conditionally setting `refreshInterval` value.

```ts
{
  refreshInterval: shouldPoll ? 1000 : 0
}
```

In my swr config I also have an `onSuccess` callback in which I change my flag value

```ts
{
  refreshInterval: shouldPoll ? 1000 : 0,
  onSuccess() {
    setShouldPoll(false);
  }
}
```

When the flag is set to `false` as the first fetch finishes, the polling stops normally as expected. Everything is fine so far.

However, if I set the flag value in the callback not of the first fetch but the revalidation fetches, polling is not stopping.

```ts
{
  refreshInterval: intValue < 2 ? 1000 : 0, // intValue is initially 0
  onSuccess() {
    setIntValue(value => value + 1); 
  }
}
```

After the first fetch, intValue is set to 1, and polling is supposed to continue. After the 2nd fetch, intValue is set to 2, and polling is supposed to stop because refreshInterval is now 0. But the polling continued.

promer94:
Fix #632 

## Notice
- The old implementation of pooling can't stop timer if refreshInterval changes **during revalidate**. This problem can be solved by using the configRef to get the latest refreshInterval.
- The old implementation of pooling will stop the old timer if refreshInterval changes. ~~But the new implementation of pooling ***will not clear the old timer***~~.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -571,6 +571,7 @@ function useSWR<Data = any, Error = any>(
     addRevalidator(RECONNECT_REVALIDATORS, onReconnect)
     addRevalidator(CACHE_REVALIDATORS, onUpdate)
 
+    let timer = null
     const tick = async () => {
       if (
         !stateRef.current.error &&
@@ -583,11 +584,11 @@ function useSWR<Data = any, Error = any>(
         await revalidate({ dedupe: true })
       }
       if (configRef.current.refreshInterval) {
-        setTimeout(tick, configRef.current.refreshInterval)
+        timer = setTimeout(tick, configRef.current.refreshInterval)
       }
     }
     if (configRef.current.refreshInterval) {
-      setTimeout(tick, configRef.current.refreshInterval)
+      timer = setTimeout(tick, configRef.current.refreshInterval)
     }
 
     return () => {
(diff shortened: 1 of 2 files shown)


src/use-swr.ts
```diff
@@ -571,49 +571,37 @@ function useSWR<Data = any, Error = any>(
...
-      if (timer) clearTimeout(timer)
+      // cleanup
+      dispatch = () => null
+
+      // mark it as unmounted
+      unmountedRef.current = true
+
+      removeRevalidator(FOCUS_REVALIDATORS, onFocus)
+      removeRevalidator(RECONNECT_REVALIDATORS, onReconnect)
+      removeRevalidator(CACHE_REVALIDATORS, onUpdate)
```
shuding, line 602: I think we should set up a `timer = setTimeout(...)` and clear it when exiting the effect here.

shuding (commented):
What if `refreshInterval` changes from `1000` → `0` → `1000`? I think it will stop at `0` and never resume.

shuding (approved):
Thanks!

PR #665, opened 2020-09-25
Remove unnecessary dependencies from `mutate` and `setSize`

### LINKED ISSUE #664: setSize in useSWRInfinite is not memoizing
# Bug report

I'm trying to automate fetching all pages. So, I'm using `setSize` in `useEffect` hook. When previous data contain a link for the next portion data I increment `size`. But every time when my own hook is being recalculated `useSWRInfinite` returns me different `useSize`. 

## Expected Behavior

I expected `setSize` which returning from `useSWRInfinite` not changed and changing only `size` property.

## Repro Steps / Code Example

```js
const getKey = (index, previousData) => {
    if (index === 0) return defaultUrl;

    return previousData?.data?.['@odata.nextLink'];
};

export const useMyDocuments = () => {
    const { data, setSize, mutate, isValidating } = useSWRInfinite(
        getKey,
        fetcher
    );

    useEffect(() => {
        if (data?.[data?.length - 1]?.data?.['@odata.nextLink']) {
            setSize((size) => size + 1);
        }
    }, [data, setSize]); //with `setSize` in dependencies hook executable in infinite loop.

    const docs = useMemo(() => data?.flatMap((el) => el?.data?.value), [data]);

    return {
        data: docs,
        isLoading: isValidating,
        mutate,
    };
};
```

## Additional Context

swr version - **0.3.4**

When I remove `setSize` from dependencies `useEffect` all work is fine. But this is a dirty hack for my opinion.

shuding:
The bound `mutate`/`setSize` from `useSWR` and `useSWRInfinite` should not change over re-renders, so you can include them safely inside hook dependencies (or not include them). It should work just like a `setState` in React.

### Fix 1

This PR removes the `key` dependency from `boundMutate` in `useSWR`, instead we use the ref of the key.

### Fix 2

We used to do this in `useSWRInfinite`:

```js
const swrInfinite = swr
//...
const mutate = swrInfinite.mutate
swrInfinite.mutate = useCallback(..., [mutate])
```

So after one pass, the original object's field `swr.mutate` will be updated to `useCallback(...)`. So upon the next renders, the dependency `[mutate]` will always be considered changed.

Closes #664

### DIFF
--- a/src/libs/hash.ts
+++ b/src/libs/hash.ts
@@ -13,7 +13,10 @@ export default function hash(args: any[]): string {
   let key = 'arg'
   for (let i = 0; i < args.length; ++i) {
     let _hash
-    if (args[i] === null || (typeof args[i] !== 'object' && typeof args[i] !== 'function')) {
+    if (
+      args[i] === null ||
+      (typeof args[i] !== 'object' && typeof args[i] !== 'function')
+    ) {
       // need to consider the case that args[i] is a string:
       // args[i]        _hash
       // "undefined" -> '"undefined"'
--- a/src/use-swr-infinite.ts
+++ b/src/use-swr-infinite.ts
@@ -177,18 +177,13 @@ function useSWRInfinite<Data = any, Error = any>(
     extraConfig
   )
 
-  const swrInfinite = swr as SWRInfiniteResponseInterface<Data, Error>
-
   // keep the data inside a ref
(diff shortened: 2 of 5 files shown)


sergiodxa (approved):

shuding:
Thanks @sergiodxa for reviewing!

PR #689, opened 2020-10-12
Fix some return types

ivc369:
Maybe we should remove `:any` after `set()`. And I also added some return types to `key()` and `has()`.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -15,17 +15,17 @@ export default class Cache implements CacheInterface {
     return this.__cache.get(_key)
   }
 
-  set(key: keyInterface, value: any): any {
+  set(key: keyInterface, value: any) {
     const [_key] = this.serializeKey(key)
     this.__cache.set(_key, value)
     this.notify()
   }
 
-  keys() {
+  keys(): string[] {
     return Array.from(this.__cache.keys())
   }
 
-  has(key: keyInterface) {
+  has(key: keyInterface): boolean {
     const [_key] = this.serializeKey(key)
     return this.__cache.has(_key)
   }


src/cache.ts
```diff
@@ -15,17 +15,17 @@ export default class Cache implements CacheInterface {
...
 
-  set(key: keyInterface, value: any): any {
+  set(key: keyInterface, value: any) {
     const [_key] = this.serializeKey(key)
     this.__cache.set(_key, value)
     this.notify()
   }
 
-  keys() {
+  keys(): string[] {
```
promer94, line 24: this could be inferred since this.__cache is a Map

src/cache.ts
```diff
@@ -15,17 +15,17 @@ export default class Cache implements CacheInterface {
...
     return Array.from(this.__cache.keys())
   }
 
-  has(key: keyInterface) {
+  has(key: keyInterface): boolean {
```
promer94, line 28: this could be inferred since this.__cache is a Map

promer94 (changes requested):

src/cache.ts
huozhi: @promer94 I guess `has` would be safe to return boolean type?

promer94: https://github.com/vercel/swr/blob/651c368833164945a8ee5708eb3dd6091241f05b/src/types.ts#L124-L133

i just find we already had the interface of cache.  so we may don't need to add return type for these functions.

src/cache.ts
ivc369: you are right, thanks :):)

PR #701, opened 2020-10-19
fix: error should not be cleared while revalidating

### LINKED ISSUE #657: Error cleared on next render
# Bug report

SWR is clearing the error data from cache on render even when the key is `null`, which should signal SWR not to try and fetch at all. When a user signs in to my app, I save their token to a cookie (and to react state) and the presence of the token tells SWR to perform a fetch. When this comes back with a server error, I need to display that error to the user and also clear their token from state and cookie. However, clearing the token triggers SWR to delete the cached error, so no error is displayed to the user. 

## Expected Behavior

If a render tells SWR not to fetch, don't delete old data/errors.

## Repro Steps / Code Example

Here is a [CodeSandbox](https://codesandbox.io/s/quizzical-fermat-xof56?file=/src/App.js). Try adding some text and clicking "Log In". You won't see any error message. If you comment out line 12 where I clear the token, you will see an error message. 

## Additional Context

SWR version ^0.3.2

huozhi:
fix #657

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -356,7 +356,12 @@ function useSWR<Data = any, Error = any>(
         cache.set(keyValidating, true)
         if (!shouldDeduping) {
           // also update other hooks
-          broadcastState(key, undefined, undefined, true)
+          broadcastState(
+            key,
+            stateRef.current.data,
+            stateRef.current.error,
+            true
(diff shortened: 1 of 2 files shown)


shuding (approved):
Looks good, thanks!

PR #715, opened 2020-10-24
enable typescript strict mode

promer94:


shuding:
@promer94 This PR looks great to me! But we have some conflicts due to a recent code refactoring, could you please fix them?

src/types.ts
```diff
@@ -1,4 +1,4 @@
-export type fetcherFn<Data> = ((...args: any) => Data | Promise<Data>) | null
+export type fetcherFn<Data> = (...args: any) => Data | Promise<Data>
```
shuding, line 1: Why was this removed?

src/use-swr-infinite.ts
```diff
@@ -29,58 +29,43 @@ type SWRInfiniteResponseInterface<Data = any, Error = any> = responseInterface<
...
-    }
-  }
-
-  config = Object.assign(
+  const config = Object.assign(
     {},
     defaultConfig,
     useContext(SWRConfigContext),
-    config
+    args.length === 3
```
shuding, line 48: Let's keep the exact same behavior:

```suggestion
    args.length > 2
```

src/use-swr-infinite.ts
```diff
@@ -29,58 +29,43 @@ type SWRInfiniteResponseInterface<Data = any, Error = any> = responseInterface<
...
+      ? args[2]
+      : args.length === 2 && typeof args[1] === 'object'
+      ? args[1]
+      : {}
   )
-  let {
+
+  const fn =
+    args.length === 3
```
shuding, line 56:
```suggestion
    args.length > 2
```

src/use-swr.ts
```diff
@@ -193,60 +196,51 @@ const mutate: mutateInterface = async (
...
-    } else if (typeof args[1] === 'object') {
-      config = args[1]
-    }
-  }
+  const _key = args[0]
+  const config = Object.assign(
+    {},
+    defaultConfig,
+    useContext(SWRConfigContext),
+    args.length === 3
```
shuding, line 214:
```suggestion
    args.length > 2
```

src/use-swr.ts
```diff
@@ -193,60 +196,51 @@ const mutate: mutateInterface = async (
...
+      ? args[2]
+      : args.length === 2 && typeof args[1] === 'object'
+      ? args[1]
+      : {}
+  )
+  const fn =
+    args.length === 3
```
shuding, line 221:
```suggestion
    args.length > 2
```

src/types.ts
shuding: Just saw #738, we can remove it when we implement the `passive` option (or a better name).

promer94: because the fetcherFn type is also used in useSWRInfinite and this kind of usage is not documented as i mentioned in #738 
```js
function useSWR<Data = any, Error = any>(
  ...args:
    | readonly [keyInterface]
      // allow pass null as fetcher here
    | readonly [keyInterface, fetcherFn<Data> | null]
    | readonly [keyInterface, ConfigInterface<Data, Error>]
    | readonly [
        keyInterface,
        // allow pass null as fetcher here
        fetcherFn<Data> | null,
        ConfigInterface<Data, Error>
      ]
): responseInterface<Data, Error>
```
This still allow we pass null as fetcher and wont affect the useSWRInfinite

shuding: I see :+1:

Then let's re-open this PR and set the merge target to the `next` branch, given this is a huge refactoring it's great to have it on v1.

PR #730, opened 2020-10-29
Allow passing null as the fetcher function for lazy execution

maggo:
This allows again for lazy execution since passing `undefined` broke with the introduction of a default fetcher in #367


shuding (approved):
Thank you!

PR #735, opened 2020-10-30
Fix race condition when calling mutate synchronously

### LINKED ISSUE #566: Race condition when mutate is called multiple times in short succession with the same key

shuding:
### Bug

We have this inconsistency issue in current version:

```js
// we call `mutate` twice, synchronously, with a callback that reads the latest value
// let's assume the initial data is `0`

mutate('key', v => v + 1, false)
mutate('key', v => v + 1, false)
```

The data should be `2` after these 2 synchronous mutation calls. However if you test, it is actually `1` (here's a great [bug reproduction](https://codesandbox.io/s/swr-mutate-race-condition-gnrs3) from @derindutz).

The cause is that we are always updating the cache in async (as #619 tries to fix), so the argument `v` passed to the second mutation callback is still `0`. To make it clear, here's a diagram addressing this bug:

image

### Solution

We at first [tried](https://github.com/vercel/swr/pull/619/files#diff-755ef1ecc6f2f4c1cc601c6091dfcba5f674c653863b70913a115c0856a7b632R141) to make the mutate process synchronous (if the callback is not a promise or async function), but that brings another issue which is intermediate state:

```js
mutate('key', v => v + 1, false)
mutate('key', v => v + 1, false)
```

Now it correctly renders `2` at the end, but there's a moment that `1` was rendered before that. Instead we should "batch" 
synchronous mutations. It is because that the first mutation can't get deduplicated by the second one anymore. They both render the data.

image

To ensure the **correctness** of `mutate`, we need to make state broadcasting always asynchronous (kinda like React), but cache needs to be updated ASAP. So the latter mutations will always have the result from the former, but only the last mutation will get rendered:

image

The consequence of this fix is, if you mutate something and immediately check the result, it won't be updated. So in our test I added `await 0` to delay the check in the next tick:

```js
mutate('mutate-3', 'off', false)
await 0 // if no `await` here, it's gonna be wrong
expect(container.textContent).toMatchInlineSnapshot(`"off"`)
```

Fixes #566. Related to #619.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -46,7 +46,7 @@ const MUTATION_TS = {}
 const MUTATION_END_TS = {}
 
 // generate strictly increasing timestamps
-let now = (() => {
+const now = (() => {
   let ts = 0
   return () => ts++
 })()
@@ -140,7 +140,7 @@ const mutate: mutateInterface = async (
   const beforeConcurrentPromisesTs = CONCURRENT_PROMISES_TS[key]
 
   let data, error
-  let isMutationAsync = false
+  let isAsyncMutation = false
 
   if (_data && typeof _data === 'function') {
     // `_data` is a function, call it passing current cache value
@@ -153,7 +153,7 @@ const mutate: mutateInterface = async (
 
   if (_data && typeof _data.then === 'function') {
     // `_data` is a promise
-    isMutationAsync = true
+    isAsyncMutation = true
     try {
       data = await _data
     } catch (err) {
@@ -163,15 +163,19 @@ const mutate: mutateInterface = async (
     data = _data
   }
 
-  // check if other mutations have occurred since we've started this mutation
-  // if so we don't update cache or broadcast change, just return the data
-  if (
-    beforeMutationTs !== MUTATION_TS[key] ||
-    beforeConcurrentPromisesTs !== CONCURRENT_PROMISES_TS[key]
-  ) {
-    if (error) throw error
-    return data
+  const shouldAbort = (): boolean | void => {
+    // check if other mutations have occurred since we've started this mutation
+    if (
+      beforeMutationTs !== MUTATION_TS[key] ||
+      beforeConcurrentPromisesTs !== CONCURRENT_PROMISES_TS[key]
+    ) {
+      if (error) throw error
+      return true
+    }
   }
+
+  // if there's a race we don't update cache or broadcast change, just return the data
+  if (shouldAbort()) return data
 
   if (typeof data !== 'undefined') {
     // update cached data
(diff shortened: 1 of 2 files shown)


src/use-swr.ts
```diff
@@ -45,6 +45,12 @@ const CACHE_REVALIDATORS = {}
 const MUTATION_TS = {}
 const MUTATION_END_TS = {}
 
+// generate strictly increasing timestamps
+let now = (() => {
```
huozhi, line 49:
```suggestion
const now = (() => {
```

src/use-swr.ts
```diff
@@ -122,28 +128,32 @@ const mutate: mutateInterface = async (
...
+  MUTATION_TS[key] = now() - 1
   MUTATION_END_TS[key] = 0
 
-  // keep track of timestamps before await asynchronously
+  // track timestamps before await asynchronously
   const beforeMutationTs = MUTATION_TS[key]
   const beforeConcurrentPromisesTs = CONCURRENT_PROMISES_TS[key]
 
   let data, error
+  let isMutationAsync = false
```
huozhi, line 143: maybe isMutationAsync -> isAsyncMutation? (not blocking)

src/use-swr.ts
```diff
@@ -163,13 +174,31 @@ const mutate: mutateInterface = async (
...
+  MUTATION_END_TS[key] = now() - 1
+
+  if (!isMutationAsync) {
+    // let's always broadcast in the next tick
+    // to dedupe synchronous mutation calls
+    // check out https://github.com/vercel/swr/pull/735 for more details
+    await 0
+
+    // we skip broadcasting if there's another mutation
+    // happened synchronously
```
huozhi, line 193: seems are same code with above, maybe we can make it a short function? like

```js
const resolveDataOnMutation = () => {
  if (
      beforeMutationTs !== MUTATION_TS[key] ||
      beforeConcurrentPromisesTs !== CONCURRENT_PROMISES_TS[key]
    ) {
      if (error) throw error
      return data
    }
  }
}
```

you might come up with other better names

shuding:
Great feedback @huozhi, thanks!

huozhi (approved):
LGTM! awesome work and explanation there 👍

src/use-swr.ts
```diff
@@ -122,28 +128,32 @@ const mutate: mutateInterface = async (
...
   const beforeConcurrentPromisesTs = CONCURRENT_PROMISES_TS[key]
 
   let data, error
+  let isAsyncMutation = false
 
   if (_data && typeof _data === 'function') {
     // `_data` is a function, call it passing current cache value
     try {
-      data = await _data(cache.get(key))
+      _data = _data(cache.get(key))
```
promer94, line 148: we can keep the previous 'else if' flow here instead of reassign _data here ?
```js
if (_data && typeof _data === 'function') {
    // `_data` is a function, call it passing current cache value
    try {
      data = _data(cache.get(key))
    } catch (err) {
      error = err
    }
  } else if (_data && typeof _data.then === 'function') {
    // `_data` is a promise
    isAsyncMutation = true
    try {
      data = await _data
    } catch (err) {
      error = err
    }
  } else {
    data = _data
  }
```

shuding: We can’t. If `_data` is an async function, calling it without `await` will get a promise returned, and we have to resolve it again.

promer94: On i see now 😅. I think i can add a test for mutate with async function. we don't have one now.

shuding: @promer94 that would be great, thanks!

shuding:
Tested a bit, should be fine to ship this :)

PR #736, opened 2020-10-30
feat: customizable listener

huozhi:
### Goal
provide a way to let other js runtime like RN to customize the way of listening events. then the focus/visible/connect are able to be associated with platform behaviors

### Changes

set the listener attaching process to config

add `config.setOnFocus(callback)`
add `config.setOnConnect(callback)`

for web browser, the fallback will be like following, for RN, it could be sth else

```js
setOnConnect(callback) {
   window && window.addEventListener && window.addEventListener('online', callback, false)
}
```

then `config.setOnConnect` becomes a call that can register event listeners, the event listeners are actually executing the revalidators. i.e.

```
// reconnect revalidate
if (preset.setOnConnect) {
  preset.setOnConnect(() => revalidate(RECONNECT_REVALIDATORS))
}
```

### Unsure Items (😭)

in the setup process, we attach the revalidators to window's focus/visible/online events, I try to change these attachings to customized handlers from config, but the default config cannot be changed at the beginning. unless users have a way to feed the preset to when swr code gets executed.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -46,7 +46,12 @@ const MUTATION_TS = {}
 const MUTATION_END_TS = {}
 
 // setup DOM events listeners for `focus` and `reconnect` actions
-if (!IS_SERVER && window.addEventListener) {
+setup({
+  setOnFocus: defaultConfig.setOnFocus,
+  setOnConnect: defaultConfig.setOnConnect
+})
+
+function setup(preset) {
   const revalidate = revalidators => {
     if (!defaultConfig.isDocumentVisible() || !defaultConfig.isOnline()) return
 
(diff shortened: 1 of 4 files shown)


src/use-swr.ts
```diff
@@ -46,7 +46,12 @@ const MUTATION_TS = {}
 const MUTATION_END_TS = {}
 
 // setup DOM events listeners for `focus` and `reconnect` actions
-if (!IS_SERVER && window.addEventListener) {
+setup({
```
huozhi, line 49: this call is executed at the very beginning of the code. so if we want to customize this to support RN. the default config should be able to change at the begining. but I'm not sure if it's a good way to expose defaultConfig out

shuding: Yeah, I think we can create 2 files which tackle this process:

```
src/
  use-swr.ts
  setup.ts
  setup.native.ts
```

And then inside `use-swr.js`, we simply `import './setup'`. IIRC RN will automatically import the native one.

(I never tested, need to verify)

shuding:
This PR looks shippable to me, do you have any other concerns since it's still a draft?

huozhi:
I can surly convert it to ready PR, we can carry the setup work in the further PRs. maybe we just don't document these configs first since they won't take effect outside of regular web browser runtime even you set the onFocus/onConnect.

src/use-swr.ts
huozhi: or could it become a preset?

```js
// on user land code
function rnPreset() {
   return {
     setOnFocus() {},
     isDocumentVisible() {}, 
     // ...
   }
}

SWRConfig.use(rnPreset()) // override the default config
```

since we renamed the context provider to `SWRConfig` for convenience, I'm wondering if it's possible to add more special methdos onto it. just a thought.

shuding: How about:

```jsx
<SWRConfig preset={[rn]} use={[...plugins]} config={cfg} />
```

shuding (approved):

src/use-swr.ts
huozhi: I feel provider doesn't solve this since currently stepup process is happened at the beginning without provider specified, which is before the hook gets render.

if we setup in the provider, will we still keep the default setup for web browser runtime?

PR #741, opened 2020-10-31
enable typescript strict mode

promer94:


src/use-swr-infinite.ts
```diff
@@ -29,58 +29,44 @@ type SWRInfiniteResponseInterface<Data = any, Error = any> = responseInterface<
...
     useContext(SWRConfigContext),
-    config
+    args.length > 2
+      ? args[2]
+      : args.length === 2 && typeof args[1] === 'object'
+      ? args[1]
+      : {}
   )
-  let {
+  // in typescript args.length > 2 is not same as args.lenth === 3
```
shuding, line 54: Thanks for the note!

shuding (approved):

PR #745, opened 2020-11-04
Fix `mutateCallback` types

umidbekkarimov:
~1. `Promise<Data | undefined>` because `mutate` implementation handles `undefined` values.~

`mutate` implementation can  pass `undefined` values to the `mutateCallback`: https://github.com/vercel/swr/blob/9e734aaff7d3082fc9fd7945f615fb63621b2bc7/src/use-swr.ts#L138-L144

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -64,8 +64,8 @@ export type triggerInterface = (
   shouldRevalidate?: boolean
 ) => Promise<any>
 export type mutateCallback<Data = any> = (
-  currentValue: Data
-) => Promise<Data> | Data
+  currentValue: undefined | Data
+) => Promise<undefined | Data> | undefined | Data
 export type mutateInterface<Data = any> = (
   key: keyInterface,
   data?: Data | Promise<Data> | mutateCallback<Data>,
--- a/test/use-swr.test.tsx
+++ b/test/use-swr.test.tsx
@@ -1493,6 +1493,26 @@ describe('useSWR - local mutation', () => {
     expect(callback).toHaveBeenCalledWith('cached data')
   })
 
+  it('should call function with undefined if key not cached', async () => {
+    const increment = jest.fn(currentValue =>
+      currentValue == null ? undefined : currentValue + 1
+    )
+
+    await mutate('dynamic-15.1', increment, false)
+
+    expect(increment).toHaveBeenCalledTimes(1)
+    expect(increment).toHaveBeenLastCalledWith(undefined)
+    expect(increment).toHaveLastReturnedWith(undefined)
+
+    cache.set('dynamic-15.1', 42)
+
+    await mutate('dynamic-15.1', increment, false)
+
+    expect(increment).toHaveBeenCalledTimes(2)
+    expect(increment).toHaveBeenLastCalledWith(42)
+    expect(increment).toHaveLastReturnedWith(43)
+  })
+
   it('should return results of the mutation', async () => {
     // returns the data if promise resolved
     expect(mutate('dynamic-16', Promise.resolve('data'))).resolves.toBe('data')


shuding:
Actually `mutate` doesn't accept `undefined` as the data. The implementation is for specific cases like `mutate(key)` (no data specified, only revalidate).

Generally we don't want you to do this:

```js
mutate(key, () => undefined)
```

Because in SWR `undefined` is like a special placeholder which means "no data available", so you can't set `undefined` as the data.

umidbekkarimov:
> Actually `mutate` doesn't accept `undefined` as the data. The implementation is for specific cases like `mutate(key)` (no data specified, only revalidate).
> 
> Generally we don't want you to do this:
> 
> ```js
> mutate(key, () => undefined)
> ```
> 
> Because in SWR `undefined` is like a special placeholder which means "no data available", so you can't set `undefined` as the data.

Thanks for clearing things out.

____

The main problem that we've faced is that sometimes when we're calling `mutate` with callback and `currentValue` comes as `undefined`, so `mutateCallback` should look like that: 

```diff
export type mutateCallback<Data = any> = (
-  currentValue: Data
+  currentValue: undefined | Data
) => Promise<Data> | Data
```
But this leads to another problem when we have only part of the value to update:

```ts
function updateStatus(guid: string, status: string): void {
  mutate(`/user/${guid}`, (user) => ({ ...user, status }));
}
```

It will create `{ status: string }` object when we do not have `currentValue`, so we want to check if `currentValue` is `undefined`

```ts
function updateStatus(guid: string, status: string): void {
  mutate(`/user/${guid}`, (user) => !user ? user : ({ ...user, status }));
}
```

Here TS will fail because it expects the result of the `mutateCallback` to be `Promise<Data> | Data`, so we have to change `mutateCallback` types to:

```diff
  export type mutateCallback<Data = any> = (
-  currentValue: Data
- ) => Promise<Data> | Data
+  currentValue: undefined | Data
+ ) => Promise<Data | undefined> | Data | undefined
```

umidbekkarimov:
@shuding hey, sorry to disturb, but can you check one more time please, and if it can't be merged I'll just close this PR.

shuding:
@umidbekkarimov Sorry for the delay!

Now I understand your use case and I think that makes sense. We can have it. 👍
The only request is can we add a test case for it, since it's a new behavior? It's fine if you are busy and we can take this PR and open an issue (for the test case) for now.

Thanks for the contribution!

umidbekkarimov:
> @umidbekkarimov Sorry for the delay!
> 
> Now I understand your use case and I think that makes sense. We can have it. 👍
> The only request is can we add a test case for it, since it's a new behavior? It's fine if you are busy and we can take this PR and open an issue (for the test case) for now.
> 
> Thanks for the contribution!

Done. I could add check TS because #741 was merged to `next`, but I add repro in the ts playground: https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcC2BXGBDGwDC2ANkQEbYDGA1gDwAiu2cAvHNgHaIB8LcAFACg4cChihRg7GADViGYAC44GdgBNgAMwCW7YKrgAfOAxwCAlCx4AFKBDRaAzsBor123fqMnsPI6806eobGjAKgkLAIyKgSDpDsTgCSUsBQGpTO3gA0cACi4tA8rADeQnCqjAD8St5lqbZQ1XkFUGUSAG7EWhV4SnwWzNa29k40pBAQRMAcXGWYOL38ZcI92E3ewTZ2jpmMvuhYuATEZJS03lxZy3AOABYQGESqAErAnUTdR03jk9PsZQMhttRhs-GoAh5ZsJHLIPj0dABzJQ-KYcAQAXwEAgA9Ni4AABGAOAC04WAFBgpJaAgoEAS8Fi8ScSkZdKSKTSGRo7AwaFIqRy+QaRTgxUxAlZCWAADp5kc+HxROJJDI5MBAaKykqJFJYfJpTAIAAxLQgIK4uDpLREBxlLHCC0AQQoFGAYCJynB7j0ZS0Gn4AEJtSq9erNcJhBIYGJ2J63IFVABuMriyPAaNQWPB3Vq5PosxAA

shuding (approved):
Thanks! This is already looking great

Zertz:
First, thanks for improving the types, this will definitely help make apps more robust!

I'm not sure where to ask this and I understand swr is 0.x but are type breaks like this considered a breaking change or more like a bug?

shuding:
@Zertz This PR is a bug fix and should be available in the latest release already (if not please let me know).

But regarding TS strict mode, I'd consider it a pretty breaking change because we need to revamped the core code base and the tests. The goal is to also have better type names in the 1.0 version.

PR #749, opened 2020-11-06
feature: setup once with first config and add test for reconnect options

huozhi:
### Changes

* only setup **once** across all hooks
* do setup in hook to prevent setup on JS parsing phase. then `setOnFocus` and `setOnConnect` will be available to customize

#### example

in useSWR `options` or in value of `SWRConfig`

```js
{
  setOnConnect((callback) => {
     addEventListener(eventName, callback)
     return () => { removeEventListener(eventName, callback) }
  })
}
```

if there's no such `setOnConnect` / `setOnFocus` from context, it will use setup with default preset. if there's one specified in context, the listeners are set up in the context provider

### Tests
* add custom `setOnConnect` config to support online events in jsdom env. since these options are more platform based, so I believe first setup will be enough

### DIFF
--- a/src/libs/events.ts
+++ b/src/libs/events.ts
@@ -1,0 +1,39 @@
+import { listenerInterface } from '../types'
+
+function addListener(
+  listenerMap: Record<string, listenerInterface[]>,
+  key: string,
+  callback: listenerInterface
+) {
+  if (!callback) return
+  if (listenerMap[key]) {
+    listenerMap[key].push(callback)
+  } else {
+    listenerMap[key] = [callback]
+  }
+}
+
+function removeListener(
+  listenerMap: Record<string, listenerInterface[]>,
+  key: string,
+  callback: listenerInterface
+) {
+  const listeners = listenerMap[key] || []
+  const index = listeners.indexOf(callback)
+  if (index >= 0) {
+    // 10x faster than splice
+    // https://jsperf.com/array-remove-by-index
+    listeners[index] = listeners[listeners.length - 1]
+    listeners.pop()
+  }
+}
+
+function invokeEvent(
+  listenerMap: Record<string, listenerInterface[]>,
+  key: string
+): void {
+  const listeners = listenerMap[key] || []
+  listeners.forEach(listener => listener())
+}
+
+export { addListener, removeListener, invokeEvent }
--- a/src/libs/web-preset.ts
+++ b/src/libs/web-preset.ts
@@ -32,7 +32,7 @@ const fetcher = (url: any) => fetch(url).then(res => res.json())
 const invokeFocus = () => invokeEvent(domListenersMap, 'focus')
 const invokeOnline = () => invokeEvent(domListenersMap, 'online')
 
-function useOnFocus(callback: listenerInterface): () => void {
+function setOnFocus(callback: listenerInterface): () => void {
   if (!isWindowEventTarget) return () => {}
   // when first listener attach
   if (hasNoListeners('focus')) {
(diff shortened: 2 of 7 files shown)


shuding (commented):
Do you think if we can move the `setup()` call into `SWRConfig`? I think that makes more sense given this is more like a global config.

```jsx
const SWConfig = ({value, children}) => {
  useEffect(() => {
    // we can setup global listeners here
  }, [])
  return <SWRContext.Provider value={value}>{children}</SWRContext.Provider>
}
```

huozhi:
> Do you think if we can move the `setup()` call into `SWRConfig`? I think that makes more sense given this is more like a global config.
> 
> ```js-jsx
> const SWConfig = ({value, children}) => {
>   useEffect(() => {
>     // we can setup global listeners here
>   }, [])
>   return <SWRContext.Provider value={value}>{children}</SWRContext.Provider>
> }
> ```

yeah for the customized listeners/events for sure we can do that, but I still want to find a way to keep the basic usage of useSWR simple. you can still only use the hook without SWRConfig:

* use hook without context: listeners can attach automatically once
* use hook with context(customized config): customized events trigger (i.e. customized online event) can replace the default behavior

any idea on that?

huozhi:
latest update:

setup event listeners in context if there're specified, else set up in the hook.
but the **ISSUE** is: it will setup for each hook, which means, the event listeners are attached multiple times if there's 

if there's no such `useOnXxx` from context, it will use setup with default preset. if there's one specified in context, the listeners are set up in the context provider, it means we bring the issue https://github.com/vercel/swr/issues/474 back..

huozhi:
image
added some logic on web preset to make sure the events only attach once

src/use-swr.ts
```diff
@@ -270,6 +257,26 @@ function useSWR<Data = any, Error = any>(
     configRef.current = config
   })
 
+  useEffect(() => {
+    if (swrContextConfig !== null) return
```
huozhi, line 261: if there's no context, attach revalidators for focus and connect

src/libs/events.ts
```diff
@@ -0,0 +1,39 @@
+import { listenerInterface } from '../types'
+
```
huozhi, line 2: extract the events attach/dettach as a utils set

src/use-swr.ts
```diff
@@ -778,7 +757,31 @@ function useSWR<Data = any, Error = any>(
   return memoizedState
 }
 
-const SWRConfig = SWRConfigContext.Provider
+function SWRConfig(props: {
+  children: React.ReactElement
+  value: ConfigInterface
+}) {
```
promer94, line 763: use react build-in type for children 
```ts
const SWRConfig: React.FC<{value: ConfigInterface}> = ({ value, children }) => {/** */}
```

src/libs/web-preset.ts
```diff
@@ -22,21 +29,48 @@ function isDocumentVisible(): boolean {
...
 const fetcher = (url: any) => fetch(url).then(res => res.json())
 
-function setOnFocus(callback: (...args: unknown[]) => void) {
-  if (!isWindowEventTarget) return
-  window.addEventListener('focus', callback, false)
-  window.addEventListener('visibilitychange', callback, false)
+const invokeFocus = () => invokeEvent(domListenersMap, 'focus')
+const invokeOnline = () => invokeEvent(domListenersMap, 'online')
+
+function useOnFocus(callback: listenerInterface): () => void {
```
promer94, line 35: keep the name ***setOnFoucs*** seems batter since this function is not a custom react hook.

src/use-swr.ts
huozhi: @promer94  seems I cannot pass the tsc check without declaring these 2 types?
image

promer94: This error message comes from eslint. since we are using typescript, this [***react/prop-types***](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md) rule is useless here. 

you can disable it in our eslint config by setting ```"react/prop-types": 0```

PR #754, opened 2020-11-08
refactor: support SSR in Deno

ije:
when use SSR with swr  in Deno i got a warning:

> Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.

```jsx
import useSWR from "https://esm.sh/swr"

export default function Home() {
    const { data } = useSWR('/api/hi', fetcher);

    if (!data) {
        return <div>loading...</div>
    }
    return <div>hello {data.name}!</div>
}
```

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -24,7 +24,7 @@ import {
   updaterInterface
 } from './types'
 
-const IS_SERVER = typeof window === 'undefined' || (typeof Deno !== 'undefined' && !!Deno?.version?.deno)
+const IS_SERVER = typeof window === 'undefined' || (typeof Deno !== 'undefined' && Deno !== null && !!Deno.version && !!Deno.version.deno)
 
 // polyfill for requestAnimationFrame
 const rAF = IS_SERVER
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -24,7 +24,7 @@ import {
   updaterInterface
 } from './types'
 
-const IS_SERVER = typeof window === 'undefined'
+const IS_SERVER = typeof window === 'undefined' || (typeof Deno !== 'undefined' && !!Deno?.version?.deno)
 
 // polyfill for requestAnimationFrame
 const rAF = IS_SERVER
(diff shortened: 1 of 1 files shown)


shuding (approved):
Looks amazing, thank you! Will test and merge later.

ije:
FYI please check https://alephjs.org/docs/basic-features/import-from-npm

src/use-swr.ts
```diff
@@ -24,7 +24,7 @@ import {
   updaterInterface
 } from './types'
 
-const IS_SERVER = typeof window === 'undefined'
+const IS_SERVER = typeof window === 'undefined' || (typeof Deno !== 'undefined' && !!Deno?.version?.deno)
```
huozhi, line 27: is `typeof Deno !== 'undefined'` ok enough for determining deno environment? 
I'm wondering if we can skip the `!!Deno?.versions?.deno` seems it's crashing the tests on CI

ije: for safety, better to check the `Deno.version.deno`, i  just updated hope it can pass the tests

huozhi (approved):

src/use-swr.ts
```diff
@@ -24,7 +24,7 @@ import {
   updaterInterface
 } from './types'
 
-const IS_SERVER = typeof window === 'undefined'
+const IS_SERVER = typeof window === 'undefined' || (typeof Deno !== 'undefined' && Deno !== null && !!Deno.version && !!Deno.version.deno)
```
shuding, line 27:
```suggestion
// @ts-ignore
const IS_SERVER = typeof window === 'undefined' || !!(typeof Deno !== 'undefined' && Deno && Deno.version && Deno.version.deno)
```

It's fine to ignore the TS error here, otherwise we need an extra TS plugin.

PR #757, opened 2020-11-10
fix: mark isValidating as false when key is falsy

### LINKED ISSUE #199: [Bug] The `isValidating` flag gets stuck on `true` if a request is cancelled
Not sure what's going on, but I just realized that the `isValidating` flag returned from `useSWR` often gets stuck on `true` when cancelling a request.

I've managed to boil it down to the following example: 

```ts
import React, { useState, useCallback } from 'react';
import useSWR from 'swr';

function Test() {
  const [input, setInput] = useState('');
  const handleChange = useCallback(e => setInput(e.currentTarget.value), [setInput]);
  const handleClear = useCallback(() => setInput(''), [setInput]);

  const { data, isValidating } = useSWR(
    input ? `https://jsonplaceholder.typicode.com/posts/${input}` : null,
    async key => {
      // NOTE: This is just to make a slow request. It happens with actual slow requests too.
      await new Promise(resolve => setTimeout(resolve, 2000));
      const res = await fetch(key);
      return res.json();
    }
  );

  return (
    <>
      <form>
        <input value={input} onChange={handleChange} />
        <button type="button" onClick={handleClear}>
          Clear
        </button>
      </form>
      <output>
        {isValidating
          ? 'Validating...'
          : JSON.stringify(data, undefined, 2) || typeof data}
      </output>
    </>
  );
}
```

* If I start typing, then `isValidating` will change to `true`.
* If I then clear the input _while_ `isValidating` is still `true`, either by just backspacing until empty, by clicking the button, then `isValidating` _never_ returns to `false`.
* If I start typing in the field again, then it "wakes up" because of a new request being triggered and `isValidating` returns to `true` when the new request finishes.
* If I have any input, and the request is completed, and I then clear it, then `isValidating` will remain `false` as expected.

What's going on? 😕

huozhi:
Fix #199 

~~* tear down work when key turns to empty, unmount state should be false if it's mounted before.~~
~~* dispatch `isValidating` as `false`~~

scenario: key: v1 -> v2 -> ''

~~in this case, when transfer to empty key the unmountedRef will be set to `true`, which lead dispatch take no effect. so I set it to true if it's not really unmounted.~~
~~then dispatch `isValidating` to `false` to ignore the ongoing requests.~~
~~update `keyRef` to match key, then isValidating getter will retrive states from `stateRef`~~

forcedly return false when key is falsy

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -537,7 +537,12 @@ function useSWR<Data = any, Error = any>(
 
   // mounted (client side rendering)
   useIsomorphicLayoutEffect(() => {
-    if (!key) return undefined
+    if (!key) {
+      if (keyRef.current !== key) unmountedRef.current = false
+      keyRef.current = key
+      dispatch({ isValidating: false })
+      return undefined
+    }
 
     // after `key` updates, we need to mark it as mounted
     unmountedRef.current = false
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -537,12 +537,7 @@ function useSWR<Data = any, Error = any>(
 
   // mounted (client side rendering)
   useIsomorphicLayoutEffect(() => {
-    if (!key) {
-      if (keyRef.current !== key) unmountedRef.current = false
-      keyRef.current = key
-      dispatch({ isValidating: false })
-      return undefined
-    }
+    if (!key) return undefined
 
     // after `key` updates, we need to mark it as mounted
     unmountedRef.current = false
@@ -724,7 +719,7 @@ function useSWR<Data = any, Error = any>(
       isValidating: {
         get: function() {
           stateDependencies.current.isValidating = true
-          return stateRef.current.isValidating
+          return key ? stateRef.current.isValidating : false
         },
         enumerable: true
       }
--- a/test/use-swr.test.tsx
+++ b/test/use-swr.test.tsx
@@ -2255,6 +2255,36 @@ describe('useSWR - key', () => {
       `"data-second"`
     )
   })
+
+  it('should cleanup state when key turns to empty', async () => {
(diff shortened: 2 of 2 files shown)


src/use-swr.ts
```diff
@@ -531,7 +531,12 @@ function useSWR<Data = any, Error = any>(
 
   // mounted (client side rendering)
   useIsomorphicLayoutEffect(() => {
-    if (!key) return undefined
+    if (!key) {
+      if (keyRef.current !== key) unmountedRef.current = false
```
shuding, line 535: I'm not sure if we should touch this. Since when `key` is falsy this effect returns `undefined` so it doesn't have the cleanup to reset this ref to true (https://github.com/vercel/swr/blob/7fe001c4174a8e28c5e1b7d8b8b38f941cfe39bf/src/use-swr.ts#L654).

shuding: The test looks great, thank you!

src/use-swr.ts
```diff
@@ -719,7 +719,7 @@ function useSWR<Data = any, Error = any>(
       isValidating: {
         get: function() {
           stateDependencies.current.isValidating = true
-          return stateRef.current.isValidating
+          return key ? stateRef.current.isValidating : false
```
huozhi, line 722: I can only come up with this simple solution that return `false` when key is falsy.

since when key changes swr will always clean up and `unmountedRef.current` changes to `true`. so dispatch won't work and in won't pass the internal condition.

shuding: This fix sounds good until we have a better and complete refactoring to the codebase. 👍

shuding (approved):
Thank you for the effort!

PR #760, opened 2020-11-11
update test

promer94:
Since waitForDomChange is deprecated, i would like to replace it gradually


shuding (commented):
There're 2 failing tests:

image

shuding (approved):

PR #767, opened 2020-11-12
Return '@null' if args is null ASAP

matamatanot:
If args[i] is null, always return `@null`.

Now, check `typeof args[i] === 'string'` and run `String(args[i])`. This is unnecessary.
This fix should improve readability and speed by a small amount.

### DIFF
--- a/src/libs/hash.ts
+++ b/src/libs/hash.ts
@@ -22,7 +22,6 @@ export default function hash(args: any[]): string {
       // "undefined" -> '"undefined"'
       // undefined   -> 'undefined'
       // 123         -> '123'
-      // null        -> 'null'
       // "null"      -> '"null"'
       if (typeof args[i] === 'string') {
         _hash = '"' + args[i] + '"'
--- a/src/libs/hash.ts
+++ b/src/libs/hash.ts
@@ -13,7 +13,8 @@ export default function hash(args: any[]): string {
   let key = 'arg'
   for (let i = 0; i < args.length; ++i) {
     if (args[i] === null) {
-      return (key += '@null')
+      key += '@null'
+      continue
     }
     let _hash
     if (typeof args[i] !== 'object' && typeof args[i] !== 'function') {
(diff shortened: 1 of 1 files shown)


matamatanot:
I have committed into `vercel:master`. Was `vercel:next` more appropriate? 
I'm skipping the error on ts-ignore using` --no-verify`. It seems to have already been fixed in `vercel:next`.

shuding:
@matamatanot yeah if it’s not a breaking change please target master. `next` is for 1.0 👍

matamatanot:
There is an error. The cause is #754.

shuding:
So this changes the hook behavior:

Before:
```js
useSWR([1, 2, null, 3, 4]) // => key will be `arg@1@2@null@3@4`
```

Now:
```js
useSWR([1, 2, null, 3, 4]) // => key will be `arg@1@2@null`
```

I think that feels inconsistent and may also cause key conflicts. But it is indeed a good question: should we skip the request if **one of the args** is `null`...?

matamatanot:
@shuding 
I mistook `for` for `forEach`. Is it from a performance perspective that used `for`?

matamatanot:
This is an example of using `reduce`. It's not clear if it improved readability and performance. Instead, it may be worse.

```
function hash2(args: any[]): string {
  if (!args.length) return ''
  return args.reduce((acc: string, current: any) => {
    if (current === null) {
      return (acc += '@null')
    } else if (typeof current !== 'object' && typeof current !== 'function') {
      if (typeof current === 'string') {
        return (acc += '@' + '"' + current + '"')
      } else {
        return (acc += '@' + String(current))
      }
    } else {
      if (!table.has(current)) {
        table.set(current, counter++)
        return (acc += '@' + (counter - 1))
      } else {
        return (acc += '@' + table.get(current))
      }
    }
  }, 'arg')
}
```

shuding:
Yeah I think as a small util, it's better to use `for` due to performance.

matamatanot:
Using the continue statement. If you don't find it useful, you can close this.

shuding (approved):
Thanks!

PR #777, opened 2020-11-17
Fix suspense

### LINKED ISSUE #494: React experimental concurrent issue

promer94:
fix #494

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -319,7 +319,7 @@ function useSWR<Data = any, Error = any>(
     if (shouldUpdateState || config.suspense) {
       // if component is unmounted, should skip rerender
       // if component is not mounted, should skip rerender
-      if (unmountedRef.current || !firstMountedRef.current) return
+      if (unmountedRef.current || !initialMountedRef.current) return
       rerender({})
     }
   }, [])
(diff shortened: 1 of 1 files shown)


promer94:
https://codesandbox.io/s/swr-494-ievc8

shuding (commented):
Thanks for the fix! Can confirm that error is gone. (BTW I suggest renaming this ref to `initialMountedRef`)

shuding (approved):

huozhi:
> Thanks for the fix! Can confirm that error is gone. (BTW I suggest renaming this ref to `initialMountedRef`, what do you think?)

yup the first unmountRef is kinda confused, not really for mount/unmount, but for dependency changes. maybe we can rename it later as well

shuding:
@huozhi yep I agree 👍

PR #787, opened 2020-11-23
fix: do mount check in config callback

promer94:
config callback may also cause ```Can't perform a React state update on a component that hasn't mounted yet```

## Repro Steps / Code Example
[Repo](https://codesandbox.io/s/long-sun-me0tz?file=/src/App.tsx)


shuding (approved):
Great catch, thanks!

PR #794, opened 2020-11-30
read cache during render

### LINKED ISSUE #387: [BUG?] Inconsistent value while doing conditional rendering.
I am doing conditional rendering. So when the data is not ready I don't show children.

**Expected behavior:**
Consistent value across components regardless of conditional rendering.

**Current behavior:**
Value update is not synchronized, previous value is shown in Children.

**Minimal reproduction:**
https://codesandbox.io/s/swr-mutate-bug-uryb6
When you click into the mutate button twice, you see inconsistent values in the console.

### LINKED ISSUE #463: useSWR calls with the same key in the same component yielding different data
# Bug report

Hi! First, I just wanted to say thank you for your work on this library 😊  It's been a ton of fun to use and has made both my development process and the product I'm working on faster.

useSWR calls in the same component with the same key yield different data

## Expected Behavior

I'd expect useSWR calls in the same component to yield the same data. I was using helper hooks and two helpers called useSWR with the same key. My code relied on both calls returning the same data. This led to a weird and unexpected bug that took multiple hours to figure out.

## Repro Steps / Code Example

https://codesandbox.io/s/swr-test-9jcdz?file=/src/App.js

## Additional Context

SWR version: 0.2.3

Thanks for your help!

### LINKED ISSUE #789: Error from different key is returned
# Bug report

When a request takes long to load, `useSWR` returns the error from the previous request, even though the keys are different.  

## Expected Behavior

`useSWR` should not return error from a different key. 

## Repro Steps / Code Example

1. Use Chrome
2. Install [URL Throttle chrome extension](https://chrome.google.com/webstore/detail/url-throttler/kpkeghonflnkockcnaegmphgdldfnden?hl=en) to simulate slow network for a certain URL.
3. Enable the extension and add this url: `https://pokeapi.co/api/v2/pokemon/ditto` with `10000` ms delay. Toggle on the `Enabled` switch.
4. Open the sample in this Codesandbox: [https://codesandbox.io/s/async-sound-9n2xm?file=/src/App.js](https://codesandbox.io/s/async-sound-9n2xm?file=/src/App.js)
5. First type in the input text: `fafafa`, or any non-pokemon text.
6. The text: `Error searching fafafa: fafafa not found` will appear.
7. Next, enter in the input text: `ditto`.
8. Since we add 10-second delay to the url to search for `ditto`, the request will take 10 seconds to finish. So I expect the text `Searching ditto` will appear, but instead the text: `Error searching ditto: dfadfa not found` appears. It seems `useSWR` returns the error from the previous key, `fafafa`.
9. After 10 seconds, the error text disappears as expected and the response is displayed.

Since the key changes, `useSWR` should not return error from previous request imo.

## Additional Context

SWR version: 0.3.9

promer94:
## The problem
In swr, we have stateRef to store the data from the global cache.

1. In first render, stateRef will read data from cache then render it.
2. It will start a revalidation after the first mount and register an updater for the mutate function.
3. During the revalidation, if the fetcher gives new data, the cache and stateRef will update and it will schedule a rerender. So useSWR will return the fresh data
4. when mutate is called, the cache will update synchronously and updater will schedule a rerender to make useSWR return the fresh data
5. when key is changed, it will start a revalidation. repeat 3.

So if the fetcher or the mutate function makes useSWR rerender, the stateRef is always sync with the cache.

But if useSWR rerender because of the sibling hooks rerender or the parent component rerender, it may return 'previous' data and stateRef is not sync with the cache

## Related issues
Fixes #387: the parents rerender because of the mutate function, the child component will render twice, the first render will give a stale value and the second render which is scheduled by the mutate function.

Fixes #463: the sibling hooks rerender because of the fetcher, so other hook will return the 'stale' value.

Fixes #789: the parents rerender because of the timer. so the previous error is returned from the stateRef.

## Repo
|      | react 17                               | react experimental                                     |
|------|----------------------------------------|--------------------------------------------------------|
| #387 | https://codesandbox.io/s/swr-387-hspvd | https://codesandbox.io/s/swr-387-concurrent-mode-3j8iy    |
| #463 | https://codesandbox.io/s/swr-463-9gtkq | https://codesandbox.io/s/swr-463-concurrent-mode-8ng8g |
| #789 | https://codesandbox.io/s/swr-789-m21g2 | https://codesandbox.io/s/swr-789-concurrent-mode-757nt |

## Possible solution

~~what if we do step 1 in every render ?~~

During the render, we pass the current cache to the dispatch function.
* if stateRef is same as cache, nothing will happen.
* if stateRef is not same as cache,  the stateRef will be update and react will schedule another render.
https://github.com/vercel/swr/blob/2248899df280e3d6fcd78bd57d0ed6938d82ce89/src/use-swr.ts#L306-L325

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -290,34 +290,6 @@ function useSWR<Data = any, Error = any>(
 
   const rerender = useState<unknown>(null)[1]
 
-  let dispatch = useCallback(
-    (payload: Action<Data, Error>) => {
-      let shouldUpdateState = false
-      for (let k in payload) {
-        // @ts-ignore
-        if (stateRef.current[k] === payload[k]) {
-          continue
-        }
-        // @ts-ignore
-        stateRef.current[k] = payload[k]
-        // @ts-ignore
-        if (stateDependencies.current[k]) {
-          shouldUpdateState = true
-        }
-      }
-
-      if (shouldUpdateState) {
-        // if component is unmounted, should skip rerender
-        // if component is not mounted, should skip rerender
-        if (unmountedRef.current || !initialMountedRef.current) return
-        rerender({})
-      }
-    },
-    // config.suspense isn't allowed to change during the lifecycle
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    []
-  )
-
   // error ref inside revalidate (is last request errored?)
   const unmountedRef = useRef(false)
   const keyRef = useRef(key)
@@ -326,15 +298,43 @@ function useSWR<Data = any, Error = any>(
   const initialMountedRef = useRef(false)
 
   // do unmount check for callbacks
-  const eventsCallback = useCallback(
-    (event, ...params) => {
+  const safeCallback = useCallback(
+    (callback: () => void) => {
       if (unmountedRef.current) return
       if (!initialMountedRef.current) return
       if (key !== keyRef.current) return
-      // @ts-ignore
-      configRef.current[event](...params)
+      callback()
     },
     [key]
+  )
+
+  let dispatch = useCallback(
+    (payload: Action<Data, Error>) =>
+      safeCallback(() => {
+        let shouldUpdateState = false
+        for (let k in payload) {
+          // @ts-ignore
+          if (stateRef.current[k] === payload[k]) {
+            continue
+          }
+          // @ts-ignore
+          stateRef.current[k] = payload[k]
+          // @ts-ignore
+          if (stateDependencies.current[k]) {
+            shouldUpdateState = true
+          }
+        }
+
+        if (shouldUpdateState) {
+          // if component is unmounted, should skip rerender
+          // if component is not mounted, should skip rerender
+          //if (unmountedRef.current || !initialMountedRef.current) return
+          rerender({})
+        }
+      }),
+    // config.suspense isn't allowed to change during the lifecycle
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [safeCallback]
   )
 
   const boundMutate: SWRResponse<Data, Error>['mutate'] = useCallback(
@@ -400,7 +400,7 @@ function useSWR<Data = any, Error = any>(
           )
         }
 
-        let newData
+        let newData: Data
         let startAt
 
         if (shouldDeduping) {
@@ -413,7 +413,8 @@ function useSWR<Data = any, Error = any>(
           // we trigger the loading slow event.
           if (config.loadingTimeout && !cache.get(key)) {
             setTimeout(() => {
-              if (loading) eventsCallback('onLoadingSlow', key, config)
+              if (loading)
+                safeCallback(() => configRef.current.onLoadingSlow(key, config))
             }, config.loadingTimeout)
           }
 
@@ -434,7 +435,7 @@ function useSWR<Data = any, Error = any>(
 
           // trigger the success event,
           // only do this for the original request.
-          eventsCallback('onSuccess', newData, key, config)
+          safeCallback(() => configRef.current.onSuccess(newData, key, config))
         }
 
         // if there're other ongoing request(s), started after the current one,
@@ -525,13 +526,15 @@ function useSWR<Data = any, Error = any>(
         }
 
         // events and retry
-        eventsCallback('onError', err, key, config)
+        safeCallback(() => configRef.current.onError(err, key, config))
         if (config.shouldRetryOnError) {
           // when retrying, we always enable deduping
-          eventsCallback('onErrorRetry', err, key, config, revalidate, {
-            retryCount: retryCount + 1,
-            dedupe: true
-          })
+          safeCallback(() =>
+            configRef.current.onErrorRetry(err, key, config, revalidate, {
+              retryCount: retryCount + 1,
+              dedupe: true
+            })
+          )
         }
       }
 
@@ -663,9 +666,6 @@ function useSWR<Data = any, Error = any>(
     addRevalidator(CACHE_REVALIDATORS, onUpdate)
 
     return () => {
-      // cleanup
-      dispatch = () => null
-
       // mark it as unmounted
       unmountedRef.current = true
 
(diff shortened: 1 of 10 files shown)


shuding:
> In first render, stateRef will read data from cache then render it.

This will possibly break the React hydration, we need to consider that. BTW, thanks for the sandboxes!!

promer94:
> > In first render, stateRef will read data from cache then render it.
> 
> This will possibly break the React hydration, we need to consider that. BTW, thanks for the sandboxes!!

if we only sync the stateRef with cache when key changes, would it possibly break the React hydration ?
```
if (initialMountedRef && key !== keyRef.current) {
    stateRef.current = {
      data: initialData,
      error: initialError,
      isValidating: initialIsValidating
    }
  }

File: src/use-swr.ts [hunk 1]
```diff
@@ -330,7 +331,13 @@ function useSWR<Data = any, Error = any>(
 
   // check if component is mounted in suspense mode
   const initialMountedRef = useRef(false)
-
+  if (initialMountedRef && keyRef.current !== key) {
```
shuding, line 334:
```suggestion
  if (initialMountedRef.current && keyRef.current !== key) {
```

shuding:
I don't know if we should do that (setting a ref's `.current` during render) but couldn't think of a better way to handle it...

src/use-swr.ts
```diff
@@ -330,7 +331,13 @@ function useSWR<Data = any, Error = any>(
 
   // check if component is mounted in suspense mode
   const initialMountedRef = useRef(false)
-
+  if (initialMountedRef.current && keyRef.current !== key) {
```
huozhi, line 334: I feel this change is trying to approach the similar functionality of the return values? wonder if it's possible to unify them together in one place to cover these corner cases

```js
error: {
  // `key` might be changed in the upcoming hook re-render,
  // but the previous state will stay
  // so we need to match the latest key and data (fallback to `initialData`)
  get: function() {
    stateDependencies.current.error = true
    return keyRef.current === key ? stateRef.current.error : initialError
  },
  enumerable: true
},
data: {
  get: function() {
    stateDependencies.current.data = true
    return keyRef.current === key ? stateRef.current.data : initialData
  },
  enumerable: true
},
isValidating: {
  get: function() {
    stateDependencies.current.isValidating = true
    return key ? stateRef.current.isValidating : false
  },
  enumerable: true
}
```

src/use-swr.ts
```diff
@@ -330,16 +305,74 @@ function useSWR<Data = any, Error = any>(
...
+   * ```
+   */
+  const dispatch = useCallback(
+    (payload: Action<Data, Error>) =>
+      safeCallback(() => {
+        let shouldUpdateState = false
+        const { data, error, isValidating } = payload
+        const keys = Object.keys(payload)
+        if (
+          keys.includes('data') &&
```
shuding, line 343:
```suggestion
        if (
          'data' in payload &&
```

src/use-swr.ts
```diff
@@ -268,7 +268,10 @@ function useSWR<Data = any, Error = any>(
   }
 
   const resolveIsValidating = () => {
-    return !!cache.get(keyValidating) || (key && willRevalidateOnMount())
+    return (
+      !!cache.get(keyValidating) ||
+      (key && willRevalidateOnMount() ? true : false)
```
huozhi, line 273:
```suggestion
      (key && willRevalidateOnMount())
```

src/use-swr.ts
```diff
@@ -330,16 +305,74 @@ function useSWR<Data = any, Error = any>(
...
+        let shouldUpdateState = false
+        const { data, error, isValidating } = payload
+        const keys = Object.keys(payload)
+        if (
+          keys.includes('data') &&
+          !config.compare(stateRef.current.data, data)
+        ) {
+          stateRef.current.data = data
+          shouldUpdateState =
+            shouldUpdateState || stateDependencies.current.data
```
huozhi, line 348: if you update data to null/false/undefined, will `shouldUpdateState` still be false?

src/use-swr.ts
```diff
@@ -330,16 +305,74 @@ function useSWR<Data = any, Error = any>(
...
+        }
+
+        // always update error
+        // because it can be `undefined`
```
huozhi, line 352: will checking `'error' in payload` avoid extra renders?

src/use-swr.ts
promer94:
```js
// In this way, the type of resolveIsValidating() would be boolean | ''
const resolveIsValidating = () => {
    return !!cache.get(keyValidating) || (key && willRevalidateOnMount())
}

// result = ''
const result = (() => {
    return !!false || ('' && false)
})()
```

huozhi: how about `!!(key && willRevalidateOnMount())` and can add a return type to resolveIsValidating?

src/use-swr.ts
shuding: That's `stateDependencies.current.data` (if the field is accessed by `render` or not) so should be OK.

src/use-swr.ts
```diff
@@ -213,149 +232,93 @@ function useSWR<Data = any, Error = any>(
...
+  const isValidating = resolveValidating()
 
   // do unmount check for callbacks
-  const eventsCallback = useCallback(
-    (event, ...params) => {
+  // if key changed during the revalidation, old dispatch and config callback should not take effect.
+  const safeCallback = useCallback(
+    (callback: () => void) => {
       if (unmountedRef.current) return
-      if (!initialMountedRef.current) return
```
promer94, line 320: I think we still need to this initialMountedRef check for config callback

https://codesandbox.io/s/swr-787-165v7

shuding: Good catch, thanks!

shuding (approved):

shuding:
Thank you so much for working on this! This is a pretty big change and lucky nothing breaks and the code is cleaner.

shuding:
Published as `swr@1.0.0-beta.1` :tada:

PR #796, opened 2020-12-05
Add cleanUpForTest() and a test

### LINKED ISSUE #781: SWR cache persistent across tests even with cache clearing and deduping interval at 0
# Bug report

I have found that some sort of SWR cache(?) persists between tests regardless of clearing the cache and setting the deduping interval to 0. The results are unpredictable and vary each time I run the tests.

## Expected Behavior

With cache clearing and deduping interval set to 0, I expected the components using SWR to return the correct data from the API for that test case, not from the previous test case.

## Repro Steps / Code Example
Code example here: https://github.com/frdwhite24/swr-testing
Example test script is here: https://github.com/frdwhite24/swr-testing/blob/master/src/component/FetchData.test.js

`git clone https://github.com/frdwhite24/swr-testing.git`
`cd swr-testing`
`npm install`
`npm test` (try running this 10 times and watch the different results)

## Additional Context

SWR v0.3.9
Jest as the testing lib

I have implemented mocking of my API using MSW (mock service worker) and am replacing an existing query handler with a runtime request handler. Starting out with a minimal code base, I tested out the MSW runtime request handler resetting with a vanilla fetch command. After this worked 100% of the time I introduced a component using SWR to get the data, and tests that run before, with and after I implement the runtime req handler. I found that one of three scenarios occur:

- All tests pass as expected (very rarely ~5%)
- The test with the runtime request handler returns data in the original handler and so will fail but all others will pass
- The test after the runtime request handler has been implemented but reset fails as it returns the new data set in the previous test's updated request handler, but all others will pass

I have tried putting a vanilla fetch between the SWR tests and this seems to help it return the correct data.
I have also put two SWR tests after each other which are identical, and the first one will fail but the second one will pass.
I have logged the SWR cache before and after each test and can see that it is being emptied.
I have logged the number of request handlers being used by MSW and have confirmed that they are as expected.
I have also forced MSW to reset the handlers between tests using the specific list of handlers used at the initial state.
I have tried running each test in isolation using `test.only` and each one passes 100% of the time.

I've been fighting this one for many days now and appreciate the time you spend looking into it!

petehunt:
Fixes #781. This adds a new `cleanUpForTest()` method that users should call in their jest `afterEach()` hooks.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -788,7 +788,27 @@ function useSWR<Data = any, Error = any>(
   return memoizedState
 }
 
+const cleanUpForTest = async () => {
+  let promises = new Set<Promise<any>>()
+  // eslint-disable-next-line no-constant-condition
+  while (true) {
+    const sizeBefore = promises.size
+    for (const promise of Object.values(CONCURRENT_PROMISES)) {
+      promises.add(promise as Promise<any>)
+    }
+    if (promises.size === sizeBefore) {
+      break
+    }
+    try {
+      await Promise.all(promises)
+    } catch (e) {
+      // swallow exceptions since they should be handled somewhere else.
+    }
+  }
+  cache.clear()
+}
+
 const SWRConfig = SWRConfigContext.Provider
 
-export { trigger, mutate, SWRConfig }
+export { trigger, mutate, SWRConfig, cleanUpForTest }
 export default useSWR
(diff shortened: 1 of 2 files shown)


nstepien:
I don't think this will handle cases where we expect multiple fetches to happen within one test with `dedupingInterval` set to `0`.

PR #799, opened 2020-12-08
First page of useSWRInfinite should reuse the cache from useSWR

shuding:
Currently `useSWRInfinite` always refetches the first page because of this line:

```js
(typeof force === 'undefined' && i === 0)
```

and it will not reuse the cache from `useSWR`. This PR fixes that so if there is no `originalData` (the first render), but has `pageData` (has cache), we directly return the cache.


huozhi (approved):

shuding:
Thanks @huozhi!

PR #814, opened 2020-12-14
fix: ensure `revalidateOnFocus` is work

u3u:
When using the same hook in multiple places, setting `revalidateOnFocus` in one place does not work

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -60,7 +60,9 @@ if (!IS_SERVER && window.addEventListener) {
     if (!defaultConfig.isDocumentVisible() || !defaultConfig.isOnline()) return
 
     for (const key in revalidators) {
-      if (revalidators[key][0]) revalidators[key][0]()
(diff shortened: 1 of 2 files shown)


shuding (commented):
This is a good point! But there're other issues like `onReconnect` isn't returning `true` or `false`, also the `revalidateOnReconnect` option has the same problem. 🤔

u3u:
Yes, it’s missing here, it’s been updated

huozhi:
@u3u would you mind giving an usage example to explain more about this issue? or it would be nice if supplement more information to the description!

u3u:
@huozhi 
In most cases, I want `revalidateOnFocus` to be turned off by default (this can avoid too many unnecessary requests). So I disabled it in `SWRConfig`. But under certain conditions, I want to enable it (such as a designated tab). Since this hook may be called in multiple places, if only the first hook is checked, the configuration of other components will be invalid

PR #817, opened 2020-12-15
Keep isValidating be true when there are two concurrent requests

MoonBall:
For case:
            req1 -----> res1
                 req2 ---------------->res2

The `isValidating` was false over the period of res1 and res2, and should be true.
This PR try to fix it.

See test case for detail


huozhi (commented):
lgtm on my side though, might want codeowner confirm if it's expected behavior. the change is introduced in #435

shuding (approved):
Great catch, thank you so much for adding the test and cleaning up the code. 👍

Also kudos to @huozhi for reviewing

PR #831, opened 2020-12-18
close #830

### LINKED ISSUE #830: RefreshInterval Config Not Respected After Errors
I have set up the useSWR hook in a Next.js application that is intended to continually refresh some API data. I was noticing that after some errors (502s technically, but any error is fine to reproduce) the `refreshInterval: 2000` I had set wasn't being used anymore and only manually re-focusing the page would cause a revalidation/refresh of the data. Even then, the `refreshInterval` didn't not pick back up after the error went away.

I have also tried manually specifying an `onErrorRetry` method which did not change the outcome:

```
onErrorRetry: (error, key, config, revalidate, { retryCount }) => {
    // Retry after 5 seconds.
    setTimeout(() => revalidate({ retryCount: retryCount + 1 }), 5000)
}
```

Any guidance here would be much appreciated.

## Expected Behavior

I was expecting the `refreshInterval` to be respected after the error retry logic was able to get a positive result again.

## Repro Steps / Code Example

https://codesandbox.io/s/gifted-gagarin-od5ut

## Additional Context

SWR version 0.3.9

promer94:
fix my mistake at #638 

Repro: https://codesandbox.io/s/swr830-65m4c?file=/pages/index.js

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -675,7 +675,7 @@ function useSWR<Data = any, Error = any>(
         await revalidate({ dedupe: true })
       }
       // Read the latest refreshInterval
-      if (configRef.current.refreshInterval && !stateRef.current.error) {
+      if (configRef.current.refreshInterval) {
         timer = setTimeout(tick, configRef.current.refreshInterval)
       }
     }


shuding (approved):
Thanks!

PR #833, opened 2020-12-19
Fix #832

### LINKED ISSUE #832: Mutate with custom updater when no useSWR seems to have wrong state
# Bug report

Simple case is:

- a user action that will trigger a mutate on a `key` with a **custom updater**
- after this action is done mount a component (or in a next js app push to router) that will use this `key`

-> the data that I see is not correct.

## Expected Behavior

Expected behavior are:

- if mutate is call and then a first `useSWR` is done on the same key, we should expect correct data -> request is correctly done, but data does not seems to be use / drop

## Repo Steps / Code Example

if you run this example: click on the button you will see that request is done -> but only the mutate data is use.

https://codesandbox.io/s/clever-knuth-2z40o?file=/src/App.js:1128-1161

## Additional Context

`0.3.9`

promer94:
close #832

Repro: https://codesandbox.io/s/swr832-ysc9m?file=/src/App.js

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -51,7 +51,7 @@ const MUTATION_END_TS = {}
 // generate strictly increasing timestamps
 const now = (() => {
   let ts = 0
-  return () => ts++
+  return () => ++ts
 })()
 
 // setup DOM events listeners for `focus` and `reconnect` actions
@@ -215,10 +215,6 @@ const mutate: mutateInterface = async (
       if (error) throw error
       return cache.get(key)
     })
-  } else {
-    // eslint-disable-next-line require-atomic-updates
-    MUTATION_TS[key] = undefined
-    MUTATION_END_TS[key] = undefined
   }
   // throw error or return data to be used by caller of mutate
   if (error) throw error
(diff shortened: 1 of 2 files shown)


shuding (commented):
After some local debugging, I found the case which caused this condition:

https://github.com/vercel/swr/blob/7858407a184dd273a4bc2158e270ea6a0ab5a4c7/src/use-swr.ts#L445-L463

is `MUTATION_END_TS[key] === 0`. I was surprised and actually, the root cause was because of another mistake here:

https://github.com/vercel/swr/blob/7858407a184dd273a4bc2158e270ea6a0ab5a4c7/src/use-swr.ts#L52-L55

So the [first call](https://github.com/vercel/swr/blob/7858407a184dd273a4bc2158e270ea6a0ab5a4c7/src/use-swr.ts#L138) of `now()` will get -1, and the [second call](https://github.com/vercel/swr/blob/7858407a184dd273a4bc2158e270ea6a0ab5a4c7/src/use-swr.ts#L191) will get 0. That's why `MUTATION_END_TS[key] === 0`. 

🤦

Could you change https://github.com/vercel/swr/blob/7858407a184dd273a4bc2158e270ea6a0ab5a4c7/src%2Fuse-swr.ts#L54 to `++ts` in this PR? I believe that will fix the bug.

Thanks for the test and looking into this btw, super helpful!

shuding (approved):
Thanks!

PR #834, opened 2020-12-20
test(fetch) add test for default fetcher

darrenjennings:
I wrote a test when porting over the default fetcher and thought I'd share it.

From: https://github.com/Kong/swrv/pull/120

### DIFF
--- a/test/use-swr.test.tsx
+++ b/test/use-swr.test.tsx
@@ -355,6 +355,32 @@ describe('useSWR', () => {
       `"hello, SWR"`
     )
   })
+
+  it('should use fetch api as default fetcher', async () => {
+    const users = [{ name: 'bob' }, { name: 'sue' }]
+    global['fetch'] = () => Promise.resolve()
+    const mockFetch = body =>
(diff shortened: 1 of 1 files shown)


test/setupJest.ts
```diff
@@ -0,0 +1 @@
+import 'whatwg-fetch'
```
huozhi, line 1: seems whatwg-fetch is not really been used? just for jest to call spyOn and override implementation. how about providing a fake fetch on nodejs `global` and clean it up when test is finished

```diff
 it('should use fetch api as default fetcher', async () => {
    const users = [{ name: 'bob' }, { name: 'sue' }]
+    global['fetch'] = () => Promise.resolve()
    const mockFetch = body =>
      Promise.resolve({ json: () => Promise.resolve(body) } as any)
    ...
    expect(container.firstChild.textContent).toMatchInlineSnapshot(
      `"hello, bob and sue"`
    )
+    delete global['fetch']
  })
```

darrenjennings: yes this is much better, thanks!

huozhi (approved):

huozhi:
Thank you!

PR #845, opened 2020-12-31
feature: provide config.isIdle indicating idle state and stop revalidation

huozhi:
### Description

give user ability to decide whether they want revalidation in some edge cases:

* page transition on slow connection that revalidations can cause more errors
* page is in background sometimes they don't want any revalidation

### Changes
* add `config.isPaused()`
* stop doing revalidation on `config.isPaused()` is `true`

related discussion [841](https://github.com/vercel/swr/discussions/841)

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -376,6 +376,7 @@ function useSWR<Data = any, Error = any>(
     ): Promise<boolean> => {
       if (!key || !fn) return false
       if (unmountedRef.current) return false
+      if (config.isIdle()) return false
       revalidateOpts = Object.assign({ dedupe: false }, revalidateOpts)
 
       let loading = true
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -376,7 +376,7 @@ function useSWR<Data = any, Error = any>(
     ): Promise<boolean> => {
       if (!key || !fn) return false
       if (unmountedRef.current) return false
-      if (config.isIdle()) return false
+      if (config.isPaused()) return false
       revalidateOpts = Object.assign({ dedupe: false }, revalidateOpts)
 
       let loading = true
@@ -499,6 +499,13 @@ function useSWR<Data = any, Error = any>(
       } catch (err) {
         delete CONCURRENT_PROMISES[key]
         delete CONCURRENT_PROMISES_TS[key]
+
+        if (config.isPaused()) {
+          dispatch({
+            isValidating: false
+          })
+          return false
+        }
 
         cache.set(keyErr, err)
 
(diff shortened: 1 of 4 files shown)


test/use-swr.test.tsx
```diff
@@ -1905,6 +1905,41 @@ describe('useSWR - context configs', () => {
     await act(() => new Promise(res => setTimeout(res, 110))) // update
     expect(container.firstChild.textContent).toMatchInlineSnapshot(`"data: 1"`)
   })
+
+  it('should stop revalidations when config.isIdle returns true', async () => {
+    let value = 0
+    let isSwrdle = false
```
shuding, line 1911:
```suggestion
    let isSwrIdle = false
```

test/use-swr.test.tsx
```diff
@@ -1905,6 +1905,41 @@ describe('useSWR - context configs', () => {
...
+    const fetcher = () => value++
+
+    function Page() {
+      // config provider
+      const { mutate: boundMutate, data } = useSWR('config-1', fetcher, {
+        revalidateOnMount: true,
+        isIdle() {
+          return isSwrdle
```
shuding, line 1919:
```suggestion
          return isSwrIdle
```

test/use-swr.test.tsx
```diff
@@ -1905,6 +1905,41 @@ describe('useSWR - context configs', () => {
...
+        }
+      })
+
+      useEffect(() => {
+        // revalidate on mount and turn to idle
+        isSwrdle = true
```
shuding, line 1925:
```suggestion
        isSwrIdle = true
```

shuding (commented):
This is a good start! We can have built-in `beforeunload` event handling and other features based on this. 2 questions:
Why using a function (`isIdle: () => boolean`) instead of a value (`isIdle: boolean`)? Value should be more consistent with our existing APIs. Another concern is that I think `pause` might be a better name, `idle` makes me feel that it means SWR is currently not active (not fetching anything).

huozhi:
@shuding Thanks for the nice feedback! I was thinking the `paused` state might be related to sth (like DOM state) out of react, i.e. swr can be paused when `document.visible` is false like `isOnline`.

will update PR by comments 🙏

shuding:
But the behavior of this option is to pause requests right (everything will be skipped)?

huozhi:
yup, I've changed to `config.isPuased()` to let user add some condition like 
```js
isPuased() { return document.visibility === 'hidden' }
```

and add one more case in revalidation catch branch, which is ignoring the errors when promises are been rejected under paused state

src/use-swr.ts
```diff
@@ -499,6 +500,13 @@ function useSWR<Data = any, Error = any>(
         delete CONCURRENT_PROMISES[key]
         delete CONCURRENT_PROMISES_TS[key]
 
+        if (config.isPaused()) {
+          dispatch({
+            isValidating: false
+          })
+          return false
+        }
```
shuding, line 508: Good catch! 👍

shuding (approved):

shuding:
Thanks, this is a good fix!

PR #853, opened 2021-01-02
fix: clear refresh timer on effects change and check it on starting new timer

### LINKED ISSUE #852: SWR keeps refreshing with old bearer token
# Bug report

I have encountered strange behavior when accessing endpoints with bearer token. I have a fetcher like this:
```javascript
const fetcherWithToken = async (url, token) => {
  try {
    const response = await fetch(url, {
      headers: {
        'Authorization': 'Bearer ' + token
      },
    })

    let text = await response.text()

    const data = text ? JSON.parse(text) : null

    if (response.ok) {
      return data
    } else if (response.statusCode === 401) {
      const error = new NotAuthorizedError(response.statusMessage);
      error.status = response.status
      throw error
    }

    const error = new Error(response.statusText)
    error.response = response
    error.status = response.status
    error.data = data
    throw error
  } catch (error) {
    if (!error.data) {
      error.data = { message: error.message }
    }
    throw error
  }
}
```

and a hook like this:
```javascript
const { data, error } = useSWR(
    ["https://api.example.com/xxx", token],
    fetcherWithToken,
    { refreshInterval: 5000 }
  );
```
After token refresh SWR keeps refreshing with both tokens - the old and the new.

## Expected Behavior

I would expect SWR to refresh only with the new token.

## Repro Steps / Code Example

Source code: https://codesandbox.io/s/eager-hoover-v4ryk?file=/pages/index.js

1) Let the app refresh data several times
2) Click at "Click me" button as soon as possible after a new message ("Fetching with Bearer token: X") in console appears
swr-keeps-refreshing-with-old-token

## Additional Context

swr: 0.3.11

I have come across this problem when building next.js application with [@react-keycloak/ssr](https://www.npmjs.com/package/@react-keycloak/ssr).

PS: I am not a JS developer so the code is a mess and it might just be PEBCAK.

huozhi:
Fix #852

### Description

https://github.com/vercel/swr/blob/00c5847ce6e95902920933aa1d49c7143f8ce4ba/src/use-swr.ts#L687-L697

when timer is been cancelled, but to the timer procedure, it cannot know if the last timer is cancelled. so it started a new timer again.

steps:
start timer > revalidate > timer is cancelled due to effects > but still start a new timer (with legacy `revalidate`).

### Changes

mark timer as null when cancel it and the cancel state is kept as condition. then swr can avoid to start a new timer again.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -671,7 +671,7 @@ function useSWR<Data = any, Error = any>(
   }, [key, revalidate])
 
   useIsomorphicLayoutEffect(() => {
-    let timer = null
+    const timer = { id: null }
     const tick = async () => {
       if (
         !stateRef.current.error &&
@@ -685,15 +685,18 @@ function useSWR<Data = any, Error = any>(
         await revalidate({ dedupe: true })
       }
       // Read the latest refreshInterval
-      if (configRef.current.refreshInterval) {
-        timer = setTimeout(tick, configRef.current.refreshInterval)
+      if (configRef.current.refreshInterval && timer.id) {
+        timer.id = setTimeout(tick, configRef.current.refreshInterval)
       }
     }
     if (configRef.current.refreshInterval) {
-      timer = setTimeout(tick, configRef.current.refreshInterval)
+      timer.id = setTimeout(tick, configRef.current.refreshInterval)
     }
     return () => {
-      if (timer) clearTimeout(timer)
+      if (timer.id) {
+        clearTimeout(timer.id)
+        timer.id = null
+      }
     }
   }, [
     config.refreshInterval,
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -671,7 +671,7 @@ function useSWR<Data = any, Error = any>(
   }, [key, revalidate])
 
   useIsomorphicLayoutEffect(() => {
-    const timer = { id: null }
+    let timer = null
     const tick = async () => {
       if (
         !stateRef.current.error &&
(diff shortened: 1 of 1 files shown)


shuding (commented):
This is a good catch! Indeed there's an async process inside `tick` so it's not safe to guarantee that cancel is safe. But I don't understand why you need an object to bind the latest timer value to the closure, I suppose just `timer` would work too (by checking if it's `null` on line 688)?

huozhi:
@shuding yeah it works well! thanks

shuding (approved):
Thank you so much 😊

promer94:
It seems i am a little late 😂.  I just find a solution
```js
    // Read the latest refreshInterval and make sure the key is not changed
      if (configRef.current.refreshInterval && key === keyRef.current) {
        timer = setTimeout(tick, configRef.current.refreshInterval)
      }
```

huozhi:
@promer94 that's a very good point! but the effects clean up can due to several change, we cannot guarantee the `key` will be the only factor in the long time.

promer94:
Another problem is that the actual refreshInterval is the ```response time + refreshInterval option``` 

We should mention this in the docs

shuding:
Seems some interval tests are failing

test/use-swr.test.tsx
```diff
@@ -543,41 +543,41 @@ describe('useSWR - refresh', () => {
...
-      fireEvent.click(container.firstElementChild)
       // it will clear 200ms timer and setup a new 300ms timer
       return new Promise(res => setTimeout(res, 200))
     })
-    expect(container.firstChild.textContent).toMatchInlineSnapshot(`"count: 2"`)
+    expect(container.firstChild.textContent).toMatchInlineSnapshot(`"count: 3"`)
     await act(() => {
       return new Promise(res => setTimeout(res, 110))
     })
-    expect(container.firstChild.textContent).toMatchInlineSnapshot(`"count: 3"`)
```
huozhi, line 555: @promer94 would you mind taking a look on those changes on this test case to see if there's any concern or mistakes? found they were created by you in #207 🙏

promer94: The behavior is changed here. When the refreshInterval changes after revalidating, the previous timer will not be clear because
```js
if (timer) {
  timer = null
  // clearTime wont take any effect here
  clearTimeout(timer)
}
```

I think it would be batter if we could keep the same behavior by

```js
if (timer) {
  clearTimeout(timer)
  timer = null
}
```

huozhi: oooh thanks for catching my obvious mistake 🙈👍

PR #779, opened 2020-11-18
update rest tests

promer94:

PR #872, opened 2021-01-08
test: fix all warnings in unit tests

koba04:
When we run unit testing, we see many warnings in the console output even all tests have been passed. This makes it difficult to find problems from the console output.

https://travis-ci.org/github/vercel/swr/builds/753389280

This PR fixed all warnings in unit testing.
I could do more refactoring, but I tried to keep the current implementation and changes minimum.

### DIFF
--- a/test/use-swr.test.tsx
+++ b/test/use-swr.test.tsx
@@ -28,14 +28,18 @@ class ErrorBoundary extends React.Component<{ fallback: ReactNode }> {
 describe('useSWR', () => {
   afterEach(cleanup)
 
-  it('should return `undefined` on hydration', () => {
+  it('should return `undefined` on hydration', async () => {
     function Page() {
       const { data } = useSWR('constant-1', () => 'SWR')
       return <div>hello, {typeof data === 'undefined' ? '' : 'ERROR'}</div>
     }
-    const { container } = render(<Page />)
-
-    expect(container.firstChild.textContent).toMatchInlineSnapshot(`"hello, "`)
+    await act(async () => {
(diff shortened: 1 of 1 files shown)


shuding (commented):
Thank you so much for fixing the warnings! cc @promer94 what do you think?

koba04:
@shuding This is off-topic, but what do you think about migrating CI from TravisCI to GitHub Actions? I think GitHub Actions is faster than TravisCI, so I'll send a PR for that if it could be accepted.

shuding:
@koba04 That's a great idea 👍

promer94:
@koba04 Thank you for fixing the warning. But I don't know if it should be addressed in this way.

```js
await 0
expect(....)
```

This pattern here feels to be a poor man's version of async [queries](https://testing-library.com/docs/dom-testing-library/api-queries#findby) provided by testing-library.  And i have already did a refactor on our next branch and fixed all the warnings by using async queries.  #779 

I highly recommend two articles to you which are [common-mistakes-with-react-testing-library](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library) and [fix-the-not-wrapped-in-act-warning](https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning).

koba04:
@promer94 Thank you! I didn't know that you have already fixed this on the next branch, so I'll close this.
 I'm not familiar with testing-library, so these resources are helpful for me, thank you!
Should I send a PR to the next branch rather than the master branch if there is an improvement point for the test code?

Issue #522, opened 2020-07-14
ReferenceError: navigator is not defined


niamleeson:
# Bug report

## Description / Observed Behavior

When building the app with NextJS in Jenkins, the build fails with the following error message:

```
ReferenceError: navigator is not defined
    at Object.641 (/app/node_modules/swr/dist/index.js:1:11816)
    at __webpack_require__ (/app/node_modules/swr/dist/index.js:1:154)
    at Object.516 (/app/node_modules/swr/dist/index.js:1:3661)
    at __webpack_require__ (/app/node_modules/swr/dist/index.js:1:154)
    at Object.325 (/app/node_modules/swr/dist/index.js:1:1394)
    at __webpack_require__ (/app/node_modules/swr/dist/index.js:1:154)
    at startup (/app/node_modules/swr/dist/index.js:1:291)
    at module.exports.202 (/app/node_modules/swr/dist/index.js:1:323)
    at Object.<anonymous> (/app/node_modules/swr/dist/index.js:1:333)
    at Module._compile (internal/modules/cjs/loader.js:959:30) {
```

## Expected Behavior

The build should succeed

## Repro Steps / Code Example

Or share your code snippet or a [CodeSandbox](https://codesandbox.io) link is also appreciated!

## Additional Context

NextJS 9.1.5
SWR 0.2.3

huozhi:
@niamleeson how do you run the tests and what are the testing tools you're using? from the log I saw the module resolved the cjs build of swr so I think it's testing running inside Node.js instead of browser runtime.

did you try to use jsdom to polyfill browser runtime stuff like (window, navigator, etc.)? or probably try jest with jsdom env enabled to run the tests?

niamleeson:
This was happening during development as well. I had to do the following in `_app.jsx`

```
import App, { Container } from 'next/app';
if (typeof global.navigator === 'undefined') global.navigator = {};
// const navigator = {};

class MyApp extends App {
    // ...
}
```

huozhi:
swr will access `window.navigator` in browser runtime only rather than the node.js or any server side. could you provide minial reproducible sample or code snippet for this issue? will be great helpful for us to identify the root cause.

niamleeson:
I tried really hard to get a reproducible code, but it looked like it has something to do with the compilation process. I even got to a place where I commented out one component that didn't use `useSWR` and the next build would finish successfully. Then I would uncomment the component back and the build would fail with the navigator message.

[...]

PR #877, opened 2021-01-09
fix: add protection for navigator

huozhi:
related to #522 

`navigator.isOnline` might be directly accessed in some corner cases leading to crashes. 

known runtime: 
* node.js with next.ks
* ink


huozhi:
UT `useSWR - refresh › should update data upon interval changes -- changes happened during revalidate` seems fails sometimes randomly on travisCI, not sure if it's about it's a bit slow to cause it. after manually restart the build, it gets back to normal

shuding (approved):

PR #879, opened 2021-01-09
ci: add GitHub Actions for CI

koba04:
This PR adds GitHub Action for typecheck, test, and lint tasks.
ref. https://github.com/vercel/swr/pull/872#issuecomment-756849888

The setup is based on the configuration of TravisCI.
https://github.com/vercel/swr/blob/master/.travis.yml

The result of a GitHub Action is here.
https://github.com/koba04/swr/runs/1673115189?check_suite_focus=true

If you have any requests like running on various environments (Node versions and OS), I'll add them to the configuration.
I have also left the setting for TravisCI, so let me know if you no longer need the setting, then I'll remove it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,21 @@
+name: TypeCheck, Lint, and Test
+
+on: [push]
+
+jobs:
+  test:
+
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


shuding (approved):
Thank you!

koba04:
@shuding Thank you!

Issue #163, opened 2019-11-29
Provide a ES5 version


KennyWho:
In `tsconfig.js`, the `target` is `esnext`, but in real world, we need support some device in iOS 9, so we perhaps need a ES5 version.

quietshu:
Should be fixed in v0.1.14.

nghiepit:
@quietshu In v0.1.15, I still see the syntax es6 in sources code build.

![image](https://user-images.githubusercontent.com/4768095/71306420-71792f00-2412-11ea-9775-a4faf1929282.png)

PR #881, opened 2021-01-09
Switch to microbundle

shuding:
This PR switched our bundler from ncc to microbundle. The benefits are:
- Smaller bundle: CJS output is now 28% smaller.
- Simpler pipeline: one command to build CJS, ESM and modern mode.

Also upgraded TypeScript and related packages, removed travis.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,21 +2,20 @@
   "name": "swr",
   "version": "0.4.0",
   "description": "React Hooks library for remote data fetching",
-  "main": "./dist/index.js",
-  "module": "./esm/index.js",
-  "react-native": "./esm/index.js",
-  "types": "./dist/index.d.ts",
+  "source": "src/index.ts",
+  "main": "dist/index.js",
+  "module": "dist/index.module.js",
+  "exports": "dist/index.modern.js",
+  "react-native": "dist/index.module.js",
+  "types": "dist/index.d.ts",
   "files": [
-    "dist/**",
-    "esm/**"
+    "dist/**"
   ],
   "repository": "vercel/swr",
   "homepage": "https://swr.vercel.app",
   "license": "MIT",
   "scripts": {
-    "build": "npm run build:esm && npm run build:cjs",
-    "build:cjs": "ncc build src/index.ts -o dist -m -e react",
-    "build:esm": "tsc --module ES6 --outDir esm",
+    "build": "microbundle --no-sourcemap -f modern,es,cjs",
     "watch": "tsc --watch",
     "types:check": "tsc --noEmit",
     "format": "prettier --write \"{src,test,examples}/**/*.{ts,tsx}\"",
@@ -41,9 +40,8 @@
     "@types/jest": "24.0.20",
     "@types/node": "11.12.0",
     "@types/react": "16.9.11",
-    "@typescript-eslint/eslint-plugin": "2.5.0",
-    "@typescript-eslint/parser": "2.5.0",
-    "@zeit/ncc": "0.20.5",
+    "@typescript-eslint/eslint-plugin": "4.12.0",
+    "@typescript-eslint/parser": "4.12.0",
     "eslint": "6.6.0",
     "eslint-config-prettier": "6.5.0",
     "eslint-plugin-react": "7.20.6",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -2,21 +2,20 @@
   "name": "swr",
   "version": "0.4.0",
   "description": "React Hooks library for remote data fetching",
-  "main": "./dist/index.js",
-  "module": "./esm/index.js",
-  "react-native": "./esm/index.js",
-  "types": "./dist/index.d.ts",
+  "source": "src/index.ts",
+  "main": "dist/index.js",
+  "module": "dist/index.module.js",
```
huozhi, line 7: how about also adding a modern bundle in output? that will be the minimal size of bundle output https://github.com/developit/microbundle#modern

```suggestion
  "module": "dist/index.module.js",
  "mordern": "dist/index.mordern.js",
```

huozhi (commented):
I double check the output, the cjs output still doesn't have the `__esModule` mark. 
if user end is consuming cjs bundle, `import useSWR from 'swr'` will fail again. 

regard to https://github.com/developit/microbundle/issues/531 this code transpiling behavior is still disabled.

package.json
shuding: It already has `"exports": "dist/index.modern.js",` though, I guess that's why the codesandbox example works? 🤔

huozhi:
FOLLOW UP:
if the bundler can resolve `module` fields first rather than `main` field, which most modern bundlers do today.
trace back to the previous swr issue with microbundle #487 , it seems the root cause is circular dependency + commonjs, and we've resolved it.

if we got passed on those outputs with next.js SSR, I think we can continue for sure!

shuding:
@huozhi You are absolutely right, it seems we got a problem (check the SWR-SSR template I just added):

image

huozhi:
@shuding  yep sad story, just found out next.js will prefer to consume `main` field on server side (nodejs + commonjs). https://github.com/vercel/next.js/blob/d72b2d6e36f951fdd0ed27053adf065342e506c3/packages/next/build/webpack-config.ts#L390

we need to make sure `__esModule` mark is insde swr commonjs bundle..

src/index.ts
```diff
@@ -14,4 +13,10 @@ export {
   responseInterface,
   CacheInterface
 } from './types'
-export default useSWR
+
+// A work around for the `__esModule` mark as microbundle
+// doesn't support mixed exports currently.
+// https://github.com/developit/microbundle/issues/306#issuecomment-459103476
+import * as exports from './use-swr'
+Object.defineProperty(exports, '__esModule', { value: true })
```
huozhi, line 21: looks like the variable `exports` is been resolved as another variable instead of cjs `exports` runtime variable by ts compiler.

// cjs bundle
```js
var E = s.Provider,
  x = { // the `exports` variable after compressed
    __proto__: null,
    trigger: R,
    mutate: j,
    SWRConfig: E,
    default: C
  };

...

Object.defineProperty(x, "__esModule", {
  value: !0
}), exports.SWRConfig = E, exports.cache ... // global exports is still there
```

we might need to find another workaround for it 😢

shuding: WTH... can we bypass TS here?

huozhi: manually adding the esModule mark for cjs build in postbuild script seems to work https://github.com/vercel/swr/pull/919

shuding: It feels too dirty, I wonder if we want to do that...

huozhi: removing main field and cjs bundle also works (for next.js case), since its webpack config on server mode can also include module field in mainFields.

but I don't know if there's other case that cjs bundle will be consumed only or it cannot be not widely adopted.

huozhi: few more updates to make this work:

* only keep `"module"` filed and add `type: "module"` can indicate the package as esm module.  reference: https://nodejs.org/api/packages.html#packages_type
* and rename `jest.config.js` to `jest.config.cjs` to still make jest run under commonjs
* also need to update changes from master (jest/ts-jest to 25.x, and few fixes for test) : )

I tried them in #919, tested with node v10 locally. it passed CI check with node v14. you can also rebase master then apply changes here, and close that one.

shuding: I personally feel that CJS is still a must for supporting v0.x (related: #163). But we can probably land this in v1.0... 💭

shuding:
Closing since we are going with another solution in #1005.

Issue #263, opened 2020-02-15
Is that correct the  effect dependencies are not exhaustive


soulcm:
In the [use-swr](https://github.com/zeit/swr/blob/master/src/use-swr.ts) file, I found some effect dependencies not exhaustive, I want to konw it is correct or not.

![image](https://user-images.githubusercontent.com/5706155/74588343-af6f8980-5036-11ea-9ee3-8550e0820aec.png)
If add eslint-plugin-react-hooks, and  add ```"react-hooks/exhaustive-deps": "warn"``` in your ```.eslintrc```  It will warn that ```React Hook useEffect has missing dependencies: 'config. suspense'```

the same as below code
```
const revalidate = useCallback(() => {
  // ... 
if (fnArgs !== null) {
            CONCURRENT_PROMISES[key] = fn(...fnArgs)
          } else {
            CONCURRENT_PROMISES[key] = fn(key)
          }
// ...
}, [key]) // the  fn args also not add the dependencies
```
anybody could make sense for me is that conrrect or not.  thanks a lot

koba04:
I've created a PR to fix this.

koba04:
This has been fixed via #886, so we could close this.

huozhi:
Close this as the warning has been disabled in latest commits. `config.suspense` is required to be static during the swr lifecycle.

PR #886, opened 2021-01-10
fix: react-hooks/exhaustive-deps warnings

koba04:
This is based on #263 and  I've fixed some warnings of `react-hooks/exhaustive-deps`, which can be safely fixed. 

- `renderer` is a setState function that `useState` returns, which is guaranteed to be stable. But the eslint plugin cannot recognize that `const rerender = useState<boolean>(false)[1]` is the setState function, so I've changed the statement.
  - https://reactjs.org/docs/hooks-reference.html#usestate
- The `dispatch` function uses `config.suspense`, but it's not in the deps array, so I've added it to the deps array.

This PR doesn't fix all warnings because some of them seem to be valid and adding missing deps variables might cause issues, so we have to fix the issues carefully and consider adding missing deps variables or `eslint-disable` comments. I'll work on that after this has been merged.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -302,7 +302,7 @@ function useSWR<Data = any, Error = any>(
   // display the data label in the React DevTools next to SWR hooks
   useDebugValue(stateRef.current.data)
 
-  const rerender = useState(null)[1]
+  const [, rerender] = useState(null)
   let dispatch = useCallback((payload: actionType<Data, Error>) => {
     let shouldUpdateState = false
     for (let k in payload) {
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -324,7 +324,9 @@ function useSWR<Data = any, Error = any>(
         rerender({})
       }
     },
-    [config.suspense]
+    // config.suspense isn't allowed to change during the lifecycle
+    // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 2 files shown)


src/use-swr.ts
```diff
@@ -302,27 +302,30 @@ function useSWR<Data = any, Error = any>(
...
-    }
-  }, [])
+      if (shouldUpdateState || config.suspense) {
+        // if component is unmounted, should skip rerender
+        // if component is not mounted, should skip rerender
+        if (unmountedRef.current || !initialMountedRef.current) return
+        rerender({})
+      }
+    },
+    [config.suspense]
```
shuding, line 327: We need to ignore this dependence because we don't allow the `suspense` option to change during the lifecycle, it has to be static (otherwise it would be difficult to track the internal state).

Maybe we should warn the user when they try to change the option.

koba04: @shuding Thank you for your feedback.  I've disabled the warning by the eslint-disable comment instead of adding it to the deps array.

> Maybe we should warn the user when they try to change the option.

That's a good idea. I think the warning isn't required on a production environment, but `swr` doesn't have a way to print warnings only for a development build. I'll propose it later.

shuding (approved):
Thank you!

koba04:
Thank you!

PR #889, opened 2021-01-11
fix: revalidateOnFocus not working on '< iOS 14' 

joshuakimDwan:
I found that revalidateOnFocus is not working on iOS 13. 
To fix, `window.addEventListener('visibilityevent', callback)` should be replaced by `document.addEventListener('visibilityevent', callback)`

Ref: https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event
image

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -55,7 +55,7 @@ const now = (() => {
 })()
 
 // setup DOM events listeners for `focus` and `reconnect` actions
-if (!IS_SERVER && window.addEventListener) {
+if (!IS_SERVER && window.addEventListener && document.addEventListener) {
   const revalidate = revalidators => {
     if (!defaultConfig.isDocumentVisible() || !defaultConfig.isOnline()) return
 
@@ -65,7 +65,7 @@ if (!IS_SERVER && window.addEventListener) {
   }
 
   // focus revalidate
-  window.addEventListener(
+  document.addEventListener(
     'visibilitychange',
     () => revalidate(FOCUS_REVALIDATORS),
     false
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -55,7 +55,7 @@ const now = (() => {
 })()
 
 // setup DOM events listeners for `focus` and `reconnect` actions
-if (!IS_SERVER) {
+if (!IS_SERVER && window.addEventListener && document.addEventListener) {
   const revalidate = revalidators => {
     if (!defaultConfig.isDocumentVisible() || !defaultConfig.isOnline()) return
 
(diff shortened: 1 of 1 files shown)


src/use-swr.ts
```diff
@@ -55,7 +55,7 @@ const now = (() => {
 })()
 
 // setup DOM events listeners for `focus` and `reconnect` actions
-if (!IS_SERVER && window.addEventListener) {
+if (!IS_SERVER && window.addEventListener && document.addEventListener) {
```
huozhi, line 58: this will crash on deno and RN runtime cuz they don't have `document` as an accessible global reference.

maybe we can do like this? 🤔 

```js
if (!IS_SERVER) {
  const revalidate = () => ...
  
  const hasWindow = typeof window !== 'undefined'
  const hasDocument = typeof document !== 'undefined'
  
  if (hasDocument) {
     document.addEventListener(...)
  }
  if (hasWindow) {
     window.addEventListener(...)
     window.addEventListener(...)
  }
}
```

this might not be very elegant, feel free to share any other approaches

joshuakimDwan: I think your solution is intuitive and readable. I will apply it.

joshuakimDwan: By the way, I don't know RN development environment. 
So, I have one question. Does RN have document object? If not, I will add logic to deal with it.

joshuakimDwan: (+) I have another question on "this will crash on deno"
`deno` is not considered a server environment?
I mean `!IS_SERVER` doesn't mean `is not on deno environment`?

huozhi: oops, right deno is included in that server side condition.

joshuakimDwan: I see. I applied your solution on 5fd759c

src/use-swr.ts
```diff
@@ -64,19 +64,39 @@ if (!IS_SERVER && window.addEventListener) {
...
-  window.addEventListener('focus', () => revalidate(FOCUS_REVALIDATORS), false)
-  // reconnect revalidate
-  window.addEventListener(
-    'online',
-    () => revalidate(RECONNECT_REVALIDATORS),
-    false
-  )
+  const hasWindow = typeof window !== 'undefined'
+  const hasDocument = typeof document !== 'undefined'
+  if (hasWindow) {
```
huozhi, line 69: I think the 2 conditions doesn't have to be nested?

```
if (hasDocument) {
   // focus revalidators
}
if (hasWindow) {
   // other revalidators
}
```

joshuakimDwan: RN environment also have `document` object like the web? 
I think that if RN doesn't have `document` object we should use `window.addEventListener`

huozhi: nope. it has window object but not document. 
I double check again locally with a RN app. it shows that window doesn't have `addEventListener`. and we need to check addEventListener. looks like your first commit also works well and it's shorter

I'll approve it on my side if can be reverted to the previous commit. sorry for my mistake and thanks for the question.

joshuakimDwan: Aha. Is this logic ONLY for web, not RN?

huozhi: yes, swr supports browser runtime by default. since react is been ported to different rumtime, people are trying adopt swr along with react apps. I was wondering if it's possible to make swr more flexible to be adopted among runtimes. seems not easy.

huozhi (approved):

mikevercoelen:
After merging this fix, it's now crashing when debugging react native and giving:

`document is not defined`

PR #890, opened 2021-01-11
refactor: use getter functions to avoid unnecessary re-renders

### LINKED ISSUE #780: Spread operator used in useSWRInfinite
# Bug report

https://github.com/vercel/swr/blob/b7052656c5c151e9c805c6074fe8f7aaa212cb29/src/use-swr-infinite.ts#L219

Using the spread operator here will trigger all the getters of the returned `swr` object, which causes unnecessary re-renders. A better way to do it is using `Object.assign`.

## Expected Behavior

## Repro Steps / Code Example

## Additional Context

SWR version.

koba04:
This PR fixes #780. I've left a comment on the intention.

### DIFF
--- a/src/use-swr-infinite.ts
+++ b/src/use-swr-infinite.ts
@@ -1,11 +1,4 @@
-import {
-  useContext,
-  useRef,
-  useState,
-  useMemo,
-  useEffect,
-  useCallback
-} from 'react'
+import { useContext, useRef, useState, useEffect, useCallback } from 'react'
 
 import defaultConfig, { cache } from './config'
 import SWRConfigContext from './swr-config-context'
@@ -191,8 +184,6 @@ function useSWRInfinite<Data = any, Error = any>(
     dataRef.current = swr.data
   }, [swr.data])
 
-  const boundMutate = useMemo(() => swr.mutate, [swr])
-
   const mutate = useCallback(
     (data, shouldRevalidate = true) => {
       if (shouldRevalidate && typeof data !== 'undefined') {
@@ -204,9 +195,9 @@ function useSWRInfinite<Data = any, Error = any>(
         cache.set(contextCacheKey, { force: true })
       }
 
-      return boundMutate(data, shouldRevalidate)
+      return swr.mutate(data, shouldRevalidate)
     },
-    [boundMutate, contextCacheKey]
+    [swr.mutate, contextCacheKey]
   )
 
   // extend the SWR API
@@ -226,11 +217,23 @@ function useSWRInfinite<Data = any, Error = any>(
   )
 
   // Use getter functions to avoid unnecessary re-renders caused by triggering all the getters of the returned swr object
-  return Object.assign(swr, {
+  return {
+    get error() {
+      return swr.error
+    },
+    get data() {
+      return swr.data
+    },
+    get revalidate() {
+      return swr.revalidate
+    },
+    get isValidating() {
+      return swr.isValidating
+    },
     mutate,
     size,
     setSize
-  }) as SWRInfiniteResponseInterface<Data, Error>
+  } as SWRInfiniteResponseInterface<Data, Error>
 }
 
 export {
--- a/src/use-swr-infinite.ts
+++ b/src/use-swr-infinite.ts
@@ -1,4 +1,11 @@
-import { useContext, useRef, useState, useEffect, useCallback } from 'react'
+import {
+  useContext,
+  useRef,
+  useState,
+  useMemo,
+  useEffect,
+  useCallback
+} from 'react'
 
 import defaultConfig, { cache } from './config'
 import SWRConfigContext from './swr-config-context'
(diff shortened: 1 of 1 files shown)


src/use-swr-infinite.ts
```diff
@@ -216,12 +216,12 @@ function useSWRInfinite<Data = any, Error = any>(
     [mutate, pageCountCacheKey]
   )
 
-  return {
-    ...swr,
+  // Use Object.assign to avoid unnecessary re-renders caused by triggering all the getters of the returned swr object
+  return Object.assign({}, swr, {
```
huozhi, line 220: Object.assign seems to have the same effect with object spread?

```js
var a = {
  get x() { console.log('getx'); return 'x' }
}

Object.assign({}, a) // "getx"
```

koba04: Thank you for your feedback!
You are right, to avoid triggering the getter function, I have to mutate `a` directly.

```js
var a = {
  get x() { console.log('getx'); return 'x' }
}

Object.assign(a, { foo: 'foo'}) // dosn't trigger the getter
```

But, that would break the current behavior, so I'll investigate a way to avoid triggering re-renders.

koba04: A way I came up with is to define getter functions to evaluate properties lazily.

```diff
diff --git a/src/use-swr-infinite.ts b/src/use-swr-infinite.ts
index 3b26b67..949cdc8 100644
--- a/src/use-swr-infinite.ts
+++ b/src/use-swr-infinite.ts
@@ -216,12 +216,16 @@ function useSWRInfinite<Data = any, Error = any>(
     [mutate, pageCountCacheKey]
   )
 
-  // Use Object.assign to avoid unnecessary re-renders caused by triggering all the getters of the returned swr object
-  return Object.assign({}, swr, {
+  // Use getter functions to avoid unnecessary re-renders caused by triggering all the getters of the returned swr object
+  return {
+    get error() { return swr.error },
+    get data() { return swr.data },
+    get revalidate() { return swr.revalidate },
+    get isValidating() { return swr.isValidating },
     mutate,
     size,
     setSize
-  }) as SWRInfiniteResponseInterface<Data, Error>
+  } as SWRInfiniteResponseInterface<Data, Error>
 }
 
 export {
```

koba04:
I've addressed the issue mentioned at https://github.com/vercel/swr/pull/890#discussion_r555124308.

src/use-swr-infinite.ts
shuding: Can you explain a bit why `Object.assign(swr, { ... })` won't work? It does trigger the getter:

image

koba04: @shuding 
The reason is that this mutates `swr` directly, and it breaks where the `swr.mutate` variable is used.
https://github.com/vercel/swr/blob/31215923ffd7eb8fef5948832876435d1aac9f1d/src/use-swr-infinite.ts#L198

It leads to failing many tests. I can fix it like the following, but the `useSWRInfinite › should keep mutate referential equal` test is still failing. Should I investigate the way to pass the test rather than using getter functions?

```diff
diff --git a/src/use-swr-infinite.ts b/src/use-swr-infinite.ts
index a0eda9c..ab0541e 100644
--- a/src/use-swr-infinite.ts
+++ b/src/use-swr-infinite.ts
@@ -184,6 +184,8 @@ function useSWRInfinite<Data = any, Error = any>(
     dataRef.current = swr.data
   }, [swr.data])

+  const _mutate = swr.mutate
+
   const mutate = useCallback(
     (data, shouldRevalidate = true) => {
       if (shouldRevalidate && typeof data !== 'undefined') {
@@ -195,9 +197,9 @@ function useSWRInfinite<Data = any, Error = any>(
         cache.set(contextCacheKey, { force: true })
       }

-      return swr.mutate(data, shouldRevalidate)
+      return _mutate(data, shouldRevalidate)
     },
-    [swr.mutate, contextCacheKey]
+    [_mutate, contextCacheKey]
   )
```

koba04: The failed test is here.
https://github.com/vercel/swr/blob/31215923ffd7eb8fef5948832876435d1aac9f1d/test/use-swr-infinite.test.tsx#L441

shuding: That's a good point! I think we need to find an elegant way to fix it then... let me see

koba04: @shuding  I agree with you, that's not an ideal solution. I'll try to find it. Thank you for your feedback!

shuding: Let's use the same logic like `use-swr` then:

```js
const swrInfinite = { size, setSize, mutate }
Object.defineProperties(swrInfinite, {
  error: {
    get: () => swr.error,
    enumerable: true
  },
  ...
})
```

The reason of not using the latest `get` syntax is that we have to specify `enumerable: true`.

koba04: I've updated specifing `enumerable: true`!

I also came up with the way to use `useMemo` like https://github.com/vercel/swr/pull/890/commits/116fced90a19a4b8b15bc811f09272ec399af8ed, but React says ["You may rely on useMemo as a performance optimization, not as a semantic guarantee. "](https://reactjs.org/docs/hooks-reference.html#usememo), so I've reverted the change.

shuding (approved):
Thanks for the help!

koba04:
Thank you!

Issue #588, opened 2020-08-06
useSWRInfinite loses initialData


valse:
Hi, with the deprecated `useSWRPages` I were able to set `initialData` (from my Next.js static props) and load only the next pages.
I'd like to do the same with the new `useSWRInfinite` but if I set the `initialData`, this is lost after the next load; here an example:

https://codesandbox.io/s/swr-infinite-initial-data-00qk4

I tried also to set `initialSize: 0` but with no luck :\

In the meantime I'm chaining the `initialData` with the result `data`

Thanks

BenoitAverty:
I'm not sure that this is a bug.

One possible intention with the initialData option is to provide the data of the first page.

When you scroll, useSwrInfinite tries to revalidate the first page, and it replaces the initial data with the content of the first page (it assumes that the first page has changed)

I'm not sure this is intended or not, it would be great if a maintainer could tell us the intention behind this option.

valse:
On the previous 0.2.3 release it works as expected; in [a SSR or SSG scenario](https://github.com/vercel/swr#ssr-with-nextjs) the initial data I think that it was came from the server and the next pages will be the other one.

doorman02:
Hi @valse @BenoitAverty @koba04 I am using useSWRInfinite with initialData set based on the first page data coming from SSR. However when I scroll down it reloads both page 1 and page 2. I would expect only page 2 to load. 
Is it intentional behaviour to reload page 1 even if it is already set as initialData? Thanks!

situplastik:
My "temporal solution" is put `initialSize=0` as @valse initial propose, and `initialData` (as array declared) inside the concatenation array:

https://codesandbox.io/s/swr-infinite-initial-data-forked-ppo2r

With that, it works as expected, hope that helps more developers. 

But would be great to admit initialData as we demand, as a simple array or and array with 1 position and the elements inside (following the return of the hook) simulating the first call. :)

koba04:
According to https://github.com/vercel/swr/pull/894#issuecomment-759645838, it is expected that SWR doesn't store `initialData` into the cache.

But I also understand that there are developers who expect that `initialData` is stored into the cache.
The following is a test case that is based on #894, but it avoids sending a request for the first page.
This returns `initialData` from the fetcher function if the request is for the first page.

```ts
  it('should re-use initialData and store it in the cache', async () => {
    const dummyResponses = {
      '/api?page=1': ['page-1-1', 'page-1-2'],
[...]

PR #894, opened 2021-01-12
test: add a test for useSWRInfinite with initialData

koba04:
This PR adds a test for `useSWRInfinite` with `initialData`.

I've added the test, but I'm still not sure of the expected behavior in the case.

How should `useSWRInfinite` behave combined with `initialData`?

- Does `useSWRInfinite` support `initialData`?
  - The `initialData` option is not mentioned in [the document](https://swr.vercel.app/docs/pagination#parameters), but [a related issue](https://github.com/vercel/swr/issues/588) is being tagged as `bug`.
- If `useSWRInfinite` supports `initialData`, how should the `initialData` be handled?
  - Is the `initialData` stored in the cache store? Currently, it seems not to be stored in the cache. If so, I'll fix the issue.
    - https://github.com/vercel/swr/compare/master...koba04:add-test-for-swrInfinite-with-initial-data?expand=1#diff-e7360ba75395f51ce8b969b7e11a5c6e78d58ce8f45a86976eafe53f747d7487R508
- If `useSWRInfinite` doesn't support `initialData`, I think `swr` should warn the usage of `initialData`. Does it make sense?

### DIFF
--- a/test/use-swr-infinite.test.tsx
+++ b/test/use-swr-infinite.test.tsx
@@ -463,4 +463,52 @@ describe('useSWRInfinite', () => {
     await act(() => new Promise(res => setTimeout(res, 10)))
     expect(container.textContent).toMatchInlineSnapshot(`"${cachedData}"`)
   })
+
+  it('should re-use initialData', async () => {
+    const dummyResponses = {
+      '/api?page=1': ['page-1-1', 'page-1-2'],
+      '/api?page=2': ['page-2-1', 'page-2-2']
+    }
+    let requests = []
+
+    function Page() {
+      const { data, size, setSize } = useSWRInfinite<string[], string>(
+        index => {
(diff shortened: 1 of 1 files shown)


shuding (approved):
Thanks, this is the expected behavior. `useSWRInfinite` should reuse `initialData` just like `useSWR` does because it's technically a wrapper of `useSWR`.

shuding:
And for the question:

> Is the initialData stored in the cache store? 

Nope, `initialData` is considered as a fallback value of `data`, it won't be pre-filled into the cache.

koba04:
Thank you!

shunkakinoki:
hi guys! if I want to pull `initialData` into from the`fallback` key of `SwrConfig` in the page root what is the correct key to put in it?

https://swr.vercel.app/docs/with-nextjs

PR #895, opened 2021-01-12
Fix the revalidation condition of the first page

shuding:
This PR fixes a `useSWRInfinite` bug introduced in #799 by me (thanks @pacocoursey for reporting). To reproduce, just simply use `refreshInterval` with `useSWRInfinite`, and the first page will never be revalidated:

Reproduction: https://codesandbox.io/s/swr-infinite-forked-mi7rf?file=/src/App.js

## Postmortem and Fix

The cause of the bug is this LOC: 

```js
(typeof force === 'undefined' && i === 0 && originalData) ||
```

...to determine if we should fetch this page or not, which can break down into:
- `typeof force === 'undefined`: It's not a force revalidation, like `mutate()`.
- `i === 0`: It's the first page, we don't want to revalidate every page upon focus.
- `originalData`: It's not the first render. **This is the mistake**.

What we want to achieve for `useSWRInfinite` is this experience: If we have the cache and the screen is empty (first render), we display the cache ASAP. So we added the third condition above. But we should check `dataRef` instead of `originalData` (`originalData` is specifically for mutations).

Also added another test.

### DIFF
--- a/src/use-swr-infinite.ts
+++ b/src/use-swr-infinite.ts
@@ -85,7 +85,7 @@ function useSWRInfinite<Data = any, Error = any>(
   // get the serialized key of the first page
   let firstPageKey: string | null = null
   try {
-    ;[firstPageKey] = cache.serializeKey(getKey(0, null))
+    [firstPageKey] = cache.serializeKey(getKey(0, null))
   } catch (err) {
     // not ready
   }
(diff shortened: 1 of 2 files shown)


huozhi (approved):
lgtm from test case perspective

PR #581, opened 2020-08-03
Improve test cases by separating swr integration test from unit tests 

B3nnyL:
Hi SWR crews,

I notice there are a mix of unit tests and integration level tests at use-swr.test.tsx. Separating them into different files can help future test implementation, also improve productivity for people who run tests in watch mode while implementing new features.

PR #901, opened 2021-01-14
separate tests

promer94:


shuding (approved):
Thank you so much!

koba04:
@promer94 Great! Is there any reason to keep duplicate tests, which have been split out to individual files, in `use-swr-test.tsx`?

test/use-swr-int.test.tsx
```diff
@@ -0,0 +1,354 @@
+import { act, render, screen } from '@testing-library/react'
```
koba04, line 1: What does the `int` in the file name (`use-swr-int.test.tsx`) mean?

huozhi: curious as well 😄 @promer94 do you mean `initial` or `basic`?

promer94: I keep the olds tests because i think #881 and #896 might need them as references.

The name comes from #581

koba04: @promer94 
Thank you!
Does the `int` means `integration`?
IMO, I prefer more explicit name like `use-swr-integration.test.tsx` because some people including me might misunderstand the meaning of the name. It's not a problem  If the name is common for an integration test.

shuding: Yeah I agree `use-swr-integration.test.tsx` would be better!

PR #902, opened 2021-01-14
refactor: fix all react-hooks/exhaustive-deps warnings in tests

koba04:
This is a PR to follow up #886 and fixes all `react-hooks/exhaustive-deps` warnings in tests.
They can safely be ignored in these cases.

As I mentioned on https://github.com/vercel/swr/pull/901#issuecomment-760216332, there are duplicated tests and I've fixed each warning twice.

I'll remove the tests if there is no reason for the duplication.


huozhi (approved):

koba04:
Thank you!

PR #903, opened 2021-01-15
test: recognize type definition for screen of testing-library/react

koba04:
Currently, `screen` isn't recognized because `@types/testing-library__react` doesn't have the type definition.

Screen Shot 2021-01-15 at 20 00 28

`jest` doesn't raise the error because `diagnostics` of `ts-jest` is disabled.

This can be fixed by updating `@testing-library/react` to v10 and I've enabled the `diagnostics` again.

But it causes another problem caused by a breaking change of `@testing-library/dom` v7.
> MutationObserver shim removed.
https://github.com/testing-library/dom-testing-library/releases/tag/v7.0.0

To fix this, I've updated `jest` packages to v25, which includes MutationObserver.

Unfortunatelly, some tests in `use-swr-test.tsx` are still failed. I can fix these tests, but they have already been fixed in separated files at #901.
I assume this PR would pass CI after #881 and #896 have been merged and removed duplicate tests.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,10 +1,5 @@
 module.exports = {
   preset: 'ts-jest',
   testRegex: '/test/.*\\.test\\.tsx$',
-  globals: {
-    'ts-jest': {
-      diagnostics: false
-    }
-  },
   modulePathIgnorePatterns: ['<rootDir>/examples/']
 }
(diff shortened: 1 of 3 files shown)


koba04:
Duplicate tests have been removed via #914, so this is now ready for review.

shuding (approved):
This is great, thank you!

koba04:
Thank you!

PR #905, opened 2021-01-15
refactor: fix all react-hooks/exhaustive-deps warnings in useSWRInfinite

koba04:
This is a PR to follow up #902 and #886, and fixes all react-hooks/exhaustive-deps warnings in `useSWRInfinite`.
I've disabled the warnings by `eslint-disable-next-line` comments and removed `swr.mutate` because it's an immutable reference.
https://github.com/vercel/swr/blob/81824c7df8666817075d386854711bfc153cc99a/src/use-swr.ts#L348-L353

### DIFF
--- a/src/use-swr-infinite.ts
+++ b/src/use-swr-infinite.ts
@@ -118,6 +118,8 @@ function useSWRInfinite<Data = any, Error = any>(
     } else {
       didMountRef.current = true
     }
+    // initialSize isn't allowed to change during the lifecycle
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [firstPageKey])
 
   // actual swr of all pages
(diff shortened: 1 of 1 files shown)


src/use-swr-infinite.ts
```diff
@@ -118,6 +118,8 @@ function useSWRInfinite<Data = any, Error = any>(
     } else {
       didMountRef.current = true
     }
+    // initialSize isn't allowed to change during the lifecycle
```
koba04, line 121: SWR doesn't support changing `initialSize` during the lifecycle, does it?

shuding: No it doesn't (it's hard to define what should happen if you change that option), maybe we can have a common util to show warning messages like this and #898?

shuding (approved):
Thank you!

koba04:
Thank you!

PR #907, opened 2021-01-15
fix: sync cache only on compare returns true

### LINKED ISSUE #906: compare function returning true still updates cache
# Bug report

When using a custom `compare` function and the comparison returns `true`, the `data` returned correctly keeps its current value, but the cached value still updates. This causes `data` to use the last received value when the component remounts and not the intended initial value we wanted to cache.

## Expected Behavior

From reading the documentation I was expecting the cache to be persisted if a comparison returns true.

## Repro Steps / Code Example

https://codesandbox.io/s/swr-using-compare-zn185?file=/src/App.tsx

1. click to mount the component
2. click "force revalidate" to... force some revalidates, and check `data` is correctly persisted.
3. click unmount to unmount the component
4. click mount again and check that `data` value is not the previously persisted value but the value of the last call

## Additional Context

In my use case, the server response contains urls of images with timestamps that I want to exclude from the data diff in order to avoid re-downloading images the browser has in disk cache.

swr v0.4.0

huozhi:
fixes #906

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -482,7 +482,6 @@ function useSWR<Data = any, Error = any>(
           return false
         }
 
-        cache.set(key, newData)
         cache.set(keyErr, undefined)
         cache.set(keyValidating, false)
 
@@ -499,6 +498,7 @@ function useSWR<Data = any, Error = any>(
           // deep compare to avoid extra re-render
           // data changed
           newState.data = newData
+          cache.set(key, newData)
         }
 
         // merge the new state
--- a/test/use-swr-refresh.test.tsx
+++ b/test/use-swr-refresh.test.tsx
@@ -1,6 +1,6 @@
 import { act, fireEvent, render, screen } from '@testing-library/react'
 import React, { useState } from 'react'
-import useSWR from '../src'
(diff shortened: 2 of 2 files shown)


shuding (commented):
Feels good to me, but this change assumes that `stateRef.current.data` equals to the current cache. I'm not sure if there's any case that it will break.

pablen:
Hi team, thanks for this quick PR :)
Is there anything that can be done to get it merged?

huozhi:
rebased to resolve the conflicts

huozhi:
> Feels good to me, but this change assumes that stateRef.current.data equals to the current cache. I'm not sure if there's any case that it will break.

I'd expect it won't be a problem with dequal since it supports various primtive data types. I feel it's ok to ship in a patch version and we can wait if there's any issue reports form community?

shuding (approved):
Gonna ship this once all tests have passed. Thank you!

pablen:
Thank you all!

promer94:
#1056 
The stateRef is not sync with cache when key changes in suspense mode. if two different keys happen to have the same response data. The cache of the second key would not be updated and it will cause infinite loop.

Issue #728, opened 2020-10-29
[RFC] Subscription mode for SWR


huozhi:
## Define the Problem

currently `useSWR` and `useSWRInfinite` are more designed for endpoints or data sources that can directly access and patch proactively. The data flow is like below

```
useSWR <=== write and read ====> data source
```

But for some cases SWR doesn't have enough solid ability to interact with data sources such as observable data (states returned from an observable stream), subscribable data (data been sent from a remote websocket endpoint).

then the `fetcher` of SWR will lose meaning since they're not really accessible initiavely from user end. we can surly do a `subscribe` or `observe` call to attach on the source. however we cannot know the state of a certain time.

besides `fetcher`, the `mutate`, `trigger` and `revalidate` APIs are also kind f disabled due to the passive mode. I want to find a new way to think of the communication path.

## Solution

propose to bring a new hook API `useSWRSubscription` to SWR to fit in the subscribable situation. unlike the basic useSWR hook, it won't return the manipulation APIs of useSWR.

#### API

```ts
function useSWRSubscription(key, subscribeFn: (onData, onError) => Disposable): {data?, error?};
```

### Usage

```js
// like the useSWRInfinite, we put it into an extra file
// or in the future we could make it import useSWRSubscription from 'swr/subscription'
import {useSWRSubscription} from 'swr'

const {data, error} = useSWRSubscription('wss://my.app/api/chat-messages', subscribeWs)

function subscribeWs(key, {onData, onError}) {
   const ws = new WebSocket(key)
   ws.onmessage = (event) => { 
      onData(event.data)
   }
   ws.onerror = (error) => {
      onError(event.data)   
   }
   return () => ws.close()
}
```

then in this case, if the key changed, SWR is able to close and re-subscribe to the source for new key;
the uncertain thing is the returned error might not always be valid since some obseravle won't throw error when the connection get closed;

### Other Possible Solutions

make revalidate and returned trigger with no effect. but mutate is still able to patch states to that key

### Pros
* support realtime data
* adaptive to any observable data source, user can also do customization upon it
* simple returned values to let user to leverage, no mind load on the meaning of revalidate/mutate/trigger

### Cons
* if there's any new API can access the current data, like geo API is subscribable but getting current position of geo is also doable. might need to have some workaround for that...
* more APIs into SWR, might bring confusion to new users to choose API

shuding:
Thank you for writing this amazing RFC! 

> `ws.onmessage = (event) => { return event.data };`

[...]

PR #913, opened 2021-01-20
[WIP] Subscribe mode

huozhi:
Draft solution to #728
To support subscription / disposable / observable data source, adding a new hook useSWRSubscription built on top of useSWR.

### API

```
useSWRSupscrition(key, subscribeFn: SWRSubscription, config?)
```

```ts
type SWRSubscription<Data> = (
  onData: (data: Data) => void,
  onError: (err: any) => void,
): () => void
```

the purpose is to provider users enough flexibility to manage their own subscription/disposable on user land, but still keep the argument shape different from function type fetcher.

### Usage

```js
const {data, error} = useSWRSubscription(key, (onData, onError) {
    const dispose = remoteDataSource.subscribe(onData, onError)
    return () => dispose()
 })
```

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 dist/**/*
 esm/**/*
 node_modules
+playground
--- a/.eslintrc
+++ b/.eslintrc
@@ -16,7 +16,9 @@
     "plugin:@typescript-eslint/eslint-recommended",
     "plugin:@typescript-eslint/recommended",
     "prettier",
-    "prettier/@typescript-eslint"
+    "prettier/@typescript-eslint",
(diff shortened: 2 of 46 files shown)


followbl:
would be awesome to get this in - we'd use it right away!

iamsaurabhgupt:
@huozhi @shuding @pacocoursey how far is this from getting into production?

huozhi:
Close this to file another one with middleware implementation

souljorje:
@huozhi hey! any updates about it?

huozhi:
The implementation is moved to https://github.com/vercel/swr/pull/1263

PR #856, opened 2021-01-03
check key before calling callback

promer94:
Fix: #854 
https://codesandbox.io/s/stale-onsuccess-call-doice?file=/pages/index.js

PR #914, opened 2021-01-20
chore: rm duplicated tests

huozhi:
delete the legacy test cases after the tidy up work from #901


shuding:
Thanks for renaming the file! Some tests were updated in #856, are those changes included?

huozhi:
@shuding good catch 👍, just updated.

Issue #738, opened 2020-10-31
concerns about passing null as fetcher 


promer94:
## Background
 [@pacocoursey 's idea for SWR as a local cache](https://paco.sh/blog/shared-hook-state-with-swr) is great.
```js
const useSharedState = (key, initial) => {
  const { data: state, mutate: setState } = useSWR(key, {
    initialData: initial
  })
  return [state, setState]
}
```
But after #367 this won't work anymore because window.fetch will be used when fn is undefined
https://github.com/vercel/swr/blob/10e18cc9e6e607309db1aee81ac22b58690a76c3/src/use-swr.ts#L245-L248

However passing null as fetcher will work since we ***only replace the fetcher to window.fetch when fetcher is undefined***.
```js
useSWR(key, null ,{ initialData: initial })
```
But the typescript will complain this kind of usage.  #730 fix this problem by accepting null as as the fetcher function.

## Problem
1. This usage is not well documented.
2. passing null as the fetcher function is like a hack here to disable the whole stale-while-revalidate process in swr. I am not sure it is a good api design. I feel like it somehow leaks the implementation details.
3.  Passing null as the fetcher function makes most of the options useless here.
 ```js
useSWR(key, null ,{ initialData: initial, /** most of options here will not be used since the revalidation is disabled here */ })
```
## Possible alternatives
1. a passive options and a different options type
```js
useSWR(key, { 
  initialData: initial, 
  passive: true 
  /** typescript should complain about set options like onSuccess 
  which will only be used at  stale-while-revalidate process  */
})
```
2. a new hooks share the caches
```js
const { data: state, mutate: setState } = useSWRPassive(key, { initialData })
```

loopmode:
I'd like to drop my two cents regarding API design and passing null for fetcher, even while it's only slightly related.
In my fetching hooks i typically use a concept where I can pass an option `autostart` or `autoload` which defaults to true, and my hooks typically return a `load` or `start` function, which can be used manually if autostart is set to false.
Maybe something similar could be considered in this library too. It creates a slightly larger API surface, but also provides huge flexibility and new features/possibilities, like submitting forms and any kind of user/gui-controlled fetching.

For this specific issue, it would mean instead of passing null for fetcher, you could just explicitly state that whatever the fetcher, it shouldn't be automatically invoked just because the hook was mounted.

TommySorensen:
[...]

PR #918, opened 2021-01-20
chore: add a comment why fetcher accepts null

koba04:
**UPDATED** This PR adds a comment to describe why `fetchFn` accepts `null`.

-----

This PR removes `null` from the type definition for `fetchFn`, but please correct me if I'm wrong.
I couldn't find any documentation about the behavior for passing null as a `fetcher` function. A case with `undefined` is documented though.
I also couldn't find any implementation for the case.

Is SWR supporting the case? If so, what is the expected behavior for the case?

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,3 +1,5 @@
+// `null` is used for a hack to manage shared state with SWR
+// https://github.com/vercel/swr/pull/918
 export type fetcherFn<Data> = ((...args: any) => Data | Promise<Data>) | null
 export interface ConfigInterface<
   Data = any,


promer94:
see #738 for more details

koba04:
@promer94 
Thank you for pointing it out. I start realizing SWR is used for various use-cases not only data fetching from this and https://github.com/vercel/swr/discussions/917.

As you mentioned at #738, it's a very hacking way, so I think it should be provided as an alternative API if SWR needs the feature.
I understand I cannot remove the type definition, so I've updated this PR to add a comment for the intention.

shuding:
Thanks @koba04 @promer94 ! It’s like a design debate on this specific API, there’re both pros and cons on having a default fetcher or not. 

And I agree that we should have an alternative way to do this. Using SWR for local data management is helpful and we are doing that too, passing null is just weird.

shuding:
I think a good way is to provide more helper hooks that are built on top of useSWR, with a good preset of configurations.

```js
import useSWRState from 'swr/state'

...

const { data, mutate } = useSWRState('key', initialVal)
```

where `useSWRState` is:

```js
useSWR(key, null, { initialData })
```

Just an idea, curious about what you guys think about it.

koba04:
I didn't know the use-case to use SWR for shared data management, so I'm not sure whether SWR should have the API or not. If so, exposing `swr/state` makes sense to me 👍 
But as the website says "SWR is a React Hooks library for remote data fetching", it seems to be out of scope for SWR. So I think SWR should consider adding the support very carefully.

shuding:
> But as the website says "SWR is a React Hooks library for remote data fetching", it seems to be out of scope for SWR. So I think SWR should consider adding the support very carefully.

Totally agree and thanks for the great reminder here! This is always the goal of this lib, to thoughtfully support more scenarios with better UX and great performance, with a minimal API surface. Let's consider that later carefully.

shuding (approved):

koba04:
Thank you!

PR #932, opened 2021-01-26
test: refactor use-swr-cache.test.tsx

koba04:
I've refactored `use-swr-cache.test.tsx` based on [Common mistakes with React Testing Library](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library).

I've sometimes experienced that the unit tests of SWR become flaky and are failed, but the failed tests got passed when I re-run the tests.
In addition, the tests are slow because they are often using `setTimeout` with 100ms or 200ms for `fetcher` functions. (This is not included in this PR because there is no test for the case, so I'll show you this as another PR)

To refactor it, I've installed some npm packages to set up. Please let me know if I should separate it from this PR.

- Using `@testing-library/jest-dom` for asserting DOM elements
- Using `eslint-plugin-jest-dom` and `eslint-plugin-testing-library` to avoid common mistakes.
  - The API of `testing-library` is a bit complex, so the plugins would be helpful.
- Refactoring `use-swr-cache.test.tsx`
  - I've replaced `queryByText` that is returned from `render` with `screen.findByText`
  - I've used `toBeInTheDocument` instead of `toMatchInlineSnapshot` because it can assert the test cases by verifying the textContent.

If this makes sense, I'll refactor all tests.


test/use-swr-cache.test.tsx
```diff
@@ -15,49 +15,32 @@ describe('useSWR - cache', () => {
...
       </React.Suspense>
     )
 
     // content should come from custom cache
-    expect(queryByText('custom cache message')).toMatchInlineSnapshot(`
-      <div>
-        custom cache message
-      </div>
-    `)
+    expect(await screen.findByText('custom cache message')).toBeInTheDocument()
```
promer94, line 25: if ```await screen.findByText('custom cache message')``` does not throw,  it means that the text content you are looking for is in the document. You don't need do another assertion here.

koba04: @promer94 
Thank you for your feedback! That's right.
But I prefer to have an assert function explicitly to indicate this line is an assertion.
What about writing like this?

```js
await expect(screen.findByText('custom cache messagea')).resolves.toBeTruthy()
```

This is the way `dom-testing-lijbrary` is using in the internal tests.
https://github.com/testing-library/dom-testing-library/blob/master/src/__tests__/queries.find.js#L41

(I don't have a strong opinion for this, so it's ok to change the way not using an assert function )

promer94:
```js
await expect(screen.findByText('custom cache messagea')).resolves.toBeTruthy()
```
In this way, we are actually testing the findByText API rather than our component.

koba04: Thank you!
I've fixed it at https://github.com/vercel/swr/pull/932/commits/d40345256c744a897d93c7b44222c29e19bbad89

test/use-swr-config-callbacks.test.tsx
```diff
@@ -88,13 +88,13 @@ describe('useSWR - config callbacks', () => {
     // props changed, but the onError callback doese not trigger yet.
     rerender(<Page text="b" />)
     expect(container.firstChild.textContent).toMatchInlineSnapshot(`"Error: 0"`)
-    expect(container.firstElementChild.getAttribute('title')).toEqual('b')
```
koba04, line 91: This raises an ESLint error and has been fixed automatically by the ESLint plugin.
In #933, this would be replaced with the way that doesn't use `container`, so it's a temporary solution.

shuding (approved):
Thanks!

koba04:
Thank you!

PR #933, opened 2021-01-26
 test: refactor use-swr-config-callbacks.test.tsx 

koba04:
This PR is based on #932, so this should be merged after #932.

The following are the commits for this PR.
- https://github.com/vercel/swr/commit/59d37fc2f93968adce0fbfa613eb3437875a31e2
- https://github.com/vercel/swr/commit/489d9f47e34580f246c3bf1498797a0dcf6f2008

I've refactored `use-swr-config-callbacks.test.tsx` as the same way with #932.
But in this PR, I've added the `createResponse` function in `test/utils.ts`. This is used to create a `fetcher` function.
It's to minimize the usage of `setTimeout` because tests that depend on the actual time is fragile, so I think we should avoid it and wait for a promise for a response from a `fetcher` function.

### DIFF
--- a/test/use-swr-config-callbacks.test.tsx
+++ b/test/use-swr-config-callbacks.test.tsx
@@ -7,15 +7,11 @@ describe('useSWR - config callbacks', () => {
   it('should trigger the onSuccess event with the latest version of the onSuccess callback', async () => {
     let state = null
     let count = 0
-    let promise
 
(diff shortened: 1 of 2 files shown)


shuding (approved):
Thanks for the refactoring!

koba04:
Thank you!

PR #937, opened 2021-01-29
chore: add comments that revalidate will be deprecated

koba04:
According to https://github.com/vercel/swr-site/pull/70#pullrequestreview-577603458, `revalidate()` will be deprecated in 1.x. I think it would be nice If it's commented as a source comment to tell the intention. It is helpful for contributors and developers who dig into the implementation.

Feel free to close this if it doesn't make sense 🙏


shuding (approved):
Thank you for adding those comments, really appreciated!

PR #946, opened 2021-02-02
types: makes types fn and config and key more strictly

koba04:
This is a PR to make internal types (`fn`, `config`, and `key`)  more strictly.
This helps refactoring and code completion.

In the future, it would be nice if SWR could enable the TypeScript's `strict` mode.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,42 +1,40 @@
-// `null` is used for a hack to manage shared state with SWR
-// https://github.com/vercel/swr/pull/918
-export type fetcherFn<Data> = ((...args: any) => Data | Promise<Data>) | null
+export type fetcherFn<Data> = (...args: any) => Data | Promise<Data>
 export interface ConfigInterface<
   Data = any,
   Error = any,
   Fn extends fetcherFn<Data> = fetcherFn<Data>
 > {
-  errorRetryInterval?: number
+  errorRetryInterval: number
   errorRetryCount?: number
-  loadingTimeout?: number
-  focusThrottleInterval?: number
-  dedupingInterval?: number
-  refreshInterval?: number
-  refreshWhenHidden?: boolean
-  refreshWhenOffline?: boolean
-  revalidateOnFocus?: boolean
+  loadingTimeout: number
+  focusThrottleInterval: number
+  dedupingInterval: number
+  refreshInterval: number
+  refreshWhenHidden: boolean
+  refreshWhenOffline: boolean
+  revalidateOnFocus: boolean
   revalidateOnMount?: boolean
-  revalidateOnReconnect?: boolean
-  shouldRetryOnError?: boolean
-  fetcher?: Fn
-  suspense?: boolean
+  revalidateOnReconnect: boolean
+  shouldRetryOnError: boolean
+  fetcher: Fn
+  suspense: boolean
   initialData?: Data
 
-  isOnline?: () => boolean
-  isDocumentVisible?: () => boolean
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -1,50 +1,48 @@
-// `null` is used for a hack to manage shared state with SWR
-// https://github.com/vercel/swr/pull/918
-export type fetcherFn<Data> = ((...args: any) => Data | Promise<Data>) | null
+export type fetcherFn<Data> = (...args: any) => Data | Promise<Data>
```
koba04, line 1: `null` is only accepted for `useSWR` and `useSWRInfinite` doesn't accept `null` as the fetcher, so I've moved `null` to the argument of `useSWR`.

src/types.ts
```diff
@@ -1,50 +1,48 @@
...
 export interface ConfigInterface<
   Data = any,
   Error = any,
   Fn extends fetcherFn<Data> = fetcherFn<Data>
 > {
-  errorRetryInterval?: number
+  errorRetryInterval: number
```
koba04, line 7: These values are having default values as `defaultConfig`, so we don't have to make the types optional.

src/use-swr-infinite.ts
```diff
@@ -34,39 +34,36 @@ function useSWRInfinite<Data = any, Error = any>(
...
 ): SWRInfiniteResponseInterface<Data, Error>
 function useSWRInfinite<Data = any, Error = any>(
   getKey: KeyLoader<Data>,
   fn?: fetcherFn<Data>,
-  config?: SWRInfiniteConfigInterface<Data, Error>
+  config?: Partial<SWRInfiniteConfigInterface<Data, Error>>
 ): SWRInfiniteResponseInterface<Data, Error>
 function useSWRInfinite<Data = any, Error = any>(
-  ...args
+  getKey: KeyLoader<Data>,
```
koba04, line 45: `getKey` is not an optional argument.

src/use-swr.ts
```diff
@@ -226,30 +226,29 @@ function useSWR<Data = any, Error = any>(
...
-  config?: ConfigInterface<Data, Error>
+  config?: Partial<ConfigInterface<Data, Error>>
 ): responseInterface<Data, Error>
 function useSWR<Data = any, Error = any>(
   key: keyInterface,
-  fn?: fetcherFn<Data>,
-  config?: ConfigInterface<Data, Error>
+  // `null` is used for a hack to manage shared state with SWR
+  // https://github.com/vercel/swr/pull/918
+  fn?: fetcherFn<Data> | null,
```
koba04, line 235: `null` should be accepted

promer94:
FYI. I had enabled the strict mode of typescript in the next branch. 😁

koba04:
@promer94 
Wow, that's so amazing!!!
Do you have any plan to merge the next branch into the master branch? What is the goal for the branch? v1?

shuding:
@koba04 We are aiming to v1 and we put all the things that might break the compatibility into that branch. But if you think that strict mode is safe to be added earlier, we can cherry-pick it and get it on master :)

shuding (approved):
Thanks!

koba04:
@shuding Thank you!
Do you have any plan to merge the next branch into the master branch?  I've found a milestone for v1, which means all issues in the milestone have to be fixed to release v1, don't they?
https://github.com/vercel/swr/issues?q=is%3Aopen+is%3Aissue+milestone%3A1.0

PR #961, opened 2021-02-03
fix: revalidate with initialData when changing the key

### LINKED ISSUE #947: SWR uses old key while using ssr
# Bug report

I'm using SWR and NextJS. I provide `initialData` via props and update key from state but it does not fetch with new key. Looks like it happens only if `initialData` is provided. 
Another strange thing: when state contains empty string refetching happens. But if you change state to any string it returns initial data.

## Expected Behavior

It should refetch data on state change.

## Repro Steps / Code Example

Here is an example in codesandbox:
https://codesandbox.io/s/dank-violet-himu9?file=/pages/index.js

koba04:
This fixes #947.
SWR seems not to revalidate when having `initialData`.

> Another strange thing: when state contains empty string refetching happens. But if you change state to any string it returns initial data.

This PR doesn't fix the above that is also mentioned on #947.
I'll investigate it as another issue.

### DIFF
--- a/test/use-swr-integration.test.tsx
+++ b/test/use-swr-integration.test.tsx
@@ -1,5 +1,5 @@
-import { act, render, screen } from '@testing-library/react'
-import React, { useEffect } from 'react'
+import { act, render, screen, fireEvent } from '@testing-library/react'
+import React, { useState, useEffect } from 'react'
 import useSWR from '../src'
 import { sleep } from './utils'
 
@@ -309,6 +309,34 @@ describe('useSWR', () => {
     expect(container.firstChild.textContent).toMatchInlineSnapshot(
       `"hello, Initial"`
     )
+  })
+
+  it('should revalidate even if initialData is provided', async () => {
+    const fetcher = key => key
+
+    function Page() {
+      const [key, setKey] = useState('initial-data-with-initial-data')
+      const { data } = useSWR(key, fetcher, {
+        initialData: 'Initial'
+      })
+      return (
+        <div onClick={() => setKey('initial-data-with-initial-data-update')}>
+          hello, {data}
+        </div>
+      )
+    }
+
+    const { container } = render(<Page />)
+
+    // render with the initial data
+    await screen.findByText('hello, Initial')
+
+    // change the key
+    await act(() => sleep(1))
+    fireEvent.click(container.firstElementChild)
+
+    // render with data the fetcher returns
+    await screen.findByText('hello, initial-data-with-initial-data-update')
   })
 
   it('should set config as second parameter', async () => {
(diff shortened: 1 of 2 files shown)


koba04:
The test has been passed in this PR, but I think this shouldn't be passed.
What does SWR return as `data` with `initialData` when being revalidating.
Currently, SWR returns `initialData` even it's not an initial mount.

```js
  it('should revalidate even if initialData is provided', async () => {
    const fetcher = async (key) => {
      sleep(100);
      return key
    }

    function Page() {
      const [key, setKey] = useState("initial-data-returns-empty-string")
      const { data } = useSWR(key, fetcher, {
        initialData: 'Initial'
      })
      return (
        <div onClick={() => setKey('initial-data-returns-empty-string-update')}>
          hello, {data}
        </div>
      )
    }

    const { container } = render(<Page />)

    // render with the initial data
    await screen.findByText('hello, Initial')

    await act(() => sleep(1))
    fireEvent.focus(window)

    await screen.findByText('hello, initial-data-returns-empty-string')

    // change the key
    await act(() => sleep(1))
    fireEvent.click(container.firstElementChild)

    await act(() => sleep(10))
    // THIS SHOULD BE FAILED, BUT IT'S PASSED
    expect(container.firstChild.textContent).toMatchInlineSnapshot(`"hello, Initial"`)

    // render with data the fetcher returns
    await screen.findByText('hello, initial-data-returns-empty-string-update')
  })
```

test/use-swr-integration.test.tsx
```diff
@@ -311,6 +311,46 @@ describe('useSWR', () => {
...
+
+    await screen.findByText('hello, initial-data-with-initial-data')
+
+    // change the key
+    await act(() => sleep(1))
+    fireEvent.click(container.firstElementChild)
+
+    // a request is still in flight
+    await act(() => sleep(10))
+    expect(container.firstChild.textContent).toMatchInlineSnapshot(`"loading"`)
```
koba04, line 348: I'm not sure what the expected value is.
But I expected SWR returns `undefined` as the return value of `data` when revalidating. But SWR returns the `initialData` Is it an expected behavior?

shuding: Yeah `initialData` works like a fallback value currently. I think we need to better name for it, maybe just `fallback`. And for the real `initialData`, we need a completely new API, which works like a prefill. (also cc @huozhi)

koba04: That seems to be great 👍  `fallback` would be more intuitive than `initialData` for the behavior.

shuding (approved):
The fix looks good, thanks!

koba04:
Thank you!

PR #963, opened 2021-02-04
fix: checking if document is defined

### LINKED ISSUE #953: SWR v0.4.1 crashes React Native when debugging
When debugging React Native, SWR v0.4.1 is causing the following error:

`document is not defined`

It's due to all changes where we use document without checking if it exists: 

https://github.com/vercel/swr/compare/0.4.0...0.4.1

```
if (!IS_SERVER && window.addEventListener) {

// ...

window.addEventListener(
```

Became this:

```
if (!IS_SERVER && window.addEventListener && document.addEventListener) {

// ...

document.addEventListener(
```

V0.4.0 is working fine

koba04:
Fixes #953.
This seems to be caused by the change of  #889.
I've added the check if `document` is defined before SWR uses `document.addEventListener`.

The check is already being used at https://github.com/vercel/swr/blob/70ffd2c37dfed75ab703e70af1fdad666889f13f/src/libs/web-preset.ts#L14

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -55,7 +55,12 @@ const now = (() => {
 })()
 
 // setup DOM events listeners for `focus` and `reconnect` actions
-if (!IS_SERVER && window.addEventListener && document.addEventListener) {
+if (
+  !IS_SERVER &&
+  window.addEventListener &&
+  typeof document !== 'undefined' &&
+  typeof document.addEventListener !== 'undefined'
+) {
   const revalidate = revalidators => {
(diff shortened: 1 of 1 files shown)


koba04:
If SWR wants to support `focus` and `online` events  on ReactNative, I should the the if statement only applies to the `visibilitychange` event.

shuding (approved):

shuding:
Thanks! I'm thinking about moving all the platform specific things into "presets", so they can be imported conditionally.

koba04:
Thank you! I agree with you 👍

PR #967, opened 2021-02-05
Optimize isValidating status

Kexin-Li:
This PR is set the initial `isValidating` to be true if there is a request will be sent on mount.

Currently, if we will fetching on mount, the `isValidating` will change like `false -> true -> false`. If we using `isValidating` to indicate a loading status, which is a bit weird.

I think it could be better to set the initial `isValidating` to `true` if we will start fetching. So it is `true -> false`.

If there is anything wrong, please correct me. 🙈 Thanks!


src/use-swr.ts
```diff
@@ -276,14 +276,25 @@ function useSWR<Data = any, Error = any>(
...
+    )
+  }
+
   const resolveData = () => {
     const cachedData = cache.get(key)
     return typeof cachedData === 'undefined' ? config.initialData : cachedData
   }
 
+  const resolveIsValidating = () => {
+    return !!cache.get(keyValidating) || willRevalidateOnMount()
```
huozhi, line 292: if `config.revalidateOnMount` is `true` then initial isValidating will always be `true`? The initial revalidation is executed in effects, I'd assume there's still a `isValidating = false` in the first render?

Kexin-Li: Yep yep I agree with you. But I'm thinking is that possible to let this feature built inside SWR?

shuding:
Thanks! I don't like the `false -> true -> false` state change either but my only concern is that this _might_ cause the [hydration mismatch](https://www.joshwcomeau.com/react/the-perils-of-rehydration). But I'm almost there's a way to determine and avoid that. I'll review and think about this again carefully.

chybisov:
I don't like the false -> true -> false either and hope this PR will be merged. @shuding did you have a chance to think about it? Thank you!

LRNZ09:
This would be very appreciated. It's odd to have the initial loading state set to `false`.

src/use-swr.ts
shuding: Yeah it avoids an extra re-render if we have this built-in :)

shuding, line 292:
```suggestion
    return !!cache.get(keyValidating) || (key && willRevalidateOnMount())
```

We need to also check `key` here because if it's falsy, the initial revalidation won't be triggered.

shuding (changes requested):

shuding:
Hi @chybisov @LRNZ09 @Kexin-Li! I think we can have this merged and release a beta version, once the above suggestion is merged 🙏 

Thanks for your effort @Kexin-Li @huozhi! And I need to test it a bit in a large application before releasing a stable version, because this change is quite critical. But once landed, this is gonna help the performance a lot!

Also, we need to update the docs when we release it: https://swr.vercel.app/advanced/performance

shuding (approved):
Thanks!

shuding:
Released as `0.4.3-beta.2` :tada:

PR #972, opened 2021-02-08
refactor: move DOM event handlers into web-preset

koba04:
This PR separates DOM event handlers from `use-swr` and defined them in `web-preset`.
ref. https://github.com/vercel/swr/pull/963#issuecomment-773380777

These handlers are only used on browsers and SWR already has a file for browser APIs. So these handers should be placed in the file.
This would also make it possible to provide `defaultConfig` for each environment, which enables us to only include implementations that are required on the platform.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -71,7 +71,9 @@ const defaultConfig: ConfigInterface = {
   fetcher: webPreset.fetcher,
   isOnline: webPreset.isOnline,
   isDocumentVisible: webPreset.isDocumentVisible,
-  isPaused: () => false
+  isPaused: () => false,
+  onFocus: webPreset.onFocus,
+  onReconnect: webPreset.onReconnect
 }
 
 export { cache }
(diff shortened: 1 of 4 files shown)


shuding (approved):
Thanks!

koba04:
Thank you!

huozhi:
Nice PR! Comparing to those events like configs `onError`, `onSuccess`, `onReconnect` and `onFocus` looks more like imperative actions to register callbacks to events, instead of being called as callbacks. I was trying to figure out more identical naming for them in #736 before. 

Some candidates in my minds are:  `registerOnFocus`, `setOnFocus`, `attachOnFocus`, it would be super clear if we can distinguish those events initializers from events callbacks.

koba04:
@huozhi Thank you for your feedback! I like your idea, so I've created https://github.com/vercel/swr/pull/993 to rename the names.

PR #977, opened 2021-02-12
fix: do not cache when mutate throws an error synchronously

### LINKED ISSUE #975: Mutator function ends up in cache if it throws an error
# Bug report

If you pass a mutator function to `mutate` and it throws an error the function itself will end up in the cache.

## Expected Behavior

Keep the old data in the cache.

## Repro Steps / Code Example

https://codesandbox.io/s/wizardly-tree-ghhb3?file=/src/index.js

## Additional Context

0.4.2

Cause:
https://github.com/vercel/swr/blob/0.4.2/src/use-swr.ts#L171

koba04:
Fixes #975.
But I'm not sure what the expected behavior when `mutate` throws an error synchronously.
As #975 mentioned, currently the `mutate` function is cached as the data, which is definitely an unexpected behavior.

This PR makes SWR skip caching the data in this case.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -155,6 +155,8 @@ const mutate: mutateInterface = async (
     try {
       _data = _data(cache.get(key))
     } catch (err) {
+      // if `_data` function throws an error synchronously, it shouldn't be cached
+      _data = undefined
       error = err
     }
   }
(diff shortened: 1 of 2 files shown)


shuding (approved):
Thanks!

koba04:
Thank you!

PR #979, opened 2021-02-15
perf: avoid unnecessary re-renders with the suspense mode

### LINKED ISSUE #454: Re-rendering too often with suspense mode
It seems like re-renders are happening too frequently when suspense mode is enabled. 

**Expected Behavior**
There should be one render to kick off the request, and then another render when the data is loaded. 

**Actual Behavior**
The component re-renders a total of 6 times. 4 times, the component throws to suspense. One more time after it has received the data. And then one more time for reasons I am unsure of.

[Here](https://codesandbox.io/s/lucid-mestorf-wi265?file=/src/App.js) is a minimum reproducible Sandbox.

koba04:
Fixes #454 
Currently, SWR doesn't apply the performance optimization mentioned in https://swr.vercel.app/advanced/performance with the `suspense` option, and it leads to unnecessary re-renders.

I've applied the same optimization using `stateDependencies` for the `suspense` mode.
Is there any reason not to apply the optimization for the `suspense` mode?

This is a fork of reproduction in https://github.com/vercel/swr/issues/454, which is based on this PR.
- https://codesandbox.io/s/swr-basic-forked-ftjio

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -318,7 +318,7 @@ function useSWR<Data = any, Error = any>(
         }
       }
 
-      if (shouldUpdateState || config.suspense) {
+      if (shouldUpdateState) {
         // if component is unmounted, should skip rerender
         // if component is not mounted, should skip rerender
         if (unmountedRef.current || !initialMountedRef.current) return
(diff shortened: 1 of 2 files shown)


shuding:
Sorry for the delay and thank you for this PR!

> Is there any reason not to apply the optimization for the suspense mode?

Mainly because that Suspense is still an experimental feature, and we are not implementing it 100% correctly (mostly with Concurrent mode). Also according to https://github.com/facebook/react/issues/17526#issuecomment-769151686, I feel that it's better to not optimize it too early, and get everyone relying on it too much.

koba04:
@shuding Thank you! I understand your concern.
I think rendering 6 times with a normal case (`const { data } = useSWR(url, fetcher, { suspense: true })`) is a bit surprising and might be a problem in complex applications. IMHO, this optimization doesn't rely on Suspense, which is just avoiding unnecessary rendering, so it would be nice if the suspense option has this optimization even this is an experimental option.

shuding (approved):

koba04:
Thank you!

PR #992, opened 2021-02-21
Sub-packages and immutable hook

shuding:
## `revalidateWhenStale`

This PR adds the `revalidateWhenStale` option to `useSWR` and it's `true` by default. If a resource has `revalidateWhenStale: false`, SWR will check if the cache, or `initialData` exists. If one of them is not `undefined`, SWR won't trigger the validation on mount.

### Idea behind this option

There're some types of resources we are sure that they will never change (during the application lifecycle), or we want to manually control their revalidation, instead of always revalidate on mount. This helps reduce the number of requests **for some specific data**.

We are explicitly calling it `revalidateWhenStale` because it follows the existing naming and, it has a smaller granularity that only controls one thing. Ideally we want all options for `useSWR` to be orthogonal, and we can build another hook layer with pre-configurations on top of them to have more scenarios covered.

## Sub-packages

We're adding sub-packages to SWR with this PR. You can import these 2 add-on hooks:

```js
import useSWRImmutable from 'swr/immutable'
import useSWRInfinite from 'swr/infinite'
```

The old `import { useSWRInfinite } from 'swr'` is removed. This enables better code-splitting/tree-shaking, so if `useSWRInfinite` isn't used in your project, it won't be bundled.

###  `useSWRImmutable`

It's a pre-configured `useSWR` hook, _basically_ the same as: 

```js
function useSWRImmutable(key, fetcher, config) {
  return useSWR(key, fetcher, {
    ...config,
    revalidateWhenStale: false,
    revalidateOnFocus: false,
    revalidateOnReconnect: false
  })
}
```

A middleware is included in the exports too so you can combine it with `useSWRInfinite`.

## Known issues

- Support of package.json `exports`
- https://github.com/microsoft/TypeScript/issues/33079
- https://github.com/microsoft/TypeScript/issues/9858

## To-do

- [x] #1160
- [x] Normally, this option needs to be set together with `revalidateOnFocus` and `revalidateOnReconnect`. We need to provide a `useSWRImmutable` hook to pre-config these options for you.
- [x] Deprecate `revalidateOnMount` in favor of this option.
- [x] More test cases.


huozhi (commented):
I wonder if we can delete the sub folders, package.json and yarn.lock. since we can rely them on `exports` in package.json. How about trying to put them all inside src folder.

```
- src/
   | - infinite/
   | - immutable/
```
package.json
```json
{
   "exports": { "./immutable": ..., "infinite": ... }
}
```

shuding:
@huozhi It seems there're still things don't support `exports` yet, for example csb doesn't seem to work...

CleanShot 2021-06-29 at 11 32 37@2x

infinite/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "swr-infinite",
+  "version": "0.0.1",
+  "main": "./dist/index.js",
+  "module": "./dist/index.esm.js",
+  "types": "./dist/infinite",
+  "peerDependencies": {
+    "swr": "*",
+    "react": "*",
+    "dequal/lite": "*"
```
huozhi, line 10:
```suggestion
    "dequal": "*"
```

bunchee should be able to resolve the sub folder imports

jest.config.js
```diff
@@ -3,6 +3,9 @@ module.exports = {
   testRegex: '/test/.*\\.test\\.tsx$',
   modulePathIgnorePatterns: ['<rootDir>/examples/'],
   setupFilesAfterEnv: ['<rootDir>/jest-setup.ts'],
+  moduleNameMapper: {
+    swr: '<rootDir>/dist/index.js'
```
huozhi, line 7: guess we should use source instead of dist for testing?

src/utils/config.ts
```diff
@@ -1,12 +1,13 @@
...
-import webPreset from './libs/web-preset'
+import webPreset from './web-preset'
 import { slowConnection } from './env'
-import { Configuration, RevalidatorOptions, Revalidator } from './types'
+import { Configuration, RevalidatorOptions, Revalidator } from '../types'
+import { UNDEFINED } from './helper'
 
 const fetcher = (url: string) => fetch(url).then(res => res.json())
-const noop = () => {}
+const noop = () => false
```
huozhi, line 10: not sure if we should change noop to return `false`, if there's a callback return defaultly to true but used noop by accident.

or maybe `const noop = (defaultValue) => () => defaultValue`

src/utils/hash.ts
```diff
@@ -12,24 +12,24 @@ export default function hash(args: any[]): string {
...
     let _hash
-    if (
-      args[i] === null ||
-      (typeof args[i] !== 'object' && typeof args[i] !== 'function')
-    ) {
-      // need to consider the case that args[i] is a string:
+    if (arg === null || (argType !== 'object' && argType !== 'function')) {
+      // need to consider the case that `arg` is a string:
       // "undefined" -> '"undefined"'
       // undefined   -> 'undefined'
```
huozhi, line 22: remove `undefined -> 'undefined'`? as JSON.stringify won't do

immutable/index.ts
```diff
@@ -0,0 +1,15 @@
+// @ts-ignore
+import useSWR from 'swr'
+
+import { withMiddleware } from '../src/utils/with-middleware'
+import { Middleware, SWRHook } from '../src/types'
+
+export const immutable: Middleware = useSWRNext => (key, fetcher, config) => {
+  // Always override all revalidate options.
+  config.revalidateOnFocus = false
+  config.revalidateWhenStale = false
```
huozhi, line 10: override revalidateOnMount here as well?

tsconfig.json
```diff
@@ -17,7 +17,8 @@
     "target": "es5",
     "typeRoots": ["./types", "./node_modules/@types"]
   },
-  "include": ["src/**/*"],
+  "include": ["src"],
+  "exclude": ["dist", "node_modules"],
```
huozhi, line 21: Is `execlude` necessary? it it for workaround on bundler?

jest.config.js
shuding: We have to do this, because the source code of infinite is using `import useSWR from 'swr'`.

immutable/index.ts
shuding: > @deprecated — revalidateOnMount will be removed. Please considering using the revalidateWhenStale option.

src/utils/config.ts
shuding: It's a hack to save some bytes... gonna revert it back. 👍

tsconfig.json
shuding: This is necessary otherwise I'll see this:

```
Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: infinite/dist/src/utils/resolve-args.d.ts.
The file must be included in at least one of the projects provided
```

huozhi (approved):
🚀🚀🚀

PR #1003, opened 2021-03-01
fix: sync mutations are applied independently

### LINKED ISSUE #994: Mutating state via onChange causes cursor jump to end of input
# Bug report

I'm attempting to use the [following code](https://github.com/vercel/swr/issues/402#issuecomment-630426798) as suggested by @sergiodxa to mutate state while typing in a text input.

The mouse cursor does not maintain its position within the text input field (if placed anywhere except at the end) while mutating the state. 

Controlling an input only with `React.useState` works as intended.

## Expected Behavior

Cursor should maintain it's current position within the text input.

## Repro Steps / Code Example

```js
import useSWR from "swr";

const fetcher = (url) => fetch(url).then((res) => res.json());

  const { data, mutate } = useSWR(
    "https://jsonplaceholder.typicode.com/todos/1",
    fetcher,
  );

  function handleChange(event) {
    mutate(
      {
        ...data,
        title: event.currentTarget.value
      },
      false
    );
  }

  if (!data) return <div>loading</div>;

  return (
    <div>
      <input value={data.title} onChange={handleChange} />
    </div>
  );
}
```

https://codesandbox.io/s/upbeat-thompson-qqeof?file=/src/App.js

## Additional Context

SWR version: 0.4.2

This bug seems to have been introduced in version [0.3.9](https://github.com/vercel/swr/releases/tag/0.3.9). I'm suspecting (without any scientific proof 🧑‍🔬) that it was introduced either by #777 or #735.

In version 0.3.8 and older it works as expected.

koba04:
fixes #994.
A codesandbox in #994 now works correctly. https://codesandbox.io/s/swr-basic-forked-qrbpc?file=/src/App.js

This is the same approach in https://github.com/vercel/swr/pull/735, so this causes unnecessary renders. But React batches updates in an event hander and has `unstable_batchedUpdates` for the API, which means developers can make those updates in a batched way.
On the other hand, the problem described in #994 is hard to avoid for developers and seems to be a more common case, so I think it should be fixed.


shuding:
Thanks for looking into this and creating a minimal reproduction in https://github.com/vercel/swr/issues/994#issuecomment-786002673! 

I'm a little bit worried that with this change, SWR no longer batches synchronous mutations and components will render intermediate result, which is another issue that we don't want to have (we can't ask users to always use `unstable_batchedUpdates`, and that API isn't officially stable). 

However I'm kinda 50/50 on this and it might be a good thing. Since React itself doesn't batch the results by default if you do 2 `setState`s in a row. So like you said, this should be something that developers should handle.

shuding:
I just verified that with this fix, intermediate results will not be rendered: https://codesandbox.io/s/swr-basic-forked-d78ie?file=/src/App.js (it won't log any odd numbers). It seems like React deduped them automatically, without `unstable_batchedUpdates`? 

Anyways this is a good fix, thank you very much!

koba04:
@shuding 
> I'm a little bit worried that with this change, SWR no longer batches synchronous mutations and components will render intermediate result, which is another issue that we don't want to have (we can't ask users to always use unstable_batchedUpdates, and that API isn't officially stable).

I agree with you and I have been hoping `unstable_batchedUpdates` would be a stable API over five years 😭 
Actually, I don't have enough confident for this change yet, so I would need more investigations.

> However I'm kinda 50/50 on this and it might be a good thing. Since React itself doesn't batch the results by default if you do 2 setStates in a row. So like you said, this should be something that developers should handle.

Yeah, this is the default behavior of React, so it makes sense for me than optimizing the update with undesirable edge cases.

> I just verified that with this fix, intermediate results will not be rendered: https://codesandbox.io/s/swr-basic-forked-d78ie?file=/src/App.js (it won't log any odd numbers). It seems like React deduped them automatically, without unstable_batchedUpdates?

React does batching updates in an event handler by default, so the updates in the CodeSandbox are batched.

shuding (approved):

shuding:
Thank you!

koba04:
Thank you!

PR #1004, opened 2021-03-01
Fix: do not depend on navigator.onLine; code optimizations

shuding:
As mentioned in the code comments, `navigator.onLine` isn't always reliable due to [this bug](https://bugs.chromium.org/p/chromium/issues/detail?id=678075). Instead of depending on it, we have to use another approach to always assume it's online (since the page loads, it's _very likely_ to be online). And when we receive the `offline` event from `window`, we set the internal state to offline.

With this change, SWR will still try to revalidate even it's offline, if **both** requirements are meet: 
- the app is running with Service Worker (or RN), and it's offline. 
- Chrome (or the navigator) is having the bug mentioned above.  

It's an extremely rare case.

Also did some code optimizations (results in a smaller bundle with ncc) and updated tests.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -47,34 +47,32 @@ const slowConnection =
   ['slow-2g', '2g'].indexOf(navigator['connection'].effectiveType) !== -1
 
 // config
-const defaultConfig: ConfigInterface = {
-  // events
-  onLoadingSlow: () => {},
-  onSuccess: () => {},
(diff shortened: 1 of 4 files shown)


huozhi (approved):

pacocoursey (approved):

shuding:
Released `0.4.3-beta.1`.

PR #919, opened 2021-01-20
Adopt micorbundle and use esmodule by default

huozhi:
* rm "main" field and cjs bundle
* add `"type": "module"`
* couple of fixes to make tests work again & rebase master

PR #1005, opened 2021-03-02
Use esbuild

shuding:
The only restriction is that we can't support ES5 as the target due to https://esbuild.github.io/api/#target.

Closes #881, #919.


shuding:
Closing since #1025 will be a better solution for now due to ES5 support.

PR #995, opened 2021-02-23
using vite2 and add playground to master

promer94:

PR #1009, opened 2021-03-05
fix: support multiple useSWRInfinite hooks in a page

### LINKED ISSUE #976: useSwrInfinite order of calling hooks and setSize matters
# Bug report

Let's say there are 2 components, A and B.
A and B are using custom hook which is returning useSwrInfinite props.
One of them, let's say B, is also setting size using setSize.

If you have:
```
return (
  <>
    <A />
    <B /> // setting size
  </>
)
```
setSize will increase size variable, but for some reason useSwrInfinite will not make a new fetch request for new page.

But if you switch order in which component setting size is rendered before:
```
return (
  <>
    <B /> // setting size
    <A />
  </>
)
```
then all works as intended.

## Expected Behavior

I should be able to set size anywhere and all my components using useSwrInfinite (and same key of course) should update.
Also data fetching should occur.

## Repro Steps / Code Example

https://codesandbox.io/s/lucid-frost-srix1?file=/src/App.js

```
import { useSWRInfinite } from "swr";

const useHook = () => {
  const { data, size, setSize } = useSWRInfinite(
    (pageIndex) => [`abc-${pageIndex}`, pageIndex],
    async (key, pageIndex) => {
      const response = await fetch(
        `https://api.github.com/repos/reactjs/react-a11y/issues?per_page=6&page=${pageIndex + 1}`
      )
        .then((response) => response.json())
        .then((data) => data);

      return response;
    }
  );

  return {
    data,
    size,
    setSize
  };
};

const A = () => {
  const { data, size, setSize } = useHook();

  return (
    <div>
      <div>
        {data &&
          data.map((item) =>
            item.map((it) => <div key={it.id}>{it.title}</div>)
          )}
      </div>
      <button onClick={() => setSize(size + 1)}>new page</button>
    </div>
  );
};

const B = () => {
  const { data } = useHook();
  console.log(data);

  return <div>B</div>;
};

  return (
    <div>
      <B />
      <A />
      {/* 
      // this works where A is before B
      <A />
      <B />
      */}
    </div>
  );
}
```

## Additional Context

React 17.0.1
SWR 0.4.2

koba04:
fixes #976
This PR fixes issues caused by using multiple `useSWRInfinite` on a page.
You can see the issues in the following link (v0.42).

- **v0.42**: https://codesandbox.io/s/swr-basic-forked-nw6z4
- **THIS PR**: https://codesandbox.io/s/swr-basic-forked-uo7r0

This PR also fixes #976.
https://codesandbox.io/s/swr-basic-forked-mbx6y

I guess this is a naive implementation and would change some behaviors of `useSWRInfinite`, so it should be merged carefully.

### DIFF
--- a/src/libs/web-preset.ts
+++ b/src/libs/web-preset.ts
@@ -1,5 +1,3 @@
-import { useEffect, useLayoutEffect } from 'react'
-
 /**
  * Due to bug https://bugs.chromium.org/p/chromium/issues/detail?id=678075,
  * it's not reliable to detect if the browser is currently online or offline
@@ -53,28 +51,10 @@ const registerOnReconnect = (cb: () => void) => {
   }
 }
 
-const IS_SERVER =
-  typeof window === 'undefined' ||
-  // @ts-ignore
-  !!(typeof Deno !== 'undefined' && Deno && Deno.version && Deno.version.deno)
-
-// polyfill for requestAnimationFrame
-const rAF = IS_SERVER
-  ? null
-  : window['requestAnimationFrame'] || (f => setTimeout(f, 1))
-
-// React currently throws a warning when using useLayoutEffect on the server.
-// To get around it, we can conditionally useEffect on the server (no-op) and
-// useLayoutEffect in the browser.
-const useIsomorphicLayoutEffect = IS_SERVER ? useEffect : useLayoutEffect
-
 export default {
   isOnline,
   isDocumentVisible,
   fetcher,
   registerOnFocus,
-  registerOnReconnect,
-  rAF,
-  useIsomorphicLayoutEffect,
-  IS_SERVER
+  registerOnReconnect
 }
--- a/src/libs/web-preset.ts
+++ b/src/libs/web-preset.ts
@@ -1,3 +1,5 @@
+import { useEffect, useLayoutEffect } from 'react'
+
 /**
  * Due to bug https://bugs.chromium.org/p/chromium/issues/detail?id=678075,
  * it's not reliable to detect if the browser is currently online or offline
@@ -51,10 +53,28 @@ const registerOnReconnect = (cb: () => void) => {
   }
 }
 
+const IS_SERVER =
+  typeof window === 'undefined' ||
+  // @ts-ignore
+  !!(typeof Deno !== 'undefined' && Deno && Deno.version && Deno.version.deno)
+
+// polyfill for requestAnimationFrame
+const rAF = IS_SERVER
+  ? null
+  : window['requestAnimationFrame'] || (f => setTimeout(f, 1))
+
+// React currently throws a warning when using useLayoutEffect on the server.
+// To get around it, we can conditionally useEffect on the server (no-op) and
+// useLayoutEffect in the browser.
+const useIsomorphicLayoutEffect = IS_SERVER ? useEffect : useLayoutEffect
+
 export default {
   isOnline,
   isDocumentVisible,
   fetcher,
   registerOnFocus,
-  registerOnReconnect
+  registerOnReconnect,
+  rAF,
+  useIsomorphicLayoutEffect,
+  IS_SERVER
 }
--- a/src/use-swr-infinite.ts
+++ b/src/use-swr-infinite.ts
@@ -1,5 +1,5 @@
 // TODO: use @ts-expect-error
-import { useContext, useRef, useEffect, useCallback } from 'react'
+import { useContext, useRef, useState, useEffect, useCallback } from 'react'
 
 import defaultConfig, { cache } from './config'
 import SWRConfigContext from './swr-config-context'
@@ -67,6 +67,8 @@ function useSWRInfinite<Data = any, Error = any>(
   } catch (err) {
     // not ready
   }
+
+  const [, rerender] = useState<boolean>(false)
 
   // we use cache to pass extra info (context) to fetcher so it can be globally shared
   // here we get the key of the fetcher context cache
@@ -187,6 +189,7 @@ function useSWRInfinite<Data = any, Error = any>(
         cache.set(pageCountCacheKey, arg)
         lastPageCountRef.current = arg
       }
+      rerender(v => !v)
       return mutate(v => v)
     },
     [pageCountCacheKey, resolvePageCount, mutate]
--- a/src/use-swr-infinite.ts
+++ b/src/use-swr-infinite.ts
@@ -87,14 +87,16 @@ function useSWRInfinite<Data = any, Error = any>(
     () => cache.get(pageCountCacheKey) || initialSize,
     [pageCountCacheKey, initialSize]
   )
+  const lastPageCountRef = useRef<number>(resolvePageCount())
 
   // every time the key changes, we reset the page size if it's not persisted
   useEffect(() => {
     if (didMountRef.current) {
-      if (!persistSize) {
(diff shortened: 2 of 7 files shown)


koba04:
Why has the CodeSandbox build been failed?

> ModuleNotFoundError
Could not find module in path: 'swr' relative to '/src/App.js'

promer94:
After #995 codesandbox ci seems broken 🥲 .  I am working on it

promer94:
https://github.com/vercel/swr/blob/66aaefddeaf3f11ae80b7480482769521bd364da/tsconfig.json#L25-L28 

Adding new item in include filed would change the esm directory structure 🙈
should be fixed in #1011

koba04:
@promer94 I got it, thank you!

koba04:
I've confirmed that this doesn't work, so I'll investigate more.
https://codesandbox.io/s/swr-basic-forked-s0dgl?file=/src/App.js

src/use-swr-infinite.ts
```diff
@@ -67,7 +67,8 @@ function useSWRInfinite<Data = any, Error = any>(
   )
   let {
     initialSize = 1,
-    revalidateAll = false,
+    // should revalidate pages when a refresh timer is invoked
+    revalidateAll = !!config.refreshInterval,
```
shuding, line 71: I'm not sure if we want to do this. Normally the `useSWRInfinite` API scenarios are cursor based, and in those cases we only need to revalidate the first page when polling because updates will likely happen in the newest page, and it will be too expensive to always revalidate all. Maybe it's better to let the user to enable it explicitly.

src/use-swr-infinite.ts
```diff
@@ -117,10 +118,11 @@ function useSWRInfinite<Data = any, Error = any>(
 
   // keep the data inside a ref
   const dataRef = useRef<Data[]>()
+  const lastFetchedDataRef = useRef<Data[]>()
 
   // actual swr of all pages
   const swr = useSWR<Data[], Error>(
-    firstPageKey ? ['many', firstPageKey] : null,
+    firstPageKey ? ['many', firstPageKey, pageCountRef.current] : null,
```
shuding, line 125: I don't think we should change the key here. It's OK to update both lists when clicking "new page" in the sandbox you shared, since the data (including all pages) is bound to the key.

koba04: @shuding Thank you for your feedback!

> It's OK to update both lists when clicking "new page"

I misunderstood the expected behavior. In that case, should SWR manage the page count in the cache rather than a ref in the component? Because the ref is stored in each component, so it seems to cause inconsistency between components.

```
A1 -> pageCount: 1, data: 1-1, 1-2, 1-3
A2 -> pageCount: 1, data: 1-1, 1-2, 1-3

A1 changes pageCount to 2

A1 -> pageCount: 2, data: 1-1, 1-2, 1-3, 2-1, 2-2, 2-3
A2 -> pageCount: 1, data: 1-1, 1-2, 1-3, 2-1, 2-2, 2-3 # pageCount should be 2

```

shuding: Yeah I think so! The page count is current kept in the cache:

```js
pageCountCacheKey = 'len@' + firstPageKey
```

But we also used a ref to handle it too, which brings the inconsistency I believe. Maybe we should only use the cached page count to manage that, and trigger a revalidation when it changes.

koba04: To support `persistSize: true`, we might have to keep something as a ref to persist the size between different keys.
https://github.com/vercel/swr/pull/490

shuding: Ah that's true, good catch! 👍 It makes things a little bit complicated here although.

koba04:
all tests now have been passed 🎉 But I have to refactor the implementation to get this merged.

src/use-swr-infinite.ts
koba04: I've addressed this.

src/use-swr-infinite.ts
```diff
@@ -185,23 +178,37 @@ function useSWRInfinite<Data = any, Error = any>(
...
-        pageCountRef.current = arg(pageCountRef.current)
+        cache.set(pageCountCacheKey, arg(resolvePageCount()))
+        lastPageCountRef.current = arg(resolvePageCount())
       } else if (typeof arg === 'number') {
-        pageCountRef.current = arg
+        cache.set(pageCountCacheKey, arg)
+        lastPageCountRef.current = arg
       }
-      cache.set(pageCountCacheKey, pageCountRef.current)
-      rerender(v => !v)
```
koba04, line 197: I didn't catch why the `rerender` was required, so please let me know if SWR still need to have it.

shuding: That's because we updated the size by changing a ref: `pageCountRef.current = arg`, which doesn't trigger a re-render (although it could be triggered by `mutate`, that might take a while). But the user needs to get the feedback, for example:

```js
const { size, setSize } = useSWRInfinite(...)

console.log(size)

return <button onClick={() => setSize(size + 1)}>inc</button>
```

When you click the button, it might not output the new page size.

koba04: Thank you!
In this case, `mutate(v => v)` seems to be enough to trigger a re-render.
https://codesandbox.io/s/swr-infinite-mv98u?file=/src/App.js
The CodeSandbox updates the `size` immediately.

src/use-swr-infinite.ts
```diff
@@ -188,23 +180,38 @@ function useSWRInfinite<Data = any, Error = any>(
...
+  useEffect(() => {
+    if (!didMountRef.current) {
+      didMountRef.current = true
+      return
+    }
+    cache.set(
+      pageCountCacheKey,
+      persistSize ? lastPageCountRef.current : initialSize
+    )
+    mutate(v => v)
```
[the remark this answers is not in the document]
koba04: Fixed as 6659aaa

koba04:
I wanted to re-use `useIsomorphicLayoutEffect` in `useSWRInfinite`, but I didn't want to expose it from `use-swr`, so I've moved it to `web-presets` and imported it.

koba04:
This is now ready for review again 🚀

src/libs/web-preset.ts
```diff
@@ -51,10 +53,28 @@ const registerOnReconnect = (cb: () => void) => {
...
 export default {
   isOnline,
   isDocumentVisible,
   fetcher,
   registerOnFocus,
-  registerOnReconnect
+  registerOnReconnect,
+  rAF,
+  useIsomorphicLayoutEffect,
+  IS_SERVER
```
shuding, line 79: I think these belong to web-preset, but shouldn't be included in `defaultConfig` because they are not "config", and we don't want to expose them to the context.

koba04: @shuding Thank you for your feedback! That makes sense 👍  What about exposing them as an external file like `env.ts`? Should I copy and have them as duplicated?

shuding: > What about exposing them as an external file like env.ts?

Yeah that’s a good idea 👍

koba04: Thank you! I've fixed it.

shuding (approved):
This fix is neat, thank you!

koba04:
Thank you for your help! 🎉

PR #1010, opened 2021-03-05
enable typescript strict mode

promer94:


jest.config.js
```diff
@@ -2,5 +2,10 @@ module.exports = {
   preset: 'ts-jest',
   testRegex: '/test/.*\\.test\\.tsx$',
   modulePathIgnorePatterns: ['<rootDir>/examples/'],
-  setupFilesAfterEnv: ['<rootDir>/jest-setup.ts']
+  setupFilesAfterEnv: ['<rootDir>/jest-setup.ts'],
+  globals: {
+    'ts-jest': {
+      diagnostics: false
```
huozhi, line 8: `diagnostics` is removed in #903, I think we want to keep it removed

promer94: find a batter way to disable strict mode in test 👌

koba04:
Great work! I prefer `@ts-expect-error` to `@ts-ignore`, but the version SWR uses doesn't support it, so it would be great if it is replaced after updating the version.

src/index.ts
```diff
@@ -11,7 +11,7 @@ export {
   revalidateType,
   RevalidateOptionInterface,
   keyInterface,
-  responseInterface,
+  ResponseInterface as responseInterface,
```
shuding, line 14: This is a breaking change, let’s do it in #1016  instead.

shuding (commented):
Generally looking great, nice work! 🎉

I might ship #1016 before this one, because the types are broken in the current beta release. But I think there won’t be too many conflicts.

src/index.ts
shuding: #1016 was merged so this PR is unblocked.

shuding (approved):
Thank you so much!  🎉 🎉

PR #1011, opened 2021-03-05
keep old esm directory structure

promer94:
#932 accidentally changed the esm directory structure.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -16,6 +16,7 @@
     "types": [
       "node", "jest"
     ],
+    "rootDir": "src",
     "target": "es5",
     "typeRoots": [
       "./types",
@@ -24,7 +25,7 @@
   },
   "include": [
     "src/**/*",
-    "./jest-setup.ts"
+    "../jest-setup.ts"
   ],
   "watchOptions": {
     // Use native file system events for files and directories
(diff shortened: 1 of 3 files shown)


promer94:
#932 accidentally changed the esm directory structure.

tsconfig.json
```diff
@@ -24,7 +25,7 @@
   },
   "include": [
     "src/**/*",
```
shuding, line 27: Should we also update this?

promer94: #TIL 
> Importantly, rootDir does not affect which files become part of the compilation. It has no interaction with the include, exclude, or files tsconfig.json settings.
https://www.typescriptlang.org/tsconfig#rootDir

shuding: That is confusing, so `includes` only determines which definitions get built?

promer94: rootDir is used to keep the output's directory structure simple and clean 
includes determines which file (must be in rootDir) will be compiled by tsc

We should have a separate tsconfig.test.json which is only used for ts-jest
In this way ```jest-setup.ts``` will not affect the output's directory structure anymore

shuding (approved):
Thanks! ❤️

PR #1016, opened 2021-03-06
Clean up types

shuding:
This PR does 2 things:
1. The exported type `ConfigInterface` was changed in #946, which might be breaking. **However it's not released as stable yet**. This PR fixes it.
2. Renames some types to be more consistent. Also marks the old ones as deprecated. 

For 1, the cause is that properties in `ConfigInterface` are no longer optional. And if a user creates a custom hook the type might be broken:

```ts
import useSWR, { ConfigInterface } from 'swr'

function useData (opts: ConfigInterface) {
  return useSWR('/api', opts)
}

useData({}) // ← error because all fields are required
```

For 2, updated types are:

| Current | New |
|:----|:---|
| `keyInterface` | `Key` |
| `ConfigInterface` | `SWRConfiguration` |
| `responseInterface` | `SWRResponse` |
| `SWRInfiniteConfigInterface` | `SWRInfiniteConfiguration` |
| `SWRInfiniteResponseInterface` | `SWRInfiniteResponse` |
| `revalidateType` | `Revalidator` |
| `RevalidateOptionInterface` | `RevalidatorOptions` |
| `CacheInterface` | `Cache` |

The deprecated ones will be kept until 1.0.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -38,7 +38,7 @@ export type Configuration<
     key: string,
     config: Configuration<Data, Error>,
     revalidate: revalidateType,
-    revalidateOpts: RevalidateOptionInterface
+    revalidateOpts: RevalidateOptions
   ) => void
   registerOnFocus?: (cb: () => void) => void
   registerOnReconnect?: (cb: () => void) => void
@@ -46,14 +46,9 @@ export type Configuration<
   compare: (a: Data | undefined, b: Data | undefined) => boolean
 }
 
-export interface RevalidateOptionInterface {
-  retryCount?: number
-  dedupe?: boolean
-}
-
 export type keyType = string | any[] | null
 type keyFunction = () => keyType
-export type keyInterface = keyFunction | keyType
+
 export type updaterInterface<Data = any, Error = any> = (
   shouldRevalidate?: boolean,
   data?: Data,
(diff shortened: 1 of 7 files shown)


src/types.ts
```diff
@@ -1,9 +1,12 @@
...
+
+export type Fetcher<Data> = (...args: any) => Data | Promise<Data>
+
+export type Configuration<
   Data = any,
   Error = any,
-  Fn extends fetcherFn<Data> = fetcherFn<Data>
-> {
+  Fn extends Fetcher<Data> = Fetcher<Data>
+> = {
```
huozhi, line 9: is this `=` by purpose?

src/types.ts
```diff
@@ -16,99 +19,198 @@ export interface ConfigInterface<
...
-  fetcher: Fn
   suspense: boolean
+  fetcher: Fn
   initialData?: Data
 
   isOnline: () => boolean
   isDocumentVisible: () => boolean
   isPaused: () => boolean
-  onLoadingSlow: (key: string, config: ConfigInterface<Data, Error>) => void
+  onLoadingSlow: (key: string, config: Configuration<Data, Error>) => void
```
huozhi, line 29: would you also want to add the `?` back to mark those properties as optional again in this PR?

src/types.ts
shuding: Yeah I changed it from interface to type

src/types.ts
shuding: This is the internal type which is more strict. The exported one has `Partial` so all properties are optional.

huozhi (approved):

huozhi (commented):
I just encounter an issue like specifying the `ConfigInterface` in another PR.  Could we consider exporting the paritial one?

```ts
export type ConfigInterface = Partial<{....}>
```

shuding:
It is already the partial one: `Partial<Configuration<Data, Error, Fn>>`

shuding:
Published as `swr@0.4.3-beta.3`.

ivan-kleshnin:
Maybe also rename `SWRConfig` to `SWRProvider`?

`SWRConfig (component) vs SWRConfiguration (type)` – way too similar names.

Issue #630, opened 2020-09-02
Add change description info to cache.subscribe


gaoxiaoliangz:
```js
import { cache } from 'swr'

const listener = (e) => {
  console.log(e) // undefined
  syncAllCacheToLocalStorage()
}

cache.subscribe(listener)
```

Currently nothing is passed to the listener function. I would like to impl some sort of cache sync in my app (sync between localStorage and swr cache). And I would like to know which cache key changed, so that I would not need to iterate through every key and update every cache to localStorage.

ryaninvents:
@gaoxiaoliangz Just as a heads up, looks like #365 will implement this when merged.

huozhi:
@gaoxiaoliangz you can use the following code to access the iterative value pairs in the swr cache, then you can sync/serialize it to local storage as a workaround

```js
const listener = () => {
  const cacheKvPairs = cache.keys().map(key => ([key, cache.get(key)]))
  syncToLocalStorage(cacheKvPairs)
}
```

but since the cache APIs are not documented officially, it means cache calls might have breaking changes in the future when you upgrade swr to higher version.

gaoxiaoliangz:
> @gaoxiaoliangz you can use the following code to access the iterative value pairs in the swr cache, then you can sync/serialize it to local storage as a workaround
> 
> ```js
> const listener = () => {
>   const cacheKvPairs = cache.keys().map(key => ([key, cache.get(key)]))
>   syncToLocalStorage(cacheKvPairs)
> }
> ```
> 
> but since the cache APIs are not documented officially, it means cache calls might have breaking changes in the future when you upgrade swr to higher version.

This is actually what I impl now, but the problem is that there is too much data to sync, and each time new data arrives I have to check them all in order to sync them to local, this is really inefficient.

huozhi:
You could use custom cache to build out this feature now. Checkout [cache#sync-cache-to-localstorage](https://swr.vercel.app/advanced/cache#sync-cache-to-localstorage) for details

Issue #16, opened 2019-10-29
Support multiple caching mechanisms


aequasi:
### Frontend based:
Session Storage
Local Storage
WebSQL
IndexedDb

### Backend Based:
Anything...?

---

Frontend based caches could help add support for persistence.
Backend based caches could help with adding support for SSR.

quietshu:
Thanks @aequasi !

We tested Local Storage and IndexedDb when started implementing this on the ZEIT dashboard. And we learned many things from it:
- Local Storage is IO heavy (it stores data on the disk), and cache read/write happens very frequently when the website starts to load. It will slow down the rendering and other resources.
- IndexedDb is async, which means the SWR cannot return the data from cache synchronously. So when navigating (back and forward) through the website with cache, React will always render without data first.
- and some other UX feedback about client side storage (e.g.: loading skeleton vs out of dated data from local cache when opening the website)

But it might be a good idea to support cache fallback (e.g.: Memory -> IndexedDb), expiration and other features. Anyway, supporting multiple caching mechanisms is definitely an important option. :+1:

netspencer:
I’d love to see a layered cache architecture. for speed and such, i think the memory cache should always be the top level. and then the actual network requests are the lowest level

the ability to then add intermediate cache levels would be very powerful (for my usecases at least)

in other words, the first cache that should be checked shall remain the in memory map, which is sync, and will ensure the above list works as expected. but then adding something like indexdb of websql as a layer before making a network request would be great

aequasi:
Fair points @quietshu !

Regarding IndexedDb, does it have to return synchronously?  Presumably, if you are wanting to use IndexedDb, you would have to accept the fact that its asynchronous, and you would have to have the 'non-data' view.

quietshu:
@aequasi 

> Regarding IndexedDb, does it have to return synchronously? 

It's okay to have async cache, and we can also read from memory cache because it's always a stream 👍that's why a layered cache architecture (as @netspencer mentioned too 🙏) can be very helpful.

derindutz:
In case it's helpful, this is what I'm doing:

```
import useZeitSWR from '@zeit/swr';
import localForage from 'localforage';

export async function useSWR(key, fn, swrOptions) {
  const result = await useZeitSWR(key, fn, swrOptions);
  if (result.data === undefined) {
    const localResult = localForage.getItem(key);
    if (localResult !== null) {
      return localResult;
    }
  }
  return result;
}
```

[...]

PR #1017, opened 2021-03-07
Custom cache provider

huozhi:
### Changes

This PR is a early phase implementation for #728 . Add `cache` property as a cache provider to SWR context

```js
import { createCache } from 'swr'

const customCache = new Map()
const { provider } = createCache(customCache)

<SWRConfig value={{ provider }}>
   <Section />
<SWRConfig>
```

This is the early support of #16 , user can have a way to access the cache state and changes ( #630 ), but won't effect swr internal cache. For example, you can use any store like localStorage / RN AsyncStorage, swr internal will still use a sync cache to guarantee the correct state updating sequence.

In the long term, we can consider to deprecate the exposed cache

### Example

```js
import { createCache } from 'swr'

const localStorageCache = {
   set(key, value) {
      window.localStorage.setItem(key, JSON.stringify(value))
   },
   delete(key) {
    window.localStorage.removeItem(key)
  },
}

const { provider } = createCache(localStorageCache)

return <SWRConfig value={{ provider }}>
```


bionicles:
I would use it! can you make examples with redux, cache api https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Client-side_web_APIs/Client-side_storage#the_future_cache_api, and Dexie.js for indexeddb?

bionicles:
Might it be better to use the Cache API? it persists the data to apps installed locally on mobile devices, and it works inside WebWorkers https://developer.mozilla.org/en-US/docs/Web/API/Cache

https://caniuse.com/?search=cache

also, what about customizing the "get" in addition to the "set" ?

shuding (commented):
Generally LGTM! 👍 

Since this is breaking and deprecates `Cache`, we need to first prepare a PR to add `@deprecates` to `Cache` for 0.x versions.

Secondly since we don't need to expose `Cache` anymore, I think we shouldn't wrap the provider with a class. We can just use a couple of lightweight functions:

```js
function get(provider, key) { return provider.get(key) }
function set(provider, key, value) { return provider.set(key, value) }
```

And we don't need those internal listeners too. This change will be useful when we pass the provider down via context, and we don't maintain the `cache` singleton itself inside SWR.

huozhi:
5834b59  latest change:

limit cache API usage inside swr, only using set/get/delete, and they're all sync APIs.

for global `mutate` and `trigger`, use `const cache = globalCache` to consume global `cache` directly;
for `useSWR` hook, it will consume provider from context or fallback to global `cache`.

shuding:
I refactored the API a little bit:

```js
const { cache, mutate } = createCache(provider)
```

But the implementation is still not ideal:
- The interface of `provider`. Should we do serialization inside or assume it's always a string?
- The internal code structure. `use-swr` depends on `use-args`, `use-args` depends on the default config and cache, but the default cache depends on `use-swr`. So I had to choose a strange way to implement this (`defaultCache = getCacheFactory(() => {})(new Map()).cache`).

I need to spend some more time thinking about it. I also moved web presets out of the default config object, because these can not be options. But I'm not very sure about that decision, let me know what you think!

shuding:
While the main goal of this PR is to better cover testing scenarios, there're still a couple of to-dos left (no blocking this PR to be merged, just writing them down):
- [ ] add a test case for multi-level cache + mutation
- [ ] add a test case for multi-cache isolation (2 components with different cache provider)

src/types.ts
```diff
@@ -19,9 +19,8 @@ export interface Configuration<
   suspense: boolean
   fetcher: Fn
   initialData?: Data
+  cache: Cache
```
huozhi, line 22: shall we still keep the presets here to avoid less changes? it's easier for other js-runtime to customize swr

huozhi (commented):
genterally LGTM! you might also want to fix those lint warnings.

src/use-swr.ts
```diff
@@ -678,9 +678,25 @@ function useSWR<Data = any, Error = any>(
...
+): {
+  cache: Cache
+  mutate: ScopedMutator<Data>
+} {
+  const cache = wrapCache<Data>(provider)
+  return {
+    cache,
+    mutate: internalMutate.bind(null, cache) as ScopedMutator<Data>
+  }
+}
```
huozhi, line 700: is deprecating `trigger` also be included in this PR's scope?

shuding: Yeah I think we have to, because it should be scoped to the cache just like `mutate`, so it's better to just deprecate it.

src/use-swr.ts
```diff
@@ -645,7 +641,7 @@ function useSWR<Data = any, Error = any>(
     (newData, shouldRevalidate) => {
       return internalMutate(cache, keyRef.current, newData, shouldRevalidate)
     },
-    []
+    [cache]
```
shuding, line 644: Do you think if we should add ESLint ignore, since we’re sure that cache is not gonna change?

shuding (approved):
🎉🎉🎉

PR #1018, opened 2021-03-07
Micro optimizations

shuding:
Make the code cleaner and more consistent:
- use the `useState({})[1]` pattern for re-rendering the component
- make the internal property names shorter in `Cache`
- make the key prefixes in useSWRInfinite 3 chars (`ctx@`, `inf@`, `len@`), and internal cache property name shorter
- make `addRevalidator` return the ubsubscriber

Closes #896.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -1,7 +1,7 @@
 import { Cache as CacheType, Key, CacheListener } from './types'
 import hash from './libs/hash'
 
-export default class Cache implements CacheInterface {
+export default class Cache implements CacheType {
   private cache: Map<string, any>
   private subs: CacheListener[]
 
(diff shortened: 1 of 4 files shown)


huozhi (approved):
generally lgtm. the lint warning is a fake one. disabling it or rollback to array destruction are both acceptable.
eslint hooks ignore the none array pattern for useState, haven't found any useful case to work around yet

shuding:
@huozhi It seems like the linter didn't yell, so I think it's fine :)

ranisalt:
@shuding care to point out what is the `useState({})[1]` trick? I'm curious on what it changes.

shuding:
@ranisalt it's an implementation detail inside SWR, previously we are using `const [,rerender] = useState(false)` to create a re-render function, now it uses `const rerender = useState({})[1]`.

PR #1024, opened 2021-03-10
Make swr zero dependency

huozhi:
make swr core dependent less and light weight. the deep equal won't change much, now make as a vendor of swr lib.

* extract lite version of deep equal as lib
* remove dependency

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,4 +1,4 @@
-import deeEqual from './libs/deep-equal'
+import deepEqual from './libs/deep-equal'
 import { Configuration, RevalidatorOptions, Revalidator } from './types'
 import Cache from './cache'
 import webPreset from './libs/web-preset'
@@ -64,7 +64,7 @@ const defaultConfig = {
   refreshWhenOffline: false,
   shouldRetryOnError: true,
   suspense: false,
-  compare: deeEqual,
+  compare: deepEqual,
 
   isPaused: () => false,
   ...webPreset
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,4 +1,4 @@
-import { dequal } from 'dequal/lite'
+import deeEqual from './libs/deep-equal'
 import { Configuration, RevalidatorOptions, Revalidator } from './types'
 import Cache from './cache'
 import webPreset from './libs/web-preset'
@@ -64,7 +64,7 @@ const defaultConfig = {
   refreshWhenOffline: false,
   shouldRetryOnError: true,
   suspense: false,
-  compare: dequal,
+  compare: deeEqual,
 
   isPaused: () => false,
   ...webPreset
--- a/src/libs/deep-equal.ts
+++ b/src/libs/deep-equal.ts
@@ -1,0 +1,31 @@
+const has = Object.prototype.hasOwnProperty
+
+export default function deeEqual(foo: any, bar: any) {
+  let ctor, len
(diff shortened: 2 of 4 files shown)


src/config.ts
```diff
@@ -1,4 +1,4 @@
-import { dequal } from 'dequal/lite'
+import deeEqual from './libs/deep-equal'
```
shuding, line 1:
```suggestion
import deepEqual from './libs/deep-equal'
```

src/config.ts
```diff
@@ -64,7 +64,7 @@ const defaultConfig = {
   refreshWhenOffline: false,
   shouldRetryOnError: true,
   suspense: false,
-  compare: dequal,
+  compare: deeEqual,
```
shuding, line 67:
```suggestion
  compare: deepEqual,
```

src/libs/deep-equal.ts
```diff
@@ -0,0 +1,31 @@
+const has = Object.prototype.hasOwnProperty
```
shuding, line 1: Let's include https://github.com/lukeed/dequal/blob/master/license as a comment at the top of this file.

shuding (changes requested):

shuding (approved):

lukeed:
This does more harm than good. 

Because swr is a library, and _not_ a built distribution (like a CLI), literally all this does is make the end-users' applications larger because any other deps that relied on `dequal` (or any shared dep) cannot do that anymore.

Dequal itself has no dependencies & has been extremely stable, so you had no motivation to do this.

This was done purely as a "dependency free" marketing tactic which is total bullshit & breeds distrust in library authors, like myself, who _actually_ make dependency-free packages.

lukeed:
While this stands, as per MIT licensing, you have to include my license within your `LICENSE` as it's now part of your package. This is theft otherwise.

src/libs/deep-equal.ts
developit: The license must be included in the repo root and/or readme such that it remains visible in the npm package, as npm packages are copies ([substantial portions](https://github.com/lukeed/dequal/blob/9aa73181ac7e081cd330cac67d313632ac04bb02/license#L13)) of software.

huozhi:
@lukeed My apologies to the inappropriate way of using `dequal` here, and the misleading PR title which is not intend to marketing swr as a dependency free library. The PR purpose should be described as use dequal lite as inline module, that's it nothing else.

I'm super appreciated all the work you've done and the awesome libraries you built, feel sorry to bring such sore feeling to you because of my bad description which is definitely not anyone's purpose to harm origin author. I'll file another PR to include the license of `dequal` into swr, so much thanks for the reminding.

lukeed:
The PR title is "Make swr zero dependency" – I'm not sure how I've been misled here.

> I'll file another PR to include the license of dequal into swr, so much thanks for the reminding

Thank you for agreeing to update the LICENSE – I'll wait for that PR – but the rest of my points still stand: inlining `dequal` does more harm than good because `swr`'s copy **cannot be shared** by/with other packages inside an application. It should be left as a dependency – because it IS a dependency.

Thank you for appreciating my work, but this does not align with those words. At best, this is _hiding_ your appreciation. And this is Vercel shipping my code as their own, which it is not: https://www.diffchecker.com/UbAYQ6Ht

lukeed:
Thank you

PR #1025, opened 2021-03-10
Adopt bunchee as bundler

huozhi:
Use [bunchee](https://github.com/huozhi/bunchee) as bundler for cjs/esm output

* replace ncc
* minify all assets
* better watch mode for all assets

reasons that didn't use common bundlers:

* microbundle - doesn't support esm interop on cjs output
* esbuild - doesn't support es5 output, might have browser compatible issues

in long term, would still love to adopt the trending bundler from community


package.json
```diff
@@ -3,22 +3,19 @@
...
   "repository": "vercel/swr",
   "homepage": "https://swr.vercel.app",
   "license": "MIT",
   "scripts": {
-    "build": "npm run build:esm && npm run build:cjs",
-    "build:cjs": "ncc build src/index.ts -o dist -m -e react",
-    "build:esm": "tsc --module ES6 --outDir esm",
-    "watch": "tsc --watch",
-    "watch:esm": "tsc --module ES6 --outDir esm --target esnext --watch",
+    "build": "bunchee src/index.ts --no-sourcemap -m",
```
huozhi, line 16: Unsure thing about whether to keep sourcemap here. Might increase the debug difficulty on production with minified code only for users.

src/use-swr.ts
```diff
@@ -584,10 +584,7 @@ function useSWR<Data = any, Error = any>(
     const softRevalidate = () => revalidate({ dedupe: true })
 
     // trigger a revalidation
-    if (
-      isUpdating ||
-      willRevalidateOnMount()
-    ) {
+    if (isUpdating || willRevalidateOnMount()) {
```
huozhi, line 587: missing lines for prettier

package.json
shuding: I think for now let's include it as it helps identifying the original code, but I doubt if anyone is using it.

shuding:
CJS build is now 32% smaller, great job! 🎉

shuding:
Published as `swr@1.0.0-beta.2` 🎉

PR #1035, opened 2021-03-14
refactor: fix a react-hooks/exhaustive-deps warning

koba04:
An ESlint warning of `react-hooks/exhaustive-deps` is caused by https://github.com/vercel/swr/commit/b18414b901d60fbe63232537f298dc847ae20822 because the rule cannot determine `rerender` is always the same reference from `rerender = useState({})[1]`. Previously, this line was `[, rerender] = useState({})` and the rule can determine it is always the same reference.

To fix this, I've added `rerender` into the deps array because I'd like to enable the rule for the `setSize` function.

### DIFF
--- a/src/use-swr-infinite.ts
+++ b/src/use-swr-infinite.ts
@@ -211,7 +211,9 @@ function useSWRInfinite<Data = any, Error = any>(
       rerender({})
       return mutate(v => v)
     },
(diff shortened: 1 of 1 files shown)


shuding (approved):
Thanks!

koba04:
Thank you!

Issue #1050, opened 2021-03-18
Illegal Invocation


tehpsalmist:
# Bug report

## Description / Observed Behavior

The code below is producing a `TypeError: Illegal Invocation` when the same key is passed to useSWR.

## Expected Behavior

Not to throw an illegal invocation error.

## Repro Steps / Code Example

 Run the code, then simulate by typing in some letters and backspacing.

```js
import React, { useState } from 'react'
import { render } from 'react-dom'
import useSWR from 'swr'

export const App = props => {
  const [search, setSearch] = useState('')

  const { data: src, error } = useSWR(search, (key) => {
    return fetch(`http://www.omdbapi.com/?apikey=276414e9&&t=${key}`)
      .then(res => res.ok ? res.json() : res.text())
      .then(json => json.Poster)
  })

  return <main>
    <input type='text' value={search} onChange={e => setSearch(e.target.value || '')} />
    <button type='submit' >submit</button>
    <img src={src || ''} alt='movie poster' />
  </main>
}

render(<App />, document.getElementById('app'))
```

## Additional Context

SWR version: 0.5.3
React/ReactDOM: 17.0.1
Can't reproduce in CodeSandbox
Using Parcel2 and React Fast Refresh to serve app.
Example code is literally all of the code in the project (except package.json and minimal index.html)

koba04:
@tehpsalmist Could you create a GitHub repository for reproducing this?

tehpsalmist:
Most definitely, I should have done that first. Uno momento.

tehpsalmist:
Reproduce with this repo: https://github.com/tehpsalmist/illegal-invocation-bug

tehpsalmist:
Noice. You can venmo me the bug bounty. :trollface:

PR #1051, opened 2021-03-19
fix: invalid invocation on raf

### LINKED ISSUE #1046: Missing requestAnimationFrame after upgrading to 0.5.3
# Bug report

After upgrading from 0.4.2 to 0.5.3, I now get the following error on a specific call of `useSWR`:

`TypeError: 'requestAnimationFrame' called on an object that does not implement interface Window.`

## Expected Behavior

It should load as it did before

## Repro Steps / Code Example

I don't know how to isolate this. Other calls work fine, this works fine too, it breaks after a rerender. Stack trace is useless, it just points that it happens inside `useSWR`, and none of my calls pass `requestAnimationFrame` calls inside useSWR.

## Additional Context

SWR version: 0.5.3

huozhi:
Fixes #1046 #1050

### Minimal repro

```js
a = {
   raf: window.requestAnimationFrame,
   raf_pure: f => window.requestAnimationFrame(f)
}

f = () => console.log(1)
a.raf(f) // Error
// Chrome: Uncaught TypeError: Illegal invocation
// Firefox: Uncaught TypeError: 'requestAnimationFrame' called on an object that does not implement interface Window.

a.raf_pure(f) // log `1` in console
```

Possible root cause:

checked the repro in #1050, the line calling rAF is compiled to `c.rAF(n)`, rAF called on an object `c`, which represents a module. We need to make it pure.

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -8,7 +8,9 @@ export const IS_SERVER =
 // polyfill for requestAnimationFrame
 export const rAF = IS_SERVER
   ? null
-  : window['requestAnimationFrame'] || (f => setTimeout(f, 1))
+  : window['requestAnimationFrame']
+  ? (f: FrameRequestCallback) => window['requestAnimationFrame'](f)
+  : (f: (...args: any[]) => void) => setTimeout(f, 1)
 
 // React currently throws a warning when using useLayoutEffect on the server.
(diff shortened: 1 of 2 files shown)


shuding (approved):
Nice find

koba04:
That's so great! Thank you for your fix!

PR #1054, opened 2021-03-19
types: expose KeyLoader

### LINKED ISSUE #707: Export KeyLoader and keyType typescript types
First off, thanks so much for this library. It's the best.

# Bug report

I'd like to use `KeyLoader` (from `./use-swr-infinite`) and `keyType` (from `./types`)

There are some types that aren't exposed.

## Expected Behavior

```ts
import type { KeyLoader, keyType } from 'swr'
```

I'd like these types to be exported so that I can wrap `useSWRInfinite` with a custom function. This is the case with `SWRInfiniteConfigInterface`, for example.

## Repro Steps / Code Example

I'm not sure if this needs a repro.

## Additional Context

SWR version: `0.3.4`

huozhi:
Resolves #707

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,6 +17,7 @@ export {
   Revalidator,
   RevalidatorOptions,
   Key,
+  KeyLoader,
   SWRResponse,
   Cache,
   // Legacy, for backwards compatibility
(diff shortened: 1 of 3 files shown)


shuding (approved):
Thank you!

PR #1057, opened 2021-03-20
test: refactor use-swr-loading.test.tsx

koba04:
This is a follow-up PR of #933.
I've refactored use-swr-loading.test.tsx


shuding (approved):
Thanks!

PR #1058, opened 2021-03-21
fix #1056

### LINKED ISSUE #1056: Concurrent requests (suspense) with same response data will remain pending and causes an infinite request loop
# Bug report

Thank you for a great library!

I am rendering a list of components each using swr internally in suspense mode. Each component fetches data with different urls and the response data may result in structurally equal response data (which seems to be the key issue when fetching concurrently).
The initial load will be successful for all requests, but triggering a subsequent request for all components will cause requests resulting in same response data to remain pending and have the caller components remain in a suspended state. Furthermore, an infinite request loop has begun in the background calling the fetcher function every ~2 seconds.

This seems to only be an issue with version 0.5.x. Version 0.4.2 does not exhibit this behavior.

## Expected Behavior

To resolve data successfully

## Repro Steps / Code Example

https://codesandbox.io/s/swr5-concurrent-suspense-bug-onut5?file=/src/App.tsx

## Additional Context

SWR version 0.5.x

promer94:
https://codesandbox.io/s/swr5-concurrent-suspense-bug-onut5?file=/src/App.tsx

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -470,9 +470,11 @@ function useSWR<Data = any, Error = any>(
           // deep compare to avoid extra re-render
           // data changed
           newState.data = newData
-          cache.set(key, newData)
         }
 
+        if (!config.compare(cache.get(key), newData)) {
+          cache.set(key, newData)
+        }
         // merge the new state
         dispatch(newState)
 
--- a/test/use-swr-suspense.test.tsx
+++ b/test/use-swr-suspense.test.tsx
@@ -1,7 +1,7 @@
-import { act, render, screen } from '@testing-library/react'
+import { act, fireEvent, render, screen } from '@testing-library/react'
 import React, { ReactNode, Suspense, useEffect, useState } from 'react'
 import useSWR, { mutate } from '../src'
-import { sleep } from './utils'
+import { createResponse, sleep } from './utils'
 
 class ErrorBoundary extends React.Component<{ fallback: ReactNode }> {
   state = { hasError: false }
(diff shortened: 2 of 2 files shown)


shuding:
Can you also add https://codesandbox.io/s/swr5-concurrent-suspense-bug-onut5?file=/src/App.tsx as a test case? Thanks!

shuding (approved):
Thank you!

damusnet:
@shuding @promer94 @pacocoursey Hi, this issue has been fixed 3 weeks ago, yet, as far as I can tell, it still has not been published. Apparently there was briefly a v0.5.5 which was then pulled. What's the hold up? Cheers

shuding:
Hi @damusnet we're aiming for the 1.0 version so new changes will be released with the "beta" tag, please try install `swr@beta` and this PR should be included.

damusnet:
Thanks @shuding for the quick response.

Is there any chance this one bug fix could make it to the v0 branch? Especially if the v1 branch is going to have breaking changes, or is further out, it would be very useful for us users to have a faster way to fix this issue.

damusnet:
edit: I see now that v0.5.5 is still available on [npm](https://www.npmjs.com/package/swr), along with multiple v1 betas. The issue in my opinion is that in the absence of tagged releases with release notes here on GitHub, this is hard to follow for people like me who are not intimately familiar with the development of the project. I can't for instance just install `swr@beta` v1 without knowing if there are any breaking changes.

What happened with v0.5.5? Why does it not have a tag and release notes here on GitHub like previous releases?

shuding:
@damusnet 0.5.5 was a hot-fix for some license related issue, and it didn't go through the normal release flow. I see your point but it's harder for the maintainers to keep tracking both versions, as patches need to be applied to both branch when developing, and there will be huge amount of conflicts when merging back potentially. I can do a patch to the 0.x version for you (let me know if you need it urgently, I can do a release tomorrow), but still encourage you to try out 1.0 (there's no breaking change in the current 1.0 beta) because new bug fixes won't be applied to 0.x eventually when 1.0 is out.

damusnet:
Thank you for being so understanding and accommodating. I (and hopefully others) would really appreciate if you could cut a v0.x release with the usual release notes. All I need is some traceability (version points to tag points to commit hash, etc.).

damusnet:
Hi @shuding 

Sorry to insist, but as far as I can tell, it's been another three weeks since we last talked, and there still doesn't seem to be any non-beta release, v1.x or v0.x, that addresses the original bug fixed in this thread.

Can you please publish a v0.5.6?

shuding:
@damusnet I just published 0.5.6.

The tricky part for us is, we have to cherry-pick this fix to the v0.5.5 commit, while we are already working on the new version for a while. We don't have 2 branches to maintain both stable and beta and make sure every fix is applied to both (and we don't have the time). In the future, we will encourage people to upgrade to the latest beta to access new features and bug fixes earlier. Otherwise they have to wait until the stable release. Thanks for your understanding!

damusnet:
@shuding thanks a lot, really appreciate the new release, and totally understand the concerns!

PR #1069, opened 2021-03-25
Allow setting the generic on mutate()

nstepien:
Without this change, it is not possible for users to do `mutate<MyType>(...)`

image


shuding (approved):
Thank you!

PR #1075, opened 2021-03-27
Fix mutate

### LINKED ISSUE #1070: Bound mutate based on current data inconsistently update data
# Bug report

When using [mutate based on current data](https://swr.vercel.app/docs/mutation#mutate-based-on-current-data) in a [bounded fashion](https://swr.vercel.app/docs/mutation#bound-mutate), 
and the data contains an array, 
sometimes the data is _not_ updated.

## Expected Behavior

The data is updated consistently when a bound mutate based on current data is used.

## Repro Steps / Code Example

https://codesandbox.io/s/useswr-async-mutate-bug-y77z1?file=/src/App.js

1. Click on the "X" button to attempt to delete a line
2. The line is supposed to disappear after 2 seconds, observe sometimes it fails. (somehow it tends to fail more easily if start to delete from the middle of the list). You can verify the mutation indeed takes place by checking the console logs, it should have a "delete fired" log every time the "X" button is clicked.

## Additional Context

swr: 0.5.4

### LINKED ISSUE #1074: Cache cant be changed if mutating happened before SWR fetching
# Bug report

when mutating with async function, if async function resolved after the fetcher, the swr cache can't be changed by mutate or fetcher

## Expected Behavior

when mutating with async function, the fetcher result should be ignored.

## Repro Steps / Code Example
https://codesandbox.io/s/swr-mutate-bug-eujb4?file=/src/index.tsx

## Additional Context

SWR version. 0.5.4

Related to #1070

promer94:
cache cant update
https://codesandbox.io/s/swr-mutate-bug-eujb4?file=/src/index.tsx

inflight request should be ignored
https://codesandbox.io/s/swr-mutate-bug2-39199?file=/src/index.tsx

mutate should update data
https://codesandbox.io/s/useswr-async-mutate-bug-y77z1?file=/src/App.js

close #1070; close #1074;

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -112,7 +112,6 @@ async function mutate<Data = any>(
 
   // track timestamps before await asynchronously
   const beforeMutationTs = MUTATION_TS[key]
-  const beforeConcurrentPromisesTs = CONCURRENT_PROMISES_TS[key]
 
   let data: any, error: unknown
   let isAsyncMutation = false
@@ -142,10 +141,7 @@ async function mutate<Data = any>(
 
   const shouldAbort = (): boolean | void => {
     // check if other mutations have occurred since we've started this mutation
-    if (
-      beforeMutationTs !== MUTATION_TS[key] ||
-      beforeConcurrentPromisesTs !== CONCURRENT_PROMISES_TS[key]
-    ) {
+    if (beforeMutationTs !== MUTATION_TS[key]) {
       if (error) throw error
       return true
     }
@@ -389,7 +385,7 @@ function useSWR<Data = any, Error = any>(
         // we have to ignore the revalidation result (res) because it's no longer fresh.
         // meanwhile, a new revalidation should be triggered when the mutation ends.
         if (
-          MUTATION_TS[key] &&
+          MUTATION_TS[key] !== undefined &&
           // case 1
           (startAt <= MUTATION_TS[key] ||
             // case 2
--- a/test/use-swr-local-mutation.test.tsx
+++ b/test/use-swr-local-mutation.test.tsx
@@ -506,4 +506,141 @@ describe('useSWR - local mutation', () => {
     expect(mutationRecivedValues).toEqual([0, 1])
     expect(renderRecivedValues).toEqual([undefined, 0, 1, 2])
   })
+
+  it('async mutation case 1 (startAt <= MUTATION_TS[key])', async () => {
+    let result = 0
+    const fetcher = jest.fn(createResponse)
+    function Component() {
+      const { data, mutate: boundMutate } = useSWR(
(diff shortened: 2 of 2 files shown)


src/use-swr.ts
```diff
@@ -442,7 +452,8 @@ function useSWR<Data = any, Error = any>(
         // we have to ignore the revalidation result (res) because it's no longer fresh.
         // meanwhile, a new revalidation should be triggered when the mutation ends.
         if (
-          MUTATION_TS[key] &&
+          // Since ts start at 0
+          MUTATION_TS[key] !== undefined &&
```
huozhi, line 456: I think ts starts from 1, now returns `++ts`

promer94: https://github.com/vercel/swr/blob/eb7fa36920fa5eaca5d68a61ece6eabe54c64907/src/use-swr.ts#L110 😂

huozhi: how about bump init ts from 0 to 1? then the `now() - 1` will start from 1. I guess it's safe to have it since ts is for comparsion?

promer94: I personally feels that ```MUTATION_TS[key] !== undefined``` might be batter here. 
Because in here we want to check the existence of the mutation timestamp. 
But I don't have strong opinion here.

shuding: Another problem here is that once `MUTATION_TS` is assigned, it will never get deleted (or reset to 0/undefined). So I think that 2 `mutate()` calls will fail here?

jameslouiz:
For "cache cant update", this works: https://codesandbox.io/s/swr-mutate-bug-forked-95od2?file=/src/index.tsx. Simply conditionally return a fetcher.

promer94:
> For "cache cant update", this works: https://codesandbox.io/s/swr-mutate-bug-forked-95od2?file=/src/index.tsx. Simply conditionally return a fetcher.

if you check the console.log, you will find fetcher never called. 
SWR actually never updates the fetcher if the key stay same. In this example, fetcher in SWR is always null

shuding (approved):

shuding:
Thank you!

jameslouiz:
> > For "cache cant update", this works: https://codesandbox.io/s/swr-mutate-bug-forked-95od2?file=/src/index.tsx. Simply conditionally return a fetcher.
> 
> if you check the console.log, you will find fetcher never called.
> SWR actually never updates the fetcher if the key stay same. In this example, fetcher in SWR is always null

Ah, yes. You're totally right.

This fix would help with a similar issue I have

jameslouiz:
I managed to get it working by using `isPaused`: https://codesandbox.io/s/swr-mutate-bug-forked-95od2?file=/src/index.tsx

PR #1077, opened 2021-03-28
Code refactoring

shuding:
This PR moves the argument resolution logic into a standalone function. It will empower future high-level hooks that are built on top of `useSWR` such as `usePassive` (name TBD.), and it has already simplified the current `useSWR` and `useSWRInfinite` hooks logic. Also to unify the behavior of `useSWRInfinite`, it is now allowed to pass `null` as `fetcher` too.


huozhi (approved):
Super clean!

PR #1088, opened 2021-04-02
fix: race problem when in dedupingInterval

MoonBall:
before:
swr-deduping

after:
fix the above problem by utilizing `CONCURRENT_PROMISES_TS[key]` is undefined after executing `delete CONCURRENT_PROMISES_TS[key]`.

see added test case for detail.


shuding (approved):
Good catch, thank you!

Issue #936, opened 2021-01-28
Support conditional fetching in useSWRInfinite


em:
I've tried passing null instead of a getKey function to useSWRInfinite. Also tried returning null from getKey.

In both cases, this results in a deadlock where no additional pages will be loaded when setSize is used to increase the size.

wildfrontend:
@em maybe you can try 
`const loadMore = useCallback(() => setSize(state=> state +1),[setSize])`

but another case is when fetch request error, useSWRInfinite will also return infinite loop

em:
Thanks wildfrontend. I was able to workaround before I filed the issue by just restructuring things with wrapper components. Conditional fetching in useSWR is really just sugar anyway and the same thing can be accomplished by conditional rendering, it can just make your component IA not very nice. I filed the issue because it's a nice-to-have feature request for useSWRInfinite to have parity with useSWR.

As an aside - there are a lot of caveats with useSWRInfinite, some are addressed in the original PR, but not documented in the official docs. Would be really nice for those to be big disclaimers at the top.

shuding:
FYI you can pass `null` as the first argument to `useSWRInfinite` and it will pause data fetching. Can you provide a reproduction?

dev-smart-ui:
Hi I have same problem , if chat id undefined on first render , useSWRInfinite send requests on second render receive data from API , but const {data }=undefined  even if i se data in network tab google chrome 
```
  const { data, error, isLoading, isValidating, mutate:mutateCurrentChat, size, setSize } =
    useSWRInfinite<IConversation>(
      (pageIndex, previousPageData) => { 
        if (previousPageData && !previousPageData.length) return null; // reached the end
        if (!chatId) return null;
        return `${API.singleChat(chatId)}?page=${pageIndex + 1}&page_size=20`;
      },
      fetcher,
      {
        focusThrottleInterval: 8000,
        refreshInterval: 150000,
      }
    );
```

PR #1096, opened 2021-04-04
Bug fixes

### LINKED ISSUE #637: data returning { force: true }
# Bug report

## Description / Observed Behaviour
Sometimes the `data` that is returned from the useSWR hook is a `{ force: true }`
This seems to be caused by the cache value being set to `{ force: true }` to trigger a re-validate? It's only happened since v0.3.x

## Expected Behaviour

`data` should either return undefined or actual returned data from fetcher

### LINKED ISSUE #774: useSWRInfinite leaks { originalData: undefined, force: false } objects from cache when setSize is called with null first page key
# Bug report

useSWRInfinite returns data as { originalData: undefined, force: false } when setSize is called with null first page key. 

## Expected Behavior

data is undefined when first key is null, regardless of calls to setSize

## Repro Steps / Code Example

WIth null first page key, call setSize.

## Additional Context

SWR version 0.3.8

We have page keys that depend on other data to load, so the first page key is while the component mounts and fetches data for the first time. Meanwhile, we call setSize based on another variable that is immediately available. Delaying our calls to setSize
until the page keys are non-null resolves this issue, but leaking the cache objects is a bug that can be resolved anyway.

shuding:
- Fix type: allow `getKey` to be `null`, relates to #936.
- Fix: changing size of `useSWRInfinite` with `null` key leaks internal state, fixes #637, fixes #774.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -129,10 +129,9 @@ export interface SWRResponse<Data, Error> {
   isValidating: boolean
 }
 
-export type KeyLoader<Data = any> = (
-  index: number,
-  previousPageData: Data | null
-) => ValueKey
+export type KeyLoader<Data = any> =
+  | ((index: number, previousPageData: Data | null) => ValueKey)
+  | null
 
 /**
  * @deprecated `SWRInfiniteConfigInterface` will be renamed to `SWRInfiniteConfiguration`.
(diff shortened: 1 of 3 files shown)


huozhi (approved):
lgtm. would be nicer if can give `mutate` and `setSize` return type later

PR #1102, opened 2021-04-05
Optimistic UI examples comments clearer

mkrtchian:
From my understanding the two examples of optimistic UI have the same behavior: they update optimistically, and then send a POST request, and update again according to the backend reply, if it replies.

The comments in the example `index.js` seem misleading (at least it was for me):

For optimistic UI we have [this comment](https://github.com/vercel/swr/blob/master/examples/optimistic-ui/pages/index.js#L13):
```javascript
// the fetch below could fail, in that case the UI will
// be in an incorrect state
```

For optimistic UI with Immer.js we have nothing about possible fetch failure, but we have [this comment](https://github.com/vercel/swr/blob/master/examples/optimistic-ui-immer/pages/index.js#L21):
```javascript
// if this fail it will rollback the optimistic update
```

I think we should add that the request may fail and lead to an incorrect state also for the example with Immer.js. The way it is now makes me think that with the Immer.js example we solve that problem.

### DIFF
--- a/examples/optimistic-ui-immer/pages/index.js
+++ b/examples/optimistic-ui-immer/pages/index.js
@@ -10,15 +10,19 @@ export default () => {
 
   async function handleSubmit(event) {
     event.preventDefault()
-    // call mutate to optimistically update the UI
-    // we use Immer produce to allow us to perform and immutable change
-    // while coding it as a normal mutation of the same object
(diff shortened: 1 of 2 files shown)


shuding (approved):
Thank you for making them better!

PR #1128, opened 2021-04-18
Replace deprecated types with latest types in axios-typescript example

eunsukimme:
# Problem
My VSCode shows warnings that `responseInterface`, `ConfigInterface` are deprecated which are used in axios-typescript example. I noticed that they have `@deprecated` annotations in type definition file and were described they will be renamed to `SWRResponse`, `SWRConfiguration` each other. 

Screen Shot 2021-04-18 at 2 00 36 PM

# Solution
So I replaced them with latest types(`SWRResponse`, `SWRConfiguration`) already existing.


huozhi (approved):
Thank you!

Issue #828, opened 2020-12-17
Clear dedupingInterval of an unmounted key


lorenzomigliorero:
# Bug report

## Description

I am developing a simple _CRUD_ application with two views:

_App.jsx_
```jsx
<Route path="/" component={User} />
<Route path="/edit" component={Edit} />
```
_User.jsx_
```jsx
const { data } = useSWR('/api/user', fetcher, { dedupingTime: 60000 });

return (
  <h1>{data.name}</h1>
)
```

_Edit.jsx_
```jsx
const editName = (name) => {
  fetch('/api/user/edit', { method: 'post', name })
  mutate('/api/user') // here User.jsx is unmounted
}

return (
  <button onClick={editName}>EDIT NAME</button>
)
```

If the user follows these steps:

- Navigate to '/'
- Navigate to '/edit'
- Click on edit name button
- Come back to '/' after 1 second

_/api/user_  data contains an old cached value.

I've tried to manually delete the cache value instead of calling _mutate_, but the issue remains.
_CONCURRENT_PROMISES_ remains fullfilled with the previous payload.

```js
const editName = (name) => {
  fetch('/api/user/edit', { method: 'post', name })
  cache.delete('/api/user')
}
```

## Expected Behavior

After calling _mutate_ on _Edit.jsx,_ I expect that a revalidation occurs when _User.jsx_ is mounted, ignoring _dedupingInterval_ value.
In other words, dedupingInterval should be cleared if mutate is called on an unmounted key.

smddzcy:
Any workaround for this? It's really important to be able to cache some requests with long `dedupingInterval`s, but also mutate them when required (e.g. after the user updates some setting on another page).

MoonBall:
@lorenzomigliorero @smddzcy I made a demo https://codesandbox.io/s/relaxed-hooks-chjyi?file=/src/App.js.
I didn't reproduce this issue. PTAL

smddzcy:
Because the prop is not `dedupingTime`, it's `dedupingInterval`. See: https://codesandbox.io/s/summer-waterfall-7j93r

MoonBall:
I would like to fix it ASAP

promer94:
This problem is a duplicate of #751 

You could update the cache data in this way
```js
mutate('/api/user', fetch('/api/user/edit', { method: 'post', name })) 
```

smddzcy:
[...]

PR #1139, opened 2021-04-25
fix: not clear promise of revalidate after mutating

MoonBall:
Fix https://github.com/vercel/swr/issues/828.

See added test case for detail.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -64,7 +64,7 @@ const trigger: Trigger = (_key, shouldRevalidate = true) => {
 
   const updaters = CACHE_REVALIDATORS[key]
 
-  if (key && updaters) {
+  if (key && updaters && updaters.length) {
     const currentData = cache.get(key)
     const currentError = cache.get(keyErr)
     const currentIsValidating = cache.get(keyValidating)
@@ -82,6 +82,11 @@ const trigger: Trigger = (_key, shouldRevalidate = true) => {
     }
     // return new updated value
     return Promise.all(promises).then(() => cache.get(key))
+  } else if (shouldRevalidate) {
+    // `updaters` would trigger revalidate if `updaters.length > 0`.
+    // So only clear promise of revalidate if `!updaters.length && shouldRevalidate`
+    delete CONCURRENT_PROMISES[key]
+    delete CONCURRENT_PROMISES_TS[key]
   }
   return Promise.resolve(cache.get(key))
 }
@@ -182,7 +187,7 @@ async function mutate<Data = any>(
     })
   } else if (shouldRevalidate) {
     // `updaters` would trigger revalidate if `updaters.length > 0`.
-    // So clear promise of revalidate if `!updaters.length && shouldRevalidate`
+    // So only clear promise of revalidate if `!updaters.length && shouldRevalidate`
     delete CONCURRENT_PROMISES[key]
     delete CONCURRENT_PROMISES_TS[key]
   }
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -168,7 +168,7 @@ async function mutate<Data = any>(
   // enter the revalidation stage
   // update existing SWR Hooks' state
   const updaters = CACHE_REVALIDATORS[key]
-  if (updaters) {
+  if (updaters && updaters.length) {
     const promises = []
     for (let i = 0; i < updaters.length; ++i) {
       promises.push(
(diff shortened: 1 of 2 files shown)


src/use-swr.ts
```diff
@@ -165,6 +165,11 @@ async function mutate<Data = any>(
...
   }
 
+  // clear promise of revalidate whether or not `revalidate({ dedupe: false })`
+  // be called subsequently.
+  delete CONCURRENT_PROMISES[key]
+  delete CONCURRENT_PROMISES_TS[key]
+
   // enter the revalidation stage
   // update existing SWR Hooks' state
   const updaters = CACHE_REVALIDATORS[key]
```
promer94, line 175: It might be batter if we delete the promise conditionally
```suggestion
 const updaters = CACHE_REVALIDATORS[key]
 if (updaters && updaters.length > 0) {
    const promises = []
    for (let i = 0; i < updaters.length; ++i) {
      promises.push(
        updaters[i](!!shouldRevalidate, data, error, undefined, i > 0)
      )
    }
    // return new updated value
    return Promise.all(promises).then(() => {
      if (error) throw error
      return cache.get(key)
    })
  } else {
    // clear dedupingInterval of an unmounted key
    delete CONCURRENT_PROMISES[key]
    delete CONCURRENT_PROMISES_TS[key]
  }
```

MoonBall: Thanks for your suggestions.

I deleted the promise conditionally at first. I think the conditional statement should be `!!shouldRevalidate && (updaters && updaters.length > 0)` if deleting it conditionally. This conditional statement is little of complicated and hard to understand.

However, Deleting the promise after mutating is easier to understand. I would like to modify my comment to express it clearly.

MoonBall: > It might be batter if we delete the promise conditionally

Thanks for your suggestions.

I deleted the promise conditionally at first. I think the conditional statement should be `!!shouldRevalidate && (updaters && updaters.length > 0)` if deleting it conditionally. This conditional statement is little of complicated and hard to understand.

However, Deleting the promise after mutating is easier to understand. I would like to modify my comment to express it clearly.

promer94:
I think the root cause of the original issue might not be the dedupingInterval ?
```js
import { mutate } from 'swr'
mutate('key')
```

When using the global mutate
-  If there are mounted hook instanced - calling ```mutate('key')``` will revalidate the data and update the cache using the hook's fetcher
-  If all the hooks are unmounted - calling ```mutate('key')``` will do nothing - you need provide the data or function to update the cache

https://github.com/vercel/swr/blob/f7876240094c02c9933757d3e8d568185122adfb/src/use-swr.ts#L391-L402

The dedupingInterval will be ignored by these conditions.

MoonBall:
You are right. But The `mutate('key')` should support to clear revalidate promise although not providing the  second param.  

Some backend post api doesn't return the data, so RD should use `useSWR` as follow:

```js
fetch('/api/user/edit', { method: 'post', name })
mutate('/api/user')
```

and then they want to fetch new data if they need.

MoonBall:
`mutate('key')` has different behavior according to  if there are mounted hook. This API should be better.

MoonBall:
@promer94 Thanks for your review.

It's better to keep `mutate()` to revalidate when `shouldRevalidate` is true. It's a good API design.

I made a new commit, PTAL.

src/use-swr.ts
```diff
@@ -180,6 +180,11 @@ async function mutate<Data = any>(
       if (error) throw error
       return cache.get(key)
     })
+  } else if (shouldRevalidate) {
+    // `updaters` would trigger revalidate if `updaters.length > 0`.
+    // So clear promise of revalidate if `!updaters.length && shouldRevalidate`
+    delete CONCURRENT_PROMISES[key]
+    delete CONCURRENT_PROMISES_TS[key]
   }
```
promer94, line 188: if hook is unmounted
```
mutate('key') // This will not clear the deduplicatedInterval

mutate("key", () => undefined) // This will clear the deduplicatedInterval
mutate("key", () => Promise.resolve(undefined)) This will also clear the deduplicatedInterval
```

https://github.com/vercel/swr/blob/f7876240094c02c9933757d3e8d568185122adfb/src/use-swr.ts#L107

promer94 (commented):
I still have some concerns about clearing the deduplicatedInterval in mutate function. I am not sure this is the best way to solve the problem. We might need more discussions about this.

src/use-swr.ts
MoonBall: if hook is unmounted, the above three cases will not clear the deduplicatedInterval because `updaters && updaters.length` is true.

MoonBall: I has an irrelevant problem. How did you embed code like this:

image

promer94: https://docs.github.com/cn/github/managing-your-work-on-github/creating-a-permanent-link-to-a-code-snippet

shuding (commented):
Thank you! I think this is a pretty neat trick to work around that special case. I'll look into it deeper to see if there's any other problems, since this is not a perfect solution.

PR #1150, opened 2021-05-02
Code optimizations

shuding:
Did a couple of code optimizations in this PR for better structure and code reusing. Also handles the case that `undefined` is manually set to another value<sup>*</sup>.

As a result, here's the size of the built `dist/index.js` with this PR:
- Before: 12072B
- After: 11361B

_*Although Rollup handles it automatically by changing it to `void 0`, it's great to have it in the source code. This trick also makes the built file smaller._


src/libs/hash.ts
```diff
@@ -12,23 +12,18 @@ export default function hash(args: any[]): string {
...
       // undefined   -> 'undefined'
       // 123         -> '123'
       // "null"      -> '"null"'
-      if (typeof args[i] === 'string') {
-        _hash = '"' + args[i] + '"'
-      } else {
-        _hash = String(args[i])
-      }
+      // null        -> 'null'
+      _hash = JSON.stringify(args[i])
```
huozhi, line 26: `JSON.stringify(undefined)` seems returning `undefined` itself instead of a string

src/env.ts
```diff
@@ -3,16 +3,30 @@ import { useEffect, useLayoutEffect } from 'react'
 export const IS_SERVER =
   typeof window === 'undefined' ||
   // @ts-ignore
-  !!(typeof Deno !== 'undefined' && Deno && Deno.version && Deno.version.deno)
+  !!(typeof Deno !== 'undefined' && Deno.version && Deno.version.deno)
```
promer94, line 6:
```suggestion
import { isUndefined } from './libs/helper'
export const IS_SERVER =
  isUndefined(window) ||
  // @ts-ignore
  !!(!isUndefined(Deno) && Deno.version && Deno.version.deno)
```

src/config.ts
```diff
@@ -15,7 +16,7 @@ function onErrorRetry(
   revalidate: Revalidator,
   opts: Required<RevalidatorOptions>
 ): void {
-  if (!webPreset.isDocumentVisible()) {
+  if (!webPreset.isVisible()) {
```
huozhi, line 19: it's not a documented option though, we can mention it in this PR as part of release notes. maybe some users are using custom `isDocumentVisible` now. like RN users

src/use-swr.ts
```diff
@@ -24,6 +25,10 @@ import {
 
 type Revalidator = (...args: any[]) => void
 
+// Generate strictly increasing timestamps.
+let __timestamp = 0
```
huozhi, line 29: do you feel it's okay to start timestamp from `1` to avoid some issue like `!now()` ?

src/env.ts
shuding: I don't think you can use an undeclared variable (like `window`) before checking it with `typeof`:

image

src/libs/hash.ts
shuding: True but later it will run `key += '@' + _hash` which converts it to string.

src/config.ts
shuding: Yeah, I think I'll just revert this change for now until we have an ultimate solution for RN.

src/use-swr.ts
shuding: :+1: (where are we using `!now()` btw?)

src/libs/hash.ts
promer94: we used to have this issue in #1075

src/use-swr.ts
promer94: we used to have this issue in #1075

shuding: I just checked again, we are using `const now = () => ++__timestamp` so it will be starting from 1 I think.

huozhi (approved):

PR #1159, opened 2021-05-08
fix global mutate generic

### LINKED ISSUE #1155: TS: global mutate() is not generic
# Bug report

I cannot set a generic type to `mutate()` calls.

## Expected Behavior

I should be able to set a generic like so:

```ts
mutate<MyType>(...)
```

## Repro Steps / Code Example

```ts
import { mutate } from 'swr';

interface MyType {
  text: string;
}

mutate<MyType>('');
```
```
src/test.ts7:8 - error TS2558: Expected 0 type arguments, but got 1.

7 mutate<MyType>('');
         ~~~~~~
```

## Additional Context

SWR version: 0.5.6
I had tried fixing it in #1069 but it looks like things have changed since then. 😔 

image

It returns `Promise<any>` which isn't quite safe.

promer94:
fix #1155

https://codesandbox.io/s/mutate-generic-e3fqj?file=/src/App.tsx

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -90,11 +90,20 @@ export type Mutator<Data = any> = (
   shouldRevalidate?: boolean
 ) => Promise<Data | undefined>
 
-export type ScopedMutator<Data> = (
-  key: Key,
-  data?: Data | Promise<Data> | MutatorCallback<Data>,
-  shouldRevalidate?: boolean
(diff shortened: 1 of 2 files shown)


shuding (approved):

shuding:
Thanks!

PR #1167, opened 2021-05-19
allow mutation with regex key

### LINKED ISSUE #761: Feature request: Please add partial matching for mutation keys
# Feature request

My previous issue was completely removed for some strange reason (not even closed) so I am opening a new one. Right now if you want to use the global mutate function you have to provide a complete match with the key. It's very weird and even react-query supports partial matching. Please add it as its very crucial. I cant replicate every key in every component.

j-jiseophan:
resolve #761 

### Description
allow mutation with RegExp as a key

### Etc
enable `downlevelIteration` in tsconfig for using `for .. of` statement with Map iterator

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -85,7 +85,7 @@ export type State<Data, Error> = {
 
 export type Mutator<Data = any> = (
   cache: Cache,
-  key: Key,
+  key: MutatorKey,
   data?: Data | Promise<Data> | MutatorCallback<Data>,
   shouldRevalidate?: boolean
 ) => Promise<Data | undefined>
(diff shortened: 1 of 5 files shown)


j-jiseophan:
> I think we should handle this with the new `createCache` API in user-land, and provide some snippets in our docs (under advanced). CC @huozhi We should work on docs 😬.

Sorry for my poor English
Which one do you mean?
1) The feature doesn't need to be implemented inside library. Instead, it should be implemented in user side with `createCache `API
or 
2) The feature(PR) should also support custom cache layer by `createCache`

huozhi:
hi @j-jiseophan we're working on documentation and examples with option 1 given above, by using `createCache` API from swr@beta.
> The feature doesn't need to be implemented inside library. Instead, it should be implemented in user side with createCache API

the approach can be:
- user gets returned `mutate` from `createCache` API with custom `provider` (can be a Map instance, or any other user preferred storage)
- wrap the `mutate` into custom function that can receive regex or glob pattern string as keys
- access keys from the `provider`,  detect the matching ones and schedule mutations

We'll update it into swr docs and examples later and let people acknowledged in the issue, hope these information can help close this PR.

PR #1174, opened 2021-05-21
enhance: add $ prefix for non-major keys

### LINKED ISSUE #242: How to check if there are no pending requests?
I have multiple independent fetch requests located in different files and initiated by useSWR.

How can I know when the last request is done, i.e. there are no pending requests?

I'd like to know the number of active requests initiated by useSWR.

My question is similar to to this [jQuery question](https://stackoverflow.com/questions/1822913/how-do-i-know-if-jquery-has-an-ajax-request-pending).

huozhi:
specialize keys for `error` and `isValidating`, and some non major keys in swr infinite, to identify them easier later if developers want to play tricks with keys or filter them from cache provider.

Resolves #242 

e.g.

### access all keys

```js
const provider = new Map()
const { cache, mutate } = createCache(provider)

// ...
const serializedKeys = Array.from(provider.keys()).filter(k => !k.startsWith('$'))
```

### access pending requests

```js
const validatingKeys = serializedKeys.filter(key => key.startsWith('$req$'))
```

### DIFF
--- a/src/use-swr-infinite.ts
+++ b/src/use-swr-infinite.ts
@@ -100,7 +100,7 @@ function useSWRInfinite<Data = any, Error = any>(
 
   // actual swr of all pages
   const swr = useSWR<Data[], Error>(
-    firstPageKey ? ['$inf', firstPageKey] : null,
+    firstPageKey ? 'inf@' + firstPageKey : null,
     async () => {
       // get the revalidate context
       const { data: originalData, force } = cache.get(contextCacheKey) || {}
--- a/src/use-swr-infinite.ts
+++ b/src/use-swr-infinite.ts
@@ -41,13 +41,13 @@ function useSWRInfiniteHandler<Data = any, Error = any>(
   // here we get the key of the fetcher context cache
   let contextCacheKey: string | null = null
   if (firstPageKey) {
-    contextCacheKey = '$ctx@' + firstPageKey
+    contextCacheKey = '$ctx$' + firstPageKey
   }
 
   // page size is also cached to share the page data between hooks having the same key
   let pageSizeCacheKey: string | null = null
   if (firstPageKey) {
-    pageSizeCacheKey = '$len@' + firstPageKey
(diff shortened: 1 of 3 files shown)


shuding (commented):
Let's also add it to `inf`, `len` and `ctx` in swr-infinite 👍

huozhi:
curious about the reason of designing page key  as `firstPageKey ? ['$inf', firstPageKey] : null`, then the array key will always be serialized into `arg@blablabla` which losing the `$` prefix

shuding:
Oh that's a good point, since it already has `$args`.

> curious about the reason of designing page key as firstPageKey ? ['$inf', firstPageKey] : null

That's because if we directly use `firstPageKey` as the key (without adding `inf`), the infinite page hook will share cache with the single page hook if it exists (`useSWR(firstPageKey)`).

src/use-swr-infinite.ts
```diff
@@ -100,7 +100,7 @@ function useSWRInfinite<Data = any, Error = any>(
 
   // actual swr of all pages
   const swr = useSWR<Data[], Error>(
-    firstPageKey ? ['inf', firstPageKey] : null,
+    firstPageKey ? ['$inf', firstPageKey] : null,
```
huozhi, line 103:
```suggestion
    firstPageKey ? 'inf@' + firstPageKey : null,
```
since firstPageKey is already a string here, how about giving it a `inf@` prefix? what's your thought here? @shuding

shuding: Sounds great to me! @huozhi

shuding: Why not adding `$` btw?

shuding:
Looks good to me, let's also change existing `@` concatenation symbols to `$` so it looks better :)

shuding (approved):
> make symbol rich

Thanks! 🤣

vaynevayne:
Is it only supported in custom provider mode?

PR #1183, opened 2021-05-29
fix: autocomplete suggestions example

### LINKED ISSUE #1129: Error on example "autocomplete-suggestions"
# Bug report

Example autocomplete-suggestions not working on my environment. it has problem with useLayoutEffect.
I think it cause by @reach/combobox. but I did not find why it is.

I made [sandbox](https://codesandbox.io/s/exciting-haze-ej4vd) for this issue. I works well. 

```log
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://fb.me/react-uselayouteffect-ssr for common fixes.
    in ForwardRef(ComboboxInput) (at pages/index.js:28)
    in div
    in Context.Provider
    in ForwardRef(Combobox) (at pages/index.js:27)
    in div (at pages/index.js:25)
    in Unknown
    in App
    in Unknown
    in Context.Provider
    in Context.Provider
    in Context.Provider
    in AppContainer
```

## Expected Behavior

## Repro Steps / Code Example

https://codesandbox.io/s/exciting-haze-ej4vd

## Additional Context

SWR version : 0.5.4

koba04:
Fixes #1129 
The issue is caused by `@reach/combobox`, so I've updated the version to fix it.

### DIFF
--- a/examples/autocomplete-suggestions/package.json
+++ b/examples/autocomplete-suggestions/package.json
@@ -4,7 +4,7 @@
   "main": "index.js",
   "license": "MIT",
   "dependencies": {
-    "@reach/combobox": "0.5.3",
+    "@reach/combobox": "0.15.1",
     "isomorphic-unfetch": "3.0.0",
     "next": "9.3.3",
     "react": "16.11.0",


huozhi (approved):
Thanks!

PR #1190, opened 2021-05-31
createCache breaks mutate fn

JCB-K:
While testing the  custom cache in the latest Beta release (great feature!) I noticed a bug: it's not possible to mutate data when using a custom cache & importing the mutate function. I'm pretty sure the issue is located here:

https://github.com/vercel/swr/blob/5a05547df9aa700d634510f629d00c9f4cbdb9c7/src/use-swr.ts#L700-L703

The mutate function is bound to the defaultConfigs cache, regardless of whether a custom cache is used or not. The existing test for mutating when using a global cache passes, because it takes the mutate function from the return value of `createCache`, which is bound.

I don't really know where to go from here, but I thought It'd start by contributing a failing test :). Any pointers as to how I could fix this?

### DIFF
--- a/test/use-swr-cache.test.tsx
+++ b/test/use-swr-cache.test.tsx
@@ -1,6 +1,10 @@
 import { act, fireEvent, render, screen } from '@testing-library/react'
 import React, { useState } from 'react'
-import useSWR, { createCache, SWRConfig } from '../src'
+import useSWR, {
+  createCache,
+  mutate as importedMutate,
+  SWRConfig
+} from '../src'
 import { sleep, createKey } from './utils'
 
 describe('useSWR - cache', () => {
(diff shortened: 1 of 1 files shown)


shuding:
Thank you for providing a test! I'll keep this PR open and take a look. cc @huozhi

huozhi:
@JCB-K `createCache` will also return a `mutate` from it, which is bound to the provider for manipulation. you should use it instead the global one.

```diff
- const { cache } = createCache(provider)
+ const { cache, mutate } = createCache(provider)
```

shuding:
We've updated the corresponding docs, you should use the `mutate` returned by `createCache`, which is bound to that cache and is also global.

PR #1213, opened 2021-06-12
fix: onErrorRetry is not de-duped

simalicrum:


shuding:
Thank you for opening this PR, it should indeed be deduplicated and the behavior is fixed now when resolving another issue, by only firing the error events for the original request.

Issue #141, opened 2019-11-20
Status indicator


conradreuter:
I think the result should have a field indicating the status of the query. It could be a simple boolean flag `loading` or a more sophisticated `status` enum (to indicate revalidation, etc.).

Checking `data` for truthiness does not cut it if you consider that the data returned from the fetcher could in theory be `null` (or even `undefined`).

quietshu:
There’s the `isValidating` value returned by `useSWR`, you can simply use it as the loading indicator.

aequasi:
You can’t reliably use that to check if dat has not been set yet though

pkhodaveissi:
That'd be great, react-async package has this great enum:

```export enum StatusTypes {
  initial = "initial",
  pending = "pending",
  fulfilled = "fulfilled",
  rejected = "rejected",
}```

**Comment by ivan-kleshnin** [comment]
> Checking data for truthiness does not cut it if you consider that the data returned from the fetcher could in theory be null (or even undefined).

Much worse, `empty data = loading` assumption works only for the first request. 
For all consequent requests (refetches) from the same `useSWR` data will not be empty.

**Comment by darrenjennings** [comment]
I have thought about adding this feature to `swrv` (vue implementation) because right now we use a user-land hook to determine the enum, which uses data/error/isValidating. It is kind of annoying that it's not built-in to the main return value.

https://docs-swrv.netlify.app/features.html#useswrvstate

```ts
const STATES = {
  PENDING: 'PENDING', // no data, initial loading
  VALIDATING: 'VALIDATING', // there is data, but validation is occurring
  SUCCESS: 'SUCCESS', // data exists and no in-flight requests
  ERROR: 'ERROR', // error, no data
  STALE_IF_ERROR: 'STALE_IF_ERROR', // error exists, but there is still some cached data we can show
}
```

shuding:
Thanks for sharing @darrenjennings! I'll also add my 2 cents.

I love the idea, but enum is a bit annoying to me. What do you think about this proposal:

```js
const { data, error, state } = useSWR(...)

state.pending // => true
state.success // => false
state.error // => false
```

The good thing is that users don't need to import all the enums, and the editor will autocomplete. 

---

The only concern for this status indicator idea is, since we are building these "high-level" states, their names have to be very clear and have to be **complete**.

**Complete** means that we have a group of 3 base values (data, error, isValidating) and all of them have 2 states, the derived states will be 2^3 = 8 different possibilities. It grows exponentially:

[...]

PR #1215, opened 2021-06-15
[RFC] feat: Use a `status` enum to label the current stage of a SWR hook

icyJoseph:
Hi!

I've been using this library, and I've collected some feedback from colleagues or people I show this hook to. 

One persistent subject is that, often developers want to derive the `status` of the hook, that is, is it loading? is it validating? is there a difference between them at all? is there any data, etc. Refer to: https://github.com/vercel/swr/discussions/563

I understand that we should open up discussions for these subjects first, but in this case it seemed best to talk over the code, because of the change itself and because I am short on bandwidth as of now.

I'll create a discussion thread and link both. Hopefully it doesn't create confusion.

At the core of this PR, we derive a new key called `status` which can be `loading`, `stale`, `error` or `validating`.

In addition, I noticed that `isValidating` doesn't really settle on error cases. This might introduce breaking changes for users.

I included some tests which helpfully help understanding the subject matter of the feature.

Last but not least, more documentation is needed. 

- [ ] Documentation
- [ ] Analyze possible breaking changes
- [ ] More units tests
- [ ] Infinite SWR

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -151,6 +151,7 @@ export interface SWRResponse<Data, Error> {
   revalidate: () => Promise<boolean>
   mutate: KeyedMutator<Data>
   isValidating: boolean
+  status: 'loading' | 'validating' | 'error' | 'stale'
 }
 
 export type KeyLoader<Data = any> =
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -447,7 +447,6 @@ function useSWR<Data = any, Error = any>(
 
         if (stateRef.current.error !== err) {
           // we keep the stale data
-          // cache.set(keyValidating, false)
           setState({
             isValidating: false,
             error: err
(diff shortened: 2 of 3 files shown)


huozhi:
As https://github.com/vercel/swr/issues/141#issuecomment-792077186 said, composing `data`, `error` and `isValidating` will result in 8 cases that make `status` enums hard to remeber for users. IMO it's better to keep automic for returned values and let users compose them flexibly on top of swr.

shuding:
Also worth noting, #1160 might help this. We can build a status middleware to attach a status enum to the SWR return object.

phaux:
It would be even better if the response was a tagged union:

```ts
type SWRResponse<Data, Error> =
  | {
    status: "loading"
    isValidating: boolean
    ...
  }
  | {
    status: "error"
    error: Error
    isValidating: boolean
    ...
  }
  | {
    status: "validating"
    data: Data
    isValidating: true
    ...
  }
  | {
    status: "stale"
    data: Data
    isValidating: false
    ...
  }
```

PR #1239, opened 2021-06-24
chore: bump typescript to use ts-expect-error instead of ts-ignore

koba04:
This PR bumps typescript to 3.9.10 so that we can use `ts-expect-error` instead of `ts-ignore`. 
I've tried to bump typescript to the latest version, but the build with `bunchee` was failed. To fix this, I would have to update some dependencies that `bunchee` has and I'll work on it later.

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -26,7 +26,7 @@ export const useIsomorphicLayoutEffect = IS_SERVER ? useEffect : useLayoutEffect
 // slow connection (<= 70Kbps)
 export const slowConnection =
   !IS_SERVER &&
-  // @ts-expect-error
+  // @ts-ignore
   navigator['connection'] &&
-  // @ts-expect-error
+  // @ts-ignore
   ['slow-2g', '2g'].indexOf(navigator['connection'].effectiveType) !== -1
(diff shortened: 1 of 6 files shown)


koba04:
I've converted this to a draft status to fix failing tests.

src/env.ts
```diff
@@ -2,17 +2,15 @@ import { useEffect, useLayoutEffect } from 'react'
...
 
 const __requestAnimationFrame = !IS_SERVER
   ? window['requestAnimationFrame']
   : null
 
 // polyfill for requestAnimationFrame
-export const rAF = IS_SERVER
-  ? null
-  : __requestAnimationFrame
+export const rAF = __requestAnimationFrame
```
koba04, line 13: I think we don't have to assign `null` to `rAF` for a server environment because `setTimeout` is enough and we don't use it on a server envrionment.

koba04: We should use `IS_SERVER` if we'd like to have code that only runs on servers.

shuding (approved):
Thanks!

koba04:
Thank you!

PR #1253, opened 2021-07-01
Add tslib

shuding:
Add `tslib` and enable `importHelpers` for tsc.

### DIFF
--- a/immutable/tsconfig.json
+++ b/immutable/tsconfig.json
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "outDir": "./dist",
     "rootDir": "..",
+    "importHelpers": true,
   },
   "include": [".", "../src"]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -10,6 +10,7 @@
     "noImplicitReturns": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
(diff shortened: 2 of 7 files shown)


tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "noImplicitReturns": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
+    "importHelpers": true,
```
huozhi, line 13: tslib is dep of bundler, bundler should do that for you.
`importHelpers` is overriden to `false` for now in bunchee. I guess I need to change a few behavior to let user customize it..

immutable/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "outDir": "./dist",
     "rootDir": "..",
+    "importHelpers": true,
```
huozhi, line 6: ok looks like it's overriden in rollup ts plugin so guess this can be removed? I'll remove the one on bundler side

shuding: I think it's still necessary to keep it here so we know it's enabled explicitly (for those who don't know about rollup plugins), and how we can disable it.

huozhi:
Can be closed now since outupt in v2 is es2018

PR #1254, opened 2021-07-01
fix: fail accessing navigator on Node

koba04:
`navigator` doesn't exist on the Node environment, so it is required to check if `navigator` is not undefined.


shuding (approved):
Thanks!

koba04:
Thank you!

PR #1255, opened 2021-07-01
chore: fix to run the build script multiple times

koba04:
Currently, `yarn build` is failed when the `dist/` directories already exist, so I've added `dist/` directories into the `exclude` setting to ignore them. I've also remove them on the `prepublishOnly` script.


shuding (approved):
Thanks, good catch!

koba04:
Thank you!

PR #1256, opened 2021-07-01
chore: run tests without doing the build

koba04:
After #992, we have to run a build script to run tests because now tests use files in `dist/` instead of TS files.
I guess that the change is related to start using the exports filed in `package.json`.

This is a bit error-prone, so I'd like to use TS files for unit testing. Does it make sense?

### DIFF
--- a/immutable/index.ts
+++ b/immutable/index.ts
@@ -1,5 +1,4 @@
-// @ts-ignore
-import useSWR from 'swr'
+import useSWR from '../src/use-swr'
 
 import { withMiddleware } from '../src/utils/with-middleware'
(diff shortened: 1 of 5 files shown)


immutable/index.ts
```diff
@@ -1,5 +1,4 @@
-// @ts-ignore
-import useSWR from 'swr'
+import useSWR from '../src/use-swr'
```
shuding, line 1: It's necessary to use `import useSWR from 'swr'`, so the `swr` core doesn't get bundled into this sub-package (`swr/immutable`). Preact also uses this approach.

koba04: Thank you! I understand the intention.

PR #1257, opened 2021-07-01
feat: add getInfiniteKey to mutate the cache for useSWRInfinite

### LINKED ISSUE #1156: Cannot use global mutate with useSWRInfinite
# Bug report

There is no way to mutate data fetched with `useSWRInfinite` without first fetching the data client-side. I saw [this answer](https://github.com/vercel/swr/pull/435#issuecomment-658219974) by @shuding and it definitely works, but it triggers a fetch when all I want to do is mutate data (if it has already been fetched).

## Expected Behavior

Instead, I should be able to mutate local data without triggering a fetch. The global `mutate` function should allow me to mutate data **that has already been fetched** (e.g. on a page that uses `useSWRInfinite`) on a page that should not trigger a fetch of that data.

My (pretty common I think) use-case:
1. I have a list of emails fetched using `useSWRInfinite` on a `/feed` page.
2. When the user opens an email, they can modify it on a `/messages/[id]` page.
3. I want to mutate the data shown on the `/feed` page when the data is edited on the `/messages/[id]` page.

But, the user can also open the `/messages/[id]` page directly, in which case the `/feed` page will never have been opened and `useSWRInfinite` will never have been called. In this case, I don't want to trigger a fetch with `useSWRInfinite`.

## Repro Steps / Code Example

This is the project I'm working on: [`nicholaschiang/hammock`](https://github.com/nicholaschiang/hammock).

## Additional Context

SWR version: 0.5.5
Related to #435.

koba04:
fixes #1156

~This PR adds a new API called `mutateInfinite`, which is a mutation function for `useSWRInfinite` and based on https://github.com/vercel/swr/issues/1156#issuecomment-852140879.~

This PR adds a new API called `getInfiniteKey` to get a key to mutate cache data for `useSWRInfinite`, which receives a key loader function and returns the key for the cache data.

I feel this is still in discussion and have not meet an agreement, so this is just a POC.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,5 @@
 // Core APIs
-export { SWRConfig, mutate, createCache, mutateCustomCache } from './use-swr'
+export { SWRConfig, mutate, createCache } from './use-swr'
 
 // useSWR
 import useSWR from './use-swr'
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,5 @@
 // Core APIs
-export { SWRConfig, mutate, createCache } from './use-swr'
+export { SWRConfig, mutate, createCache, mutateCustomCache } from './use-swr'
 
 // useSWR
 import useSWR from './use-swr'
--- a/test/use-swr-infinite.test.tsx
+++ b/test/use-swr-infinite.test.tsx
@@ -1,7 +1,6 @@
 import React, { useEffect, useState } from 'react'
 import { render, fireEvent, act, screen } from '@testing-library/react'
-
-import { mutate } from 'swr'
+import { mutate,createCache, SWRConfig } from 'swr'
 import useSWRInfinite, { mutateInfinite } from 'swr/infinite'
 import { sleep, createResponse } from './utils'
 
@@ -594,6 +593,34 @@ describe('useSWRInfinite', () => {
     await act(() =>
       mutateInfinite(index => `page-test-12-${index}`, 'local-mutation', false)
     )
+    await screen.findByText('data:local-mutation')
+  })
+
+  xit('should be able to use mutateInfinite with a custom cache', async () => {
+    const key = 'page-test-13;'
+    const customCache1 = new Map([[key, 'initial-cache']])
+    const { cache } = createCache(customCache1)
+    function Page() {
+      const { data } = useSWRInfinite<string, string>(
+        () => key,
+        () => createResponse('response data')
+      )
+      return <div>data:{data}</div>
+    }
+
+    render(
+      <SWRConfig value={{ cache }}>
+        <Page />
+      </SWRConfig>
+    )
+    screen.getByText('data:')
+
+    await screen.findByText('data:initial-cache')
+
+    await act(() => mutateInfinite(() => key))
+    await screen.findByText('data:initial-cache')
+
+    await act(() => mutateInfinite(() => key, 'local-mutation', false))
     await screen.findByText('data:local-mutation')
   })
 })
--- a/test/use-swr-infinite.test.tsx
+++ b/test/use-swr-infinite.test.tsx
@@ -1,7 +1,7 @@
 import React, { useEffect, useState } from 'react'
 import { render, fireEvent, act, screen } from '@testing-library/react'
-import { mutate,createCache, SWRConfig } from 'swr'
-import useSWRInfinite, { mutateInfinite } from 'swr/infinite'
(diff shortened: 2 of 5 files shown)


shuding:
Love this. I got another question, how can we have the global mutate function when we are using the cache provider?

src/index.ts
```diff
@@ -1,5 +1,5 @@
 // Core APIs
-export { SWRConfig, mutate, createCache } from './use-swr'
+export { SWRConfig, mutate, createCache, mutateCustomCache } from './use-swr'
```
huozhi, line 2: not sure if we want to expose the internalMutate here since `createCache` has already covered the similar usage

test/use-swr-infinite.test.tsx
```diff
@@ -572,4 +572,55 @@ describe('useSWRInfinite', () => {
...
+    render(<Page />)
+    screen.getByText('data:')
+
+    await screen.findByText('data:page-test-12-0:1')
+
+    await act(() => mutateInfinite(index => `page-test-12-${index}`))
+    await screen.findByText('data:page-test-12-0:2')
+
+    await act(() =>
+      mutateInfinite(index => `page-test-12-${index}`, 'local-mutation', false)
```
huozhi, line 594: users have the `keyLoader` definition, at the same time they also have the `createCache` to get the `mutate` function, then combining both of them can also achieve the similar approach?

koba04:
@shuding 
> how can we have the global mutate function when we are using the cache provider?

That concern makes sense.
Currently, the global mutate function seems not to work correctly with a custom cache. Is it correct?

I've added `mutateCustomCache` as a mutate function that works with a custom cache for `mutateInfinite` https://github.com/vercel/swr/pull/1257/commits/7424dcc2721c32871b53ff96443f383e0a67f502.
`mutateInfinite` works correctly with a custom cache by using the `mutateCustomCache` https://github.com/vercel/swr/pull/1257/commits/efb14597eae1afee34874bbbf35806161ef3ccc3.

But I feel this is not the best way for the feature.
Before working this, should we fix the issue how the global mutate function deals with a custom cache?

src/index.ts
koba04: Yeah, I think it's not the best way to support the use case. We might have to find out another solution like exporting a function that returns a key like `INFINITE_PREFIX + getFirstPageKey(getKey)` rather than exporting a mutate funciton.

test/use-swr-infinite.test.tsx
koba04: Do you mean like `mutate(getInfiniteKey(getKey), 'local-mutation', false)`?

huozhi: yes, that would be nice!

koba04:
I've updated this PR to add `getInfiniteKey` rather than `mutateInfinite`, which is based on the @huozhi's [comment](https://github.com/vercel/swr/pull/1257#discussion_r664609730).

Implementing `getInfiniteKey` is simpler than `mutateInfinite` and is able to support a custom cache 👍

shuding (approved):
Thank you! I think this is the best choice for now until we have a better way to work around custom cache providers.

koba04:
Thank you!🚀

nicholaschiang:
@shuding @koba04 Have you added this API to the documentation yet? If not, could you?

nicholaschiang:
@shuding could you please release this?

shuding:
@nicholaschiang A new beta released is there, please check! We will document new APIs when it's stable.

nicholaschiang:
@shuding this doesn't seem to support mutating based on current data. It's not included in [your tests](https://github.com/vercel/swr/blob/dc1444f29cc4eba1fe5c4a66ff426d2e142aba82/test/use-swr-infinite.test.tsx#L602-L607) and it's not working here:

```typescript
// I have an updated message and I need to mutate the list of messages.
const message = { ...prev, subject: 'This is an updated value.' };
await mutate(
  getInfiniteKey(getKey), 
  (data?: MessagesRes[]) => data?.map((messages: MessageRes) => {
    const idx = messages.findIndex((m) => m.id === message.id);
    if (idx < 0) return messages;
    return [...messages.slice(0, idx), message, ...messages.slice(idx + 1)];
  }), 
  false
);
```

koba04:
@nicholaschiang I've made a PR to confirm your case. 
https://github.com/vercel/swr/pull/1319
This seems to work well in the test case.

PR #1259, opened 2021-07-01
test: run concurrent rendering tests with React v18 alpha

koba04:
React has announced that they released v18 alpha, so I think it's better to use an alpha version rather than an experimental version.

https://reactjs.org/blog/2021/06/08/the-plan-for-react-18.html

I've also fixed API changes in the alpha version.


shuding (approved):
Thanks a lot for this!

koba04:
Thank you!

PR #1260, opened 2021-07-01
enhance: use ts for testing

huozhi:
Closes #1256 

* Use ts source files for testing
* upgrade jest suite, jest 27 requires setting test version to `jsdom`

PR #1261, opened 2021-07-02
chore: add watch scripts for sub packages to build them

koba04:
~This is an alternative of #1256 
Currently, we have to build each packages to run tests, which is a bit cumbersome,  and I've added npm-scripts to watch and build each package.~

~With this PR, we don't have to run a build script manually for unit testing.~

I've reopened this because I think this is useful to use local builds to run an application.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,10 @@
   "license": "MIT",
   "scripts": {
     "build": "yarn build:core && yarn build:infinite && yarn build:immutable",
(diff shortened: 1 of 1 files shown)


koba04:
Ah, I missed #1260, which is great. 👍 👍 👍

shuding (approved):
Thanks!

koba04:
Thank you!

Issue #1082, opened 2021-03-28
Manually triggering new requests


vinay-deshmukh:
In my application, I don't want `useSWR` hook to automatically call the fetcher function. I want to call it in one of my event handlers.

## What did you try?

- I tried setting `refetchOnMount` as false, but then on the next render, the fetch request happened anyway.

- Tried making `isPaused()` return true always, but this means `mutate()` calls also don't work.

##Expected behaviour

Flag/boolean config that disables "auto fetching", and only fetches when user calls `mutate()`.

promer94:
currently you could use ```isPaused``` to build a customized hook.

```js
function useControlledSWR(key, fetcher) {
  const [control, setControl] = React.useState(true);
  const swr = useSWR(key, fetcher, {
    isPaused: () => {
      return control;
    }
  });
  const { mutate } = swr;
  const start = () => {
    setControl(false);
  };

  const pause = () => {
    setControl(true);
  };

  React.useEffect(() => {
    if (!control) {
      mutate();
      /* auto pause
      .then((v) => pause());
      */
    }
  }, [control, mutate]);

  return [swr, { status: control, start, pause }];
}
```

FYI: https://codesandbox.io/s/control-swr-vf9gu?file=/src/App.js

vinay-deshmukh:
I actually don't want to start/stop the hook. I just need a stable mutate callback that will retrigger the fetching.

I tried showing a loading message using your hook, but it seems your hook is eternally in a loading state for some reason, until I call button.

![image](https://user-images.githubusercontent.com/32487576/112789853-a9dad480-907b-11eb-8c62-2105c358affd.png)
I added that line and your hook is eternally in a loading state.

https://codesandbox.io/s/useswr-try-lezg7

This sandbox shows what I want to achieve.

It will work once your comment `isPaused()` config option. Once that is done, it will only run if you call mutate, but if the component rerenders, the fetcher fucntion gets run which is undesirable.

shuding:
FWIW – You can use `useSWR(enabled ? 'key' : null)` to enable/disable the hook. But #1450 will give you the ability to manually trigger a fetch. Going to change the issue description for accuracy.

shuding:
#1450 is merged, and we can close this issue now :)

anthonyalayo:
@shuding I feel like `useSWRMutation` doesn't solve the use case.

[...]

Issue #116, opened 2019-11-13
Can SWR totally replace react-apollo?


mdegrees:
I'm maintaining a Next js application that relies on Apollo to fetch data, infinite scroll and do updates in realtime (chat like updates). SWR seems to work better with Next js with no boilerplate code. I'm considering migrating to SWR but have no idea how to replicate some react-apollo features like components subscribing to updates via Graphql. Are there any examples of SWR working with graphql Subscriptions as in react-apollo https://www.apollographql.com/docs/react/data/subscriptions/?

mdegrees:
Although I generally understand this example: https://github.com/zeit/swr/issues/41#issuecomment-548190601,  I have no idea how on a socket event, a `newComment` for e.g, we update the data from `api/allComments`.

Can you please provide a full cycle example? something like fetching api/allComments > emitting a new comment > reacting to the new comment by appending it to api/allComments data.

With react apollo, it's something like: 
```
const GET_COMMENTS_QUERY = gql`
  query GetComments() {
      comments {
        id
        content
      }
  }
`;

const COMMENTS_SUBSCRIPTION = gql`
  subscription onCommentAdded() {
    commentAdded() {
      id
      content
    }
  }
`;

function CommentsPage({ params }) {
  const { subscribeToMore, ...result } = useQuery(GET_COMMENTS_QUERY);

  return (
    <CommentsPage
      {...result}
      subscribeToNewComments={() =>
        subscribeToMore({
          document: COMMENTS_SUBSCRIPTION,
          variables: { repoName: params.repoName },
          updateQuery: (prev, { subscriptionData }) => {
            if (!subscriptionData.data) return prev;
            const newFeedItem = subscriptionData.data.commentAdded;

            return Object.assign({}, prev, {
              entry: {
                comments: [newFeedItem, ...prev.entry.comments]
              }
            });
          }
        })
      }
    />
  );
}```

Thank you!

vasco3:
can the swr mutation run a refetchQueries like apolo useMutation does?

sergiodxa:
@vasco3 if you have the query you used to get the data in useSWR you could run `mutate(GET_COMMENTS_QUERY)` and SWR will revalidate that query which means it will run it again, but you need to know the exact queries you want to run, to be more correct you need to know the keys useSWR received so that could be the query + variables in some cases.

vasco3:
@sergiodxa that would work. Thanks

HofmannZ:
Intresting! Going to play with this!

[...]

PR #1262, opened 2021-07-02
Hook for triggering validations manually (passive mode)

shuding:
## Motivation

There are a lot of existing discussions and feature requests about doing requests manually in SWR (e.g. #93, #116, #550, #1082), and we do see some valid use cases for it. In short, we need something that:

- Do not revalidate automatically (mount, focus, reconnect, retry...)
- Return a callback function for triggering the validation manually
- Backfill the request response to the cache*

*: _Potentially we can have a switch to turn on/off this step._

One option is to have a new hook, because the arguments would be slightly different from the `useSWR` hook. Since we now have better abstractions to build such a hook, namely `middlewares` and `revalidateWhenStale`, this PR is such an attempt.

## `useSWRTrigger`

```jsx
import useSWRTrigger from 'swr/trigger'

// fetcher impl.
function getData(url, token, opts) { ... }

// normal SWR API, but it will never start the request
const { data, error, mutate, trigger } = useSWRTrigger('/api/user', getData, { onError, onSuccess })

// trigger it manually, args will be passed to the fetcher (after the key arg/args)
trigger('myToken', myOptions)

// wait until it's done
await trigger('myToken', myOptions)
```

The `trigger()` call will still reflect the hook-returned `data` and `error` with a re-render, but that's more declarative.

## To-dos
- [x] Finalize the name of the hook.
- [ ] Sometimes, mutation APIs (POST, PUT, DELETE) return different payload from the GET API. It might be an issue if we write back to the cache with the same key. This behavior needs to be reconsidered.
- [ ] If the mutation APIs' results cannot be cached, that normally means that we need to revalidate those keys again if there's a `useSWR` hook with the same key. This can be another option to consider.
- [x] Ensure callbacks and `isValidating` state are working as expected.
- [x] Ensure with a test case that under trigger mode there's no deduping.
- [x] Ensure with a test case that `trigger` prevents race conditions with normal GET requests.
- [ ] (docs) Have a specific list of options, that can be used in `useSWRTrigger` and ignore the others.


trigger/index.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  // Method to trigger the request manually.
+  swr.trigger = useCallback(
+    (...extraArgs) => {
+      // If it's not ready to fetch, we throw an error.
+      if (!keyStr) throw new Error('Empty SWR key.')
+      if (!fetcher) throw new Error('Missing SWR fetcher.')
+
+      // Mutate the SWR data and return the result.
+      return swr.mutate(
```
giuseppeg, line 49: RE: 

> Sometimes, mutation APIs (POST, PUT, DELETE) return different payload from the GET API. It might be an issue if we write back to the cache with the same key. This behavior needs to be reconsidered.

When the user is not interested in updating the global cache for `key` they can use the `fetcher` right away i.e. they won't need `useSWRTrigger`.

trigger/index.ts
```diff
@@ -0,0 +1,88 @@
...
+  const [keyStr, args] = serialize(key)
+
+  // Method to trigger the request manually.
+  swr.trigger = useCallback(
+    (...extraArgs) => {
+      // If it's not ready to fetch, we throw an error.
+      if (!keyStr) throw new Error('Empty SWR key.')
+      if (!fetcher) throw new Error('Missing SWR fetcher.')
+
+      // Mutate the SWR data and return the result.
```
giuseppeg, line 48: What happens when a revalidation happens before the `useSWRTrigger` fetcher resolves?
Does SWR have mechanism to prevent race conditions?

shuding: Yeah, since we use `mutate` under the hood, race conditions will be prevented. In the case you described, the revalidation result will be ignored. 

All the 3 cases: https://github.com/vercel/swr/blob/f10e744e78bca9679c104dab7d04c39fb0f06766/src/use-swr.ts#L328-L339

trigger/index.ts
shuding: Yes. But I'm also thinking about an option to prevent updating the cache, but then `useSWRTrigger` will have nothing "SWR". Haven’t decided yet.

shuding:
This API needs more discussions, will keep it on hold for 1.0.

shuding:
Some thoughts:
- Passive hooks still need to have a key specified, even if it's not used by that request. Remote mutations are associated with normal GET requests, we need the key to avoid race conditions and do cache write back.
- Each passive hook needs to have its own states (data, error, isValidating), which won't be shared with useSWR or other passive hooks with the same key.

giuseppeg:
> Some thoughts:
> 
> * Passive hooks still need to have a key specified, even if it's not used by that request. Remote mutations are associated with normal GET requests, we need the key to avoid race conditions and do cache write back.
> * Each passive hook needs to have its own states (data, error, isValidating), which won't be shared with useSWR or other passive hooks with the same key.

The solution that I am currently using (https://twitter.com/giuseppegurgone/status/1427576808204034061) is close to this except that I don't use a key and revalidate but `mutate` programmatically. 

I think that triggering a revalidation (a GET request) when the remote mutation is complete is a good option although it could be redundant work, lead to higher consumption of resources since the mutation could already return the fresh data.

PR #1263, opened 2021-07-02
Subscription mode

huozhi:
To support subscription / disposable / observable data source, adding a new hook useSWRSubscription built on top of useSWR.

### API

```tsx
import type { Configuration } from 'swr'
import useSWRSupscription from 'swr/subscription'

useSWRSupscription(key, subscribe: SWRSubscription, config? Configuration)
```

```tsx
export type SWRSubscription<Data = any, Error = any> = (
  key: Key,
  { next }: { next: (err?: Error | null, data?: Data) => void }
) => () => void

export type SWRSubscriptionResponse<Data = any, Error = any> = {
  data?: Data
  error?: Error
}

export type SWRSubscriptionHook<Data = any, Error = any> = (
  key: Key,
  subscribe: SWRSubscription<Data, Error>,
  config?: SWRConfiguration
) => SWRSubscriptionResponse<Data, Error>
```

the purpose is to provider users enough flexibility to manage their own subscription/disposable on user land, but still keep the argument shape different from function type fetcher.

### Usage

```js
const subscribe = (key, { next }) {
  const unsubscribe = remoteSource.subscribe(
     (data) => next(null, data),
     (err) => next(err)
  )
  return () => unsubscribe()
}

const { data, error } = useSWRSubscription(key, subscribe)
```


jamesalester:
Is this likely to get merged in soon? Subscriptions would be really useful for us

huozhi:
We're trying to get it into 1.1 or 1.2. But shortly we'd love to collect more use cases of subscription pattern in real world apps so that this API can be well abstract. If there're feedback from you would be very appreciated

jamesalester:
Happy to help if I can. We're using subscriptions to have live data changes on multiple devices. We're using a graphql endpoint and currently have subscriptions set up like in [this project](https://github.com/praveenweb/swr-graphql). However, they don't handle the disposal of the websockets. I've been trying (and struggling) to implement my own disposal system using a similar hook to yours except it returns the dispose function with the data so that it can be called in the callback of a useEffect like: 

```
const { data, error, dispose } = useSWRSubscribe(key, subscribeFn)

useEffect(() => {
  return () => {
    dispose()
  }
}, [])
```

shuding:
Thanks! Yeah we still have a couple of concerns:
- Should we leverage the new React useSyncExternalStore (useMutableSource) API?
- In subscription mode, many of the SWR configs will be unusable (suspense, polling, error retry, ...), what should we do?
- Should we provide APIs for manually pausing/stopping subscriptions?

Overall I think we can try to launch this in version 1.1/1.2 as an unstable API, and polish it overtime.

jamesalester:
> Should we leverage the new React useSyncExternalStore (useMutableSource) API?

At first glance this looks interesting and could be useful. I've not used it before though so would need to read deeper into it to understand it.

> In subscription mode, many of the SWR configs will be unusable (suspense, polling, error retry, ...), what should we do?

I think the subscription mode would need its own set of config options. Some of those options wouldn't just be unusable but unnecessary too. I can't think of a reason why you would need to use polling with a subscription for example. For now it could just be a subset of the current SWR configs though it may need its own options adding in time.

> Should we provide APIs for manually pausing/stopping subscriptions?

I could see this being useful though probably not necessary for an initial release. 

I agree about launching this as an unstable API and polishing it over time.

avisra:
Is this still going to be delivered?

shuding:
Thank you for updating this! To push this forward, we need to set some minimal requirements to get this stable (under a flag or prefixed with unstable_). Here’re my thoughts:

- It needs to use the `useSyncExternalStore` API under the hood, to guarantee the stability and correctness when rendering the UI concurrently. (_This can be added in a future update_)
- Would be great if we change `(onData, onError) => dispose` to also support `(callback: (error, data) => void, ...keys) => dispose`, which follows the Node.js convenient, but also passes the current keys. This aligns with the `useSWR` hook.

subscribe/index.ts
```diff
@@ -0,0 +1,70 @@
...
+      }
+      return () => {
+        const count = subscriptions.get(key) || 1
+        subscriptions.set(key, count - 1)
+        // dispose if it's last one
+        if (count === 1) {
+          disposers.get(key)()
+        }
+      }
+    }, [key, mutate, subscribeFn])
```
shuding, line 63:
```suggestion
    }, [key])
```

`mutate` will never change and `subscribeFn` should be put into a ref like `useSWR`’s fetcher.

subscribe/index.ts
```diff
@@ -0,0 +1,70 @@
...
+        } catch (_) {
+          /* eslint-disable-line no-empty */
+        }
+      }
+
+      if (subscriptions.get(key) === 1) {
+        const dispose = subscribeFn(key, { onData, onError })
+        disposers.set(key, dispose)
+      }
+      return () => {
```
shuding, line 55: I’d wrap the cleanup logic with setTimeout(..., 0) to prevent unmount-mount scenarios caused by state updates. Think of 2 tabs with the same subscription hooks inside, but we only render the active one. If you switch the tab the subscription will be unsubscribed and then resubscribed. This is a common thing we can improve by delaying it til the next tick. 

Also, for strict effects it will be double invoked so this is a nice trick to avoid unnecessary network requests.

shuding (commented):
Looks like a good implementation for an initial unstable release! I think the ref counting approach should work with concurrent rendering, but let’s see..

subscription/index.ts
```diff
@@ -0,0 +1,88 @@
...
+const subscriptions = new Map<Key, number>()
+const disposers = new Map()
+
+export const subscription = (<Data, Error>(useSWRNext: SWRHook) =>
+  (
+    key: Key,
+    subscribe: SWRSubscription<Data, Error>,
+    config?: SWRConfiguration
+  ): SWRSubscriptionResponse<Data, Error> => {
+    const { data, error, mutate } = useSWRNext(key, null, config)
```
shuding, line 32: We should not deconstruct the object here, but use `swr = useSWRNext(...)` and only access then when necessary. You can take other hooks for reference.

subscription/index.ts
```diff
@@ -0,0 +1,88 @@
...
+        }
+      }
+
+      const callback = (_err?: any, _data?: Data) => {
+        if (_err) onError(_err)
+        else onData(_data)
+      }
+
+      if (subscriptions.get(key) === 1) {
+        const dispose = subscribeRef.current(key, callback)
```
shuding, line 64: Let's keep the second argument an object, so it can be extended in the future. It's also the reason that we made the breaking change in #1864.

```suggestion
        const dispose = subscribeRef.current(key, { callback })
```

subscription/index.ts
```diff
@@ -0,0 +1,103 @@
...
+            disposers.get(key)()
+          }
+        })
+      }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [key])
+
+    const getSnapshot = useCallback(() => disposers.size, [])
+
+    useSyncExternalStore(subscribeCallback, getSnapshot)
```
shuding, line 86: I think we can use useIsomorphicEffect instead of useSyncExternalStore here?

promer94: I agree, useIsomorphicEffect is enough.

huozhi: I thought we need to handle `disposers` as external state, no?

promer94: Since we don't use disposers for render. i think we dont need to sync the disposers size.

promer94 (approved):
🚀

isBatak:
What's the best way to handle conditional subscription?
For example key is undefined on mount.

```
const { data: token } = useToken(); // fetch token, async
const connection = useConnection(token); // connection will be undefined util token is fetched
const key= connection?.sid;
const { data: messages } = useSWR(key, () => connection.getMessages());

const subscribe = (key, { next }) => {
  if (!connection) return; // we probably don't need this early return
  
  const handler = (data) => next(undefined, data);
  const dispose = connection.on('messageAdd', handler);
  
  return () => {
      connection.off('messageAdd', handler);
  };
 }

useSWRSubscription(key, subscribe);
```
`key` is undefined on the first render, this will produce two entries in `disposers` Map.
Something like this `Map(2) {'' => ƒ, 'EXAMPLE_SID' => ƒ}`.
Is this intended behaviour?

subscription/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    const [key] = serialize(_key)
+    const swr = useSWRNext(key, null, config)
+    const subscribeRef = useRef(subscribe)
+
+    useIsomorphicLayoutEffect(() => {
+      subscribeRef.current = subscribe
+    })
+
+    useIsomorphicLayoutEffect(() => {
+      subscriptions.set(key, (subscriptions.get(key) || 0) + 1)
```
shuding, line 47:
```suggestion
      if (!key) return
      subscriptions.set(key, (subscriptions.get(key) || 0) + 1)
```

Necessary to return early if key is falsy?

isBatak: @shuding but will this line still create an unnecessary empty string key in cache?
https://github.com/vercel/swr/pull/1263/files#diff-87add53ebf9f9528edf5d95a0de40ea324d6d985abf5532ffb03363d58fa922cR39

feledori:
I like the idea of this and i am currently trying to use SWR as a cache for WebSocket connections. In my use case I want to stash all incoming messages and use them as a log. With this current API I can only update the data with the incoming data. It would be nice if there was access to the current saved data through the next function to update the current data with old data and keep the cache. I don't know if this should be a use case for this hook and just wanted to give feedback. Thanks for the work on this.

isBatak:
> I like the idea of this and i am currently trying to use SWR as a cache for WebSocket connections. In my use case I want to stash all incoming messages and use them as a log. With this current API I can only update the data with the incoming data. It would be nice if there was access to the current saved data through the next function to update the current data with old data and keep the cache. I don't know if this should be a use case for this hook and just wanted to give feedback. Thanks for the work on this.

@feledori you can always save some complex object in the cache instead of array, something like:
```
const { data, mutate } = useSWRImmutable(key, async () => {
  const messages = await connection.getMessages();
  return { messages, incoming:[] };
});

useSWRSubscription(key, (key, { next }) => {  
  const handler = (newMessage) => {     
     next(undefined, { 
        messages:  [newMessage, ...data.messages], 
        incoming: [newMessage, ...data.incoming] 
    });
  }
  connection.on('messageAdd', handler);
  
  return () => {
      connection.off('messageAdd', handler);
  };
 });
 
 console.log(data.messages);
 console.log(data.incoming);

```

mike-shtil-loop:
@huozhi hey! Any info on which milestone will include this, following **2.0**?

_internal/utils/mutate.ts
```diff
@@ -54,6 +54,8 @@ export async function internalMutate<Data>(
     typeof _opts === 'boolean' ? { revalidate: _opts } : _opts || {}
   )
 
+  console.log('opts', options)
+
```
shuding, line 58:
```suggestion
```

subscription/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    const [, set] = createCacheHelper<Data>(cache, key)
+
+    useIsomorphicLayoutEffect(() => {
+      subscriptions.set(key, (subscriptions.get(key) || 0) + 1)
+
+      const onData = (val?: Data) => swr.mutate(val, false)
+      const onError = async (err: any) => set({ error: err })
+
+      const next = (err_?: any, data_?: Data) => {
+        console.log('next:_err', err_)
```
shuding, line 54:
```suggestion
```

subscription/index.ts
```diff
@@ -0,0 +1,92 @@
...
+        if (err_) onError(err_)
```
shuding, line 55:
```suggestion
        if (err_ !== null && typeof err_ !== 'undefined') onError(err_)
```

subscription/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    const { cache } = config
+    const [, set] = createCacheHelper<Data>(cache, key)
+
+    useIsomorphicLayoutEffect(() => {
+      subscriptions.set(key, (subscriptions.get(key) || 0) + 1)
+
+      const onData = (val?: Data) => swr.mutate(val, false)
+      const onError = async (err: any) => set({ error: err })
+
+      const next = (err_?: any, data_?: Data) => {
```
shuding, line 53:
```suggestion
      const next = (err_?: Error | null, data_?: Data) => {
```

subscription/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    })
+
+    const { cache } = config
+    const [, set] = createCacheHelper<Data>(cache, key)
+
+    useIsomorphicLayoutEffect(() => {
+      subscriptions.set(key, (subscriptions.get(key) || 0) + 1)
+
+      const onData = (val?: Data) => swr.mutate(val, false)
+      const onError = async (err: any) => set({ error: err })
```
shuding, line 51:
```suggestion
      const onError = (err: any) => set({ error: err })
```

shuding (changes requested):

shuding (approved):
Let's go!

subscription/index.ts
```diff
@@ -85,13 +87,15 @@ export const subscription = (<Data, Error>(useSWRNext: SWRHook) =>
         // Prevent frequent unsubscribe caused by unmount
         setTimeout(() => {
           // TODO: Throw error during development if count is undefined.
-          const count = subscriptions.get(subscriptionKey)! - 1
+          const count = subscriptions.get(subscriptionKey)
+          if (count == null) return
```
shuding, line 91: This will never happen I believe?

huozhi: yeah for getting rid of the linting issue, but also was worrying about the setTimeout. 
And we have a comment: `// TODO: Throw error during development if count is undefined.`

PR #1266, opened 2021-07-04
Fix types generate

huozhi:
### Fix

types generation in subpath module dist files.

in `immutable/dist/immutable/index.d.ts`. but source files won't be included in the published npm package

before
```ts
import { Middleware, SWRHook } from '../src/types';
```

after, import from published module
```ts
import { Middleware, SWRHook } from 'swr';
```

### DIFF
--- a/immutable/index.ts
+++ b/immutable/index.ts
@@ -1,8 +1,5 @@
-// @ts-ignore
-import useSWR from 'swr'
-
+import useSWR, { Middleware, SWRHook } from 'swr'
 import { withMiddleware } from '../src/utils/with-middleware'
-import { Middleware, SWRHook } from '../src/types'
 
 export const immutable: Middleware = useSWRNext => (key, fetcher, config) => {
   // Always override all revalidate options.
--- a/infinite/index.ts
+++ b/infinite/index.ts
@@ -2,8 +2,6 @@
 // hook where `key` and return type are not like the normal `useSWR` types.
 
 import { useRef, useState, useCallback } from 'react'
-
-// @ts-ignore
 import useSWR from 'swr'
 
 import defaultConfig from '../src/utils/config'
@@ -20,7 +18,7 @@ import {
   SWRInfiniteResponse,
   MutatorCallback,
   Middleware
-} from '../src/types'
+} from 'swr'
 
 export const infinite = ((<Data, Error>(useSWRNext: SWRHook) => (
   getKey: KeyLoader<Data>,
--- a/infinite/index.ts
+++ b/infinite/index.ts
@@ -2,9 +2,8 @@
 // hook where `key` and return type are not like the normal `useSWR` types.
 
 import { useRef, useState, useCallback } from 'react'
-import useSWR from 'swr'
+import useSWR, { SWRConfig } from 'swr'
 
-import defaultConfig from '../src/utils/config'
 import { useIsomorphicLayoutEffect } from '../src/utils/env'
 import { serialize } from '../src/utils/serialize'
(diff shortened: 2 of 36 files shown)


jest.config.js
```diff
@@ -5,11 +5,12 @@ module.exports = {
   modulePathIgnorePatterns: ['<rootDir>/examples/'],
   setupFilesAfterEnv: ['<rootDir>/jest-setup.ts'],
   moduleNameMapper: {
-    '^swr$': '<rootDir>/src'
+    '^swr$': '<rootDir>/src',
+    '^@swr-internal/(.*)': '<rootDir>/src/$1'
```
shuding, line 9: Is `swr-internal` used somewhere?

infinite/index.ts
```diff
@@ -20,7 +18,7 @@ import {
   SWRInfiniteResponse,
   MutatorCallback,
   Middleware
-} from '../src/types'
+} from 'swr'
```
shuding, line 21: Even if we fixed those types, the dependency `import defaultConfig from '../src/utils/config'` will still import types directly from `'../src/types'`. So I feel this is a tricky problem.

huozhi: Good catch. Seems the types of whole swr core are emitted into package. Need to find a way to get rid of it.

huozhi: Only include the source files in subpath modules can avoid emitting swr types. But still fews left like `with-middlewares`. We can improve it later by later.

```
infinite/dist/
├── index.esm.js
├── index.js
├── infinite
│   └── index.d.ts
└── src
    └── utils
        ├── env.d.ts
        ├── hash.d.ts
        ├── helper.d.ts
        ├── normalize-args.d.ts
        ├── serialize.d.ts
        └── with-middleware.d.ts

3 directories, 9 files
```

infinite/index.ts
```diff
@@ -20,12 +17,12 @@ import {
...
 
 export const infinite = ((<Data, Error>(useSWRNext: SWRHook) => (
   getKey: KeyLoader<Data>,
   fn: Fetcher<Data> | null,
-  config: typeof defaultConfig & SWRInfiniteConfiguration<Data, Error>
+  config: typeof SWRConfig.default & SWRInfiniteConfiguration<Data, Error>
```
shuding, line 25: Nice one!

test/tsconfig.json
```diff
@@ -1,10 +1,11 @@
...
+  "extends": "../tsconfig.json",
   "compilerOptions": {
-    "rootDir": ".",
+    "rootDir": "..",
     "strict": false
   },
   "include": [
+    "..",
     "./test", 
     "./jest-setup.ts"
```
shuding, line 10: Maybe:

```suggestion
    "../test", 
    "../jest-setup.ts"
```

src/utils/serialize.ts
```diff
@@ -1,5 +1,5 @@
 import hash from './hash'
-import { Key } from '../types'
+import { Key } from 'swr'
```
shuding, line 2: Feels like a hack, it seems to be working fine though.

immutable/index.ts
```diff
@@ -1,8 +1,5 @@
-// @ts-ignore
-import useSWR from 'swr'
-
+import useSWR, { Middleware, SWRHook } from 'swr'
```
huozhi, line 1: only keep type imports in `immutable` and `infinite`, for reset of the generated types in `subpath/dist/src/xxx/x.d.ts` let's revisit them later

shuding (approved):

PR #1283, opened 2021-07-10
Refine createCache

### LINKED ISSUE #1275: createCache with revalidateOnFocus does not work
# Bug report

Custom cache settings in `1.0.0-beta.8` disables `revalidateOnFocus` behaviour; `focus` and `visibilitychange` events are not revalidating SWR caches.

## Expected Behavior
Revalidate on focus on `revalidateOnFocus: true`, as non-cache version does.

## Repro Steps / Code Example
CodePen: https://codepen.io/origin-master/pen/WNjwgYy

## Additional Context
SWR version: 1.0.0-beta.8

## This can be a problem...
`registerOnFocus` called in only here, and it's based on `defaultConfig.cache`:

https://github.com/vercel/swr/blob/c677d042c1307e00c641a2fbe8b559a1413670b5/src/use-swr.ts#L49-L62

Current API encourages only `cache` from `createCache(...)` should be put into `SWRConfig`, and doesn't cares `mutate`:
```js
const { cache, mutate } = createCache(new Map());
const Config = () => <SWRConfig value={{ cache }} />
```

huozhi:
Fixes #1275

### Changes

* Only need to `setupOn<Event>` for cache
* Setup global listeners inside `createCache` 

```js
createCache(new Map(), {
  // if you want to override any event that can trigger revalidation, use following options

  // override default behavior of revalidate on focus, it will revalidate on your own `focus` event now
  setupOnFocus(focusRevalidator) {
      customEvent.on('focus', focusRevalidator)
   },

   // override default behavior of revalidate on reconnect, it will revalidate on custom network events now
   setupOnReconnect(reconnectRevalidator) {
     customEvent.on('reconnect', reconnectRevalidator)
   }
})
```

### Fixes

* Fix custom cache is not bound to env like `focus`, `reconnect` events on web

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -49,13 +49,16 @@ export interface Configuration<
 
   isOnline: () => boolean
   isDocumentVisible: () => boolean
-  setupOnFocus: (cb: () => void) => () => void
-  setupOnReconnect: (cb: () => void) => () => void
 
   /**
    * @deprecated `revalidateOnMount` will be removed. Please considering using the `revalidateWhenStale` option.
    */
   revalidateOnMount?: boolean
+}
+
+export type Preset = {
+  setupOnFocus: (cb: () => void) => () => void
+  setupOnReconnect: (cb: () => void) => () => void
 }
 
 export type SWRHook = <Data = any, Error = any>(
--- a/src/types.ts
+++ b/src/types.ts
@@ -49,8 +49,8 @@ export interface Configuration<
 
   isOnline: () => boolean
   isDocumentVisible: () => boolean
-  registerOnFocus: (cb: () => void) => void
-  registerOnReconnect: (cb: () => void) => void
+  attachOnFocus: (cb: () => void) => () => void
+  attachOnReconnect: (cb: () => void) => () => void
 
   /**
    * @deprecated `revalidateOnMount` will be removed. Please considering using the `revalidateWhenStale` option.
--- a/src/types.ts
+++ b/src/types.ts
@@ -49,8 +49,8 @@ export interface Configuration<
 
   isOnline: () => boolean
   isDocumentVisible: () => boolean
-  attachOnFocus: (cb: () => void) => () => void
-  attachOnReconnect: (cb: () => void) => () => void
+  setupOnFocus: (cb: () => void) => () => void
+  setupOnReconnect: (cb: () => void) => () => void
 
   /**
    * @deprecated `revalidateOnMount` will be removed. Please considering using the `revalidateWhenStale` option.
--- a/src/types.ts
+++ b/src/types.ts
@@ -56,7 +56,7 @@ export interface Configuration<
   revalidateOnMount?: boolean
 }
 
-export type Preset = {
+export type Provider = {
   setupOnFocus: (cb: () => void) => () => void
   setupOnReconnect: (cb: () => void) => () => void
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -56,9 +56,9 @@ export interface Configuration<
   revalidateOnMount?: boolean
 }
 
-export type Provider = {
-  setupOnFocus: (cb: () => void) => () => void
-  setupOnReconnect: (cb: () => void) => () => void
+export type ProviderOptions = {
+  setupOnFocus: (cb: () => void) => void
+  setupOnReconnect: (cb: () => void) => void
 }
 
 export type SWRHook = <Data = any, Error = any>(
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -1,5 +1,6 @@
 import { useCallback, useRef, useDebugValue } from 'react'
 import defaultConfig from './utils/config'
+import { provider as defaultProvider } from './utils/web-preset'
 import { wrapCache } from './utils/cache'
 import { IS_SERVER, rAF, useIsomorphicLayoutEffect } from './utils/env'
 import { serialize } from './utils/serialize'
@@ -21,7 +22,7 @@ import {
   Cache,
   ScopedMutator,
   SWRHook,
-  Preset
+  Provider
 } from './types'
 
 type Revalidator = (...args: any[]) => void
@@ -46,12 +47,9 @@ const getGlobalState = (cache: Cache) => {
   ]
 }
 
-function setupGlobalEvents(cache: Cache, options: Partial<Preset> = {}) {
+function setupGlobalEvents(cache: Cache, _provider: Partial<Provider> = {}) {
   if (IS_SERVER) return
-
-  const setupOnFocus = options.setupOnFocus || defaultConfig.setupOnFocus
-  const setupOnReconnect =
-    options.setupOnReconnect || defaultConfig.setupOnReconnect
+  const provider = { ..._provider, ...defaultProvider } as Provider
   const [FOCUS_REVALIDATORS, RECONNECT_REVALIDATORS] = getGlobalState(cache)
   const revalidate = (revalidators: Record<string, Revalidator[]>) => {
     for (const key in revalidators) {
@@ -60,8 +58,8 @@ function setupGlobalEvents(cache: Cache, options: Partial<Preset> = {}) {
   }
   const onFocus = () => revalidate(FOCUS_REVALIDATORS)
   const onReconnect = () => revalidate(RECONNECT_REVALIDATORS)
-  setupOnFocus(onFocus)
-  setupOnReconnect(onReconnect)
+  provider.setupOnFocus(onFocus)
+  provider.setupOnReconnect(onReconnect)
 }
 
 // Setup DOM events listeners for `focus` and `reconnect` actions
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -20,7 +20,8 @@ import {
   SWRConfiguration,
   Cache,
   ScopedMutator,
-  SWRHook
+  SWRHook,
+  Preset
 } from './types'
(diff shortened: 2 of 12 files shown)


src/types.ts
```diff
@@ -49,8 +49,8 @@ export interface Configuration<
 
   isOnline: () => boolean
   isDocumentVisible: () => boolean
-  registerOnFocus: (cb: () => void) => void
-  registerOnReconnect: (cb: () => void) => void
+  attachOnFocus: (cb: () => void) => () => void
```
shuding, line 52: We can rename them to `setupOnFocus`, also more intuitive I guess?

src/use-swr.ts
```diff
@@ -488,6 +497,10 @@ export function useSWRHandler<Data = any, Error = any>(
     configRef.current = config
   })
 
+  useIsomorphicLayoutEffect(() => {
```
shuding, line 500: This will cause the listeners being added by every `useSWR` hook. I think we should add the setup/teardown logic in `SWRConfig`.

shuding (changes requested):

src/use-swr.ts
huozhi: yes but we're allowed user to pass `cache` per hook, that would also be a inconsistent

src/use-swr.ts
```diff
@@ -45,22 +45,31 @@ const getGlobalState = (cache: Cache) => {
...
-// Setup DOM events listeners for `focus` and `reconnect` actions
-if (!IS_SERVER) {
+function attachGlobalEvents(config: Configuration) {
   const [FOCUS_REVALIDATORS, RECONNECT_REVALIDATORS] = getGlobalState(
-    defaultConfig.cache
+    config.cache
   )
   const revalidate = (revalidators: Record<string, Revalidator[]>) => {
-    if (!defaultConfig.isDocumentVisible() || !defaultConfig.isOnline()) return
+    if (!config.isDocumentVisible() || !config.isOnline()) return
```
huozhi, line 53: If we don't have these 2 conditions, it will be easy to put `attachGlobalEvents` inside `createCache`...

test/use-swr-focus.test.tsx
```diff
@@ -200,4 +200,30 @@ describe('useSWR - focus', () => {
     await focusWindow()
     await screen.findByText('data: 5')
   })
+
+  it('should revalidate on focus even with custom cache', async () => {
+    let value = 0
+    const { cache } = createCache(new Map())
+
+    function Page() {
+      const { data } = useSWR('revalidaOnFocus + cache', () => value++, {
```
origin-master, line 209:
```suggestion
      const { data } = useSWR('revalidateOnFocus + cache', () => value++, {
```
typo here

src/use-swr.ts
```diff
@@ -45,22 +45,31 @@ const getGlobalState = (cache: Cache) => {
...
-  defaultConfig.registerOnFocus(() => revalidate(FOCUS_REVALIDATORS))
-  defaultConfig.registerOnReconnect(() => revalidate(RECONNECT_REVALIDATORS))
+  const onFocus = () => revalidate(FOCUS_REVALIDATORS)
+  const onReconnect = () => revalidate(RECONNECT_REVALIDATORS)
+  const dettachOnFoucs = config.attachOnFocus(onFocus)
+  const dettachOnReconnect = config.attachOnReconnect(onReconnect)
+
+  return () => {
+    dettachOnFoucs()
+    dettachOnReconnect()
```
origin-master, line 66:
```suggestion
  const detachOnFoucs = config.attachOnFocus(onFocus)
  const detachOnReconnect = config.attachOnReconnect(onReconnect)

  return () => {
    detachOnFoucs()
    detachOnReconnect()
```
typo here

src/use-swr.ts
shuding: Oh you're right... :sadparrot: then maybe it's not a good idea to allow per hook cache?

huozhi: @shuding I think we need to figure out here what we want to support for global events revalidation, there will be ways to achieve : )
I'll call `config.isDocumentVisible` and `config.isOnline` as `env cond` (environment condition) in the following sections.


We have these combinations
* default cache + default env  ✅
* default cache + custom env❓
* custom cache + default env ❓
* custom cache + custom env ❓

If a hook has customized **env cond**, the global one will still do the routine. Each cache just need to be setup once with default **env cond** or customized **env cond**.

Support default **env cond**: WeakSet to count if the custom cache is existed, setup for it only once;
Support customized **env cond**: This only occurs in `SWRConfig`, setup once for each `SWRConfig`;

IMO, ideally the **env cond** won't change frequently, WeakSet can be an option cause you may have few `SWRConfig` with different cache but same **env cond**

src/use-swr.ts
```diff
@@ -45,22 +45,31 @@ const getGlobalState = (cache: Cache) => {
...
 
     for (const key in revalidators) {
       if (revalidators[key][0]) revalidators[key][0]()
     }
   }
-  defaultConfig.registerOnFocus(() => revalidate(FOCUS_REVALIDATORS))
-  defaultConfig.registerOnReconnect(() => revalidate(RECONNECT_REVALIDATORS))
+  const onFocus = () => revalidate(FOCUS_REVALIDATORS)
+  const onReconnect = () => revalidate(RECONNECT_REVALIDATORS)
+  const teardownOnFoucs = config.setupOnFocus(onFocus)
```
origin-master, line 61:
```suggestion
  const teardownOnFocus = config.setupOnFocus(onFocus)
```
more typo

src/use-swr.ts
shuding: OK I think we should think about the most common use case: 99% apps only have 1 environment. So it's a good restriction that you shouldn't have per-hook, custom `isOnline` config, etc (they're safe to be singletons). 

Since we still haven't officially documented env settings, a good approach is to move those env settings into `createCache` (e.g. `createCache(provider, { isOnline, isDocumentVisible })`), so we can directly set up event listeners there too, before React's runtime.

src/utils/web-preset.ts
```diff
@@ -9,52 +10,66 @@ import { isUndefined } from './helper'
...
-    addDocumentEventListener('visibilitychange', cb)
-    addWindowEventListener('focus', cb)
+const setupOnFocus = (cb: () => void) => {
+  // focus revalidate
+  documentEventListener(add)('visibilitychange', cb)
+  windowEventListener(add)('focus', cb)
+  return () => {
+    documentEventListener(remove)('visibilitychange', cb)
+    windowEventListener(remove)('focus', cb)
   }
```
shuding, line 45: If we are setting up those listeners globally instead of doing it with a hook, there's no need to return unregisters I think.

src/utils/web-preset.ts
```diff
@@ -9,52 +10,66 @@ import { isUndefined } from './helper'
...
+  windowEventListener(add)('online', onOnline)
+  // nothing to revalidate, just update the status
+  windowEventListener(add)('offline', onOffline)
 
-    // nothing to revalidate, just update the status
-    addWindowEventListener('offline', () => (online = false))
+  return () => {
+    windowEventListener(remove)('online', onOnline)
+    windowEventListener(remove)('offline', onOffline)
   }
```
shuding, line 64: Same as above.

src/use-swr.ts
```diff
@@ -45,22 +47,23 @@ const getGlobalState = (cache: Cache) => {
   ]
 }
 
-// Setup DOM events listeners for `focus` and `reconnect` actions
-if (!IS_SERVER) {
-  const [FOCUS_REVALIDATORS, RECONNECT_REVALIDATORS] = getGlobalState(
-    defaultConfig.cache
-  )
+function setupGlobalEvents(cache: Cache, _provider: Partial<Provider> = {}) {
```
shuding, line 50: It seems that we are not using the second argument, what will the use case be link?

shuding:
I pushed a change due to one comment ☝️ , let me know what do you think!

huozhi:
force pushed it but honored your changes @shuding 
also add 2 test cases to cover the core changes

src/utils/web-preset.ts
```diff
@@ -9,52 +10,49 @@ import { isUndefined } from './helper'
...
-  typeof document !== 'undefined'
-    ? document.addEventListener.bind(document)
-    : null
+const hasWindow = typeof window !== 'undefined'
+const hasDocument = typeof document !== 'undefined'
+
+// For node and React Native, `add/removeEventListener` doesn't exist on window.
+const onWindowEvent =
+  hasWindow && !isUndefined(window[ADD]) ? window[ADD].bind(window) : noop
+const onDocumentEvent = hasDocument ? document[ADD].bind(document) : noop
```
shuding, line 22: I don't think we need to bind, since we are not using `this` in our callbacks. So we can make those simpler:

```js
const addWindowListener = typeof window !== 'undefined' && isUndefined(window[ADD]) ? window[ADD] : null

...

if (addWindowListener) addWindowListener('focus', cb)
```

Same for document listeners. So we can get rid of `noop` as well, make the bundle smaller.

huozhi: removed `bind` calls. but keep the others change since they make the final bundle reduced 25bytes 😌

shuding: I don't see your push somehow 🤔

huozhi: updated

shuding (approved):

PR #1303, opened 2021-07-20
Code structure refactoring

shuding:
The most significant change is to move event setup and global states into the cache wrapper, so the main logic (use-swr.ts) keeps uncoupled.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -1,7 +1,6 @@
 import { useCallback, useRef, useDebugValue } from 'react'
 import defaultConfig from './utils/config'
-import { provider as defaultProvider } from './utils/web-preset'
-import { wrapCache } from './utils/cache'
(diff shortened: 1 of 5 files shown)


src/use-swr.ts
```diff
@@ -72,7 +37,7 @@ const broadcastState: Broadcaster = (
   isValidating,
   shouldRevalidate = false
 ) => {
-  const [, , CACHE_REVALIDATORS] = getGlobalState(cache)
+  const [, , CACHE_REVALIDATORS] = SWRGlobalState.get(cache)!
```
huozhi, line 40: maybe use type gaurd or just disable this lint rule?

shuding: Adding some type guards would be great. 👍

huozhi (approved):

PR #1304, opened 2021-07-21
Breaking: drop the default fetcher

### LINKED ISSUE #1244: If a fetcher isn't defined, useSWR shouldn't fetch
# Bug report

In [Orchest](https://github.com/orchest/orchest), we recently implemented @pacocoursey's fantastic suggestion of [using SWR as a shared state hook](https://paco.sh/blog/shared-hook-state-with-swr).

The pattern works great, but has one caveat: `useSWR` still attempts to fetch the `key` value as an endpoint, even though no `fetcher` is defined.

## Expected Behavior

If a fetcher is `undefined` it should not attempt to fetch

## Repro Steps / Code Example

1. Go to https://codesandbox.io/s/swr-issue-repro-undefined-fetcher-uj8h6?file=/components/layout.jsx
2. Wait for the page to load
3. Open up the `Network` panel in your browser dev tools
4. Click the "Useless Toggle Button" (which toggles a shared `useUseless` state hook)
5. Notice how useSWR is attempting to fetch the `key` of `useless` and returning a 404

## Additional Context

SWR version - `0.5.6`

As a workaround, we're setting `fetcher` to `null`. However, this might trip up potential new users who wish to use the shared state pattern.

---

P.S. Thanks for your continued hard work on this library, it's incredible 🙏

huozhi:
Resolves #1244

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ const { data, error, isValidating, mutate } = useSWR(key, fetcher, options)
 #### Options
 
 - `suspense = false`: enable React Suspense mode [(details)](#suspense-mode)
-- `fetcher = window.fetch`: the default fetcher function
+- `fetcher`: the function to retrieve the remote data source
 - `initialData`: initial data to be returned (note: This is per-hook)
 - `revalidateOnMount`: enable or disable automatic revalidation when component is mounted (by default revalidation occurs on mount when initialData is not set, use this flag to force behavior)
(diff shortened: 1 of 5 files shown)


shuding (commented):
Let’s also add a test case that uses SWR to share local state?

shuding (approved):
Thank you!

PR #1307, opened 2021-07-22
Code refactoring

shuding:
Some tweaks to make the code more consistent and improve readability. Notable changes:
- no more `setTimeout` needed for `focusThrottleInterval`
- shared no-op function

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -488,15 +488,16 @@ export const useSWRHandler = <Data = any, Error = any>(
       configRef.current.isDocumentVisible() && configRef.current.isOnline()
 
     // Add event listeners.
-    let pending = false
+    let nextFocusRevalidatedAt = 0
     const onFocus = () => {
-      if (configRef.current.revalidateOnFocus && !pending && isActive()) {
-        pending = true
+      const now = Date.now()
+      if (
+        configRef.current.revalidateOnFocus &&
+        now > nextFocusRevalidatedAt &&
+        isActive()
+      ) {
+        nextFocusRevalidatedAt = now + configRef.current.focusThrottleInterval
         softRevalidate()
-        setTimeout(
-          () => (pending = false),
-          configRef.current.focusThrottleInterval
-        )
       }
     }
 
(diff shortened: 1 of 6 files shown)


src/use-swr.ts
```diff
@@ -457,45 +458,20 @@ export function useSWRHandler<Data = any, Error = any>(
...
     const onFocus = () => {
-      if (configRef.current.revalidateOnFocus && !pending && isActive()) {
-        pending = true
+      const now = Date.now()
+      if (
+        configRef.current.revalidateOnFocus &&
+        now > nextFocusRevalidatedAt &&
+        isActive()
+      ) {
+        nextFocusRevalidatedAt = now + configRef.current.focusThrottleInterval
```
huozhi, line 473: neat!

src/use-swr.ts
```diff
@@ -551,29 +553,31 @@ export function useSWRHandler<Data = any, Error = any>(
...
       if (
         !stateRef.current.error &&
         (refreshWhenHidden || config.isDocumentVisible()) &&
         (refreshWhenOffline || config.isOnline())
       ) {
-        // only revalidate when the page is visible
-        // if API request errored, we stop polling in this round
-        // and let the error retry function handle it
-        await revalidate({ dedupe: true })
+        revalidate({ dedupe: true }).then(_ => next())
```
huozhi, line 573: Lint fixing
```suggestion
        revalidate({ dedupe: true }).then(() => next())
```

src/use-swr.ts
```diff
@@ -551,29 +553,31 @@ export function useSWRHandler<Data = any, Error = any>(
...
     let timer: any = 0
 
-    function nextTick() {
-      if (refreshInterval) {
-        timer = setTimeout(tick, refreshInterval)
+    function next(force?: boolean) {
+      // We only start next interval if `refreshInterval` is not 0, and:
+      // - `force` is true, which is the start of polling
+      // - or `timer` is not 0, which means the effect wasn't canceled
+      if (refreshInterval && (timer || force)) {
```
huozhi, line 560: How about using `timer = -1` as initial value, then check `timer && timer !== -1` here, then we don't need `force` arg. Anyway it's not very blocking one

shuding: Good idea! I'll use `-1` to mark the polling as stopped, thanks!

huozhi (approved):

PR #1308, opened 2021-07-22
Rename isDocumentVisible to isVisible

huozhi:
To make the option meaning more general, and shorter for convenience, rename `isDocumentVisible` to `isVisible`


shuding (approved):

shuding:
Makes sense, also in the context of React Native there is not "document".

PR #1309, opened 2021-07-22
Reset cache between tests

huozhi:


test/use-swr-cache.test.tsx
```diff
@@ -201,4 +205,21 @@ describe('useSWR - cache', () => {
...
+        key,
+        () => {
+          return Array.from(provider.values()).length
+        },
+        { cache }
+      )
+      return <div>{data ? 'non-empty' : 'empty'}</div>
+    }
+    render(<Page />)
+    screen.getByText('empty')
```
shuding, line 223:
```suggestion
    expect(provider.size).toBe(0)
```

Is this better?

shuding (approved):

PR #1311, opened 2021-07-29
Fix config being ignored when fetcher is null

### LINKED ISSUE #1274: swr/infinite behaves weirdly without specifing fetcher function
# Bug report

```typescript
import useSWRInfinite from 'swr/infinite'

// Without specifying `fetcher`, `data` isn't an array of `Data`s; just an `Data`.
function useSWRInfinite(getKey): {data: Data, size: undefined, setSize: undefined}

// in contrast of specifying `fetcher`:
function useSWRInfinite(getKey, fetcher): {data: Data[], size: number, setSize: Function}
```

## Expected Behavior
Return same value between two cases, if \[specified `fetcher`\] and \[`fetcher` in `SWRConfig`\] is the same one.

## Repro Steps / Code Example
Here's a (hopefully) reproducible code example: https://codepen.io/origin-master/pen/qBmZRmM
```js
const List = ({ args = [], title = "" }) => {
	// when `args = [getKey]` => `Array.isArray(data) === false`
	// when `args = [getKey, fetcher]` => `Array.isArray(data) === true`
	const { data, setSize, size } = useSWRInfinite(...args);
```

## Additional Context
SWR version: `1.0.0-beta.8`

shuding:
When `args[1]` is `null`, `typeof args[1]` will still be `"object"` which causes this bug.

Fixes #1274.

### DIFF
--- a/src/utils/normalize-args.ts
+++ b/src/utils/normalize-args.ts
@@ -9,5 +9,5 @@ export function normalize<KeyType = Key, Data = any>(
 ): [KeyType, Fetcher<Data> | null, Partial<SWRConfiguration<Data>>] {
   return typeof args[1] === 'function'
     ? [args[0], args[1], args[2] || {}]
-    : [args[0], null, (typeof args[1] === 'object' ? args[1] : args[2]) || {}]
+    : [args[0], null, (args[1] === null ? args[2] : args[1]) || {}]
 }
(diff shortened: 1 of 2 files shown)


huozhi:
Maybe add a unit test for this helper?

huozhi (approved):

PR #1272, opened 2021-07-06
Move infinite types to swr/infinite and deprecate old ones

huozhi:
```diff
- import { SWRInfiniteConfiguration, SWRInfiniteResponse } from 'swr'
+ import { SWRInfiniteConfiguration, SWRInfiniteResponse } from 'swr/infinite'
```

PR #1326, opened 2021-08-03
chore: update useSWRInfinite import

eric-hc:
Per https://github.com/vercel/swr/pull/1272


swandogg:
quite the fantastic addition if I must say so myself! why indeed!

huozhi (approved):
Thanks!

Issue #1324, opened 2021-08-02
Issue implementing custom cache  in 1.0.0-beta.9


beetlebum:
I have implemented a custom cache following the examples in the documentation.

Given that the methods in the Cache interface take a string as a parameter, `get: (key: string)` for example, I assumed that the keys were being serialised before being passed into these methods.

I am seeing array keys coming through into the cache keys, causing all sorts of issues. 
Also, the `serialize` key method is not exported as far as I can tell.
It seems like function type keys are being converted to strings before arriving at the cache.

Am I missing something or is this a bug?
Should I stop using arrays as keys in the new version?

Thanks in advance.

huozhi:
Nice call @beetlebum , yet it's not an ideal solution to expose the serialization helper gracefully but we're working on it.
it could be either exposing the method directly or we give another parameter like raw keys to arrive cache along with string key and value.

shuding:
> I am seeing array keys coming through into the cache keys, causing all sorts of issues. Also, the serialize key method is not exported as far as I can tell.

May I ask what exactly the issues are? As a cache backend usually it doesn't need to be aware about the original cache keys (string or array of strings), it just need to provide a KV-like functionality. I'm curious about your scenario here, to improve the API we need to understand those use cases better. Thanks!

> It seems like function type keys are being converted to strings before arriving at the cache.

This is expected. The function itself isn't a key type for SWR, that function argument is used for returning an actual key.

shuding:
With v1, we've updated the [cache API](https://swr.vercel.app/docs/advanced/cache) interface, and added a new API to help you serialize keys:

```jsx
import { unstable_serialize } from 'swr'

unstable_serialize(key)
```

We might polish that API in the future, as we are thinking about supporting more options here.

PR #1337, opened 2021-08-12
Export `unstable_serialize` from SWR

shuding:
SWR handles key serialization so it can just pass a string to the cache provider. This makes all the interfaces simpler. However it makes getting data from the cache (or mutating a specific key) tricky, because the serialization util isn't exposed as an API.

Previously we had #1257 to expose the serialization of the `useSWRInfinite` key. This PR unifies the API name, as well as exposing the `serialize` API of `useSWR`.

Related: #1324, #1257.

### DIFF
--- a/infinite/index.ts
+++ b/infinite/index.ts
@@ -110,7 +110,7 @@ export const infinite = ((<Data, Error>(useSWRNext: SWRHook) => (
 
       let previousPageData = null
       for (let i = 0; i < pageSize; ++i) {
-        const [pageKey, pageArgs] = serialize(
+        const [pageKey, pageArgs] = serializeKey(
           getKey ? getKey(i, previousPageData) : null
         )
 
--- a/infinite/index.ts
+++ b/infinite/index.ts
@@ -11,7 +11,7 @@ import useSWR, {
   Middleware
 } from 'swr'
 import { useIsomorphicLayoutEffect } from '../src/utils/env'
-import { serialize } from '../src/utils/serialize'
+import { serialize as serializeKey } from '../src/utils/serialize'
 import { isUndefined, UNDEFINED } from '../src/utils/helper'
 import { withMiddleware } from '../src/utils/with-middleware'
 import { SWRInfiniteConfiguration, SWRInfiniteResponse } from './types'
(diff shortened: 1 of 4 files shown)


infinite/index.ts
```diff
@@ -11,18 +11,18 @@ import useSWR, {
...
 
 const INFINITE_PREFIX = '$inf$'
 
 const getFirstPageKey = (getKey: KeyLoader<any>) => {
-  return serialize(getKey ? getKey(0, null) : null)[0]
+  return serializeKey(getKey ? getKey(0, null) : null)[0]
 }
 
-export const getInfiniteKey = (getKey: KeyLoader<any>) => {
+export const serialize = (getKey: KeyLoader<any>) => {
```
huozhi, line 25: Do you think we should return an array here containing all the custom prefixed keys like `$ctx`, `$len` stuff?
That would be easier to access any of them

shuding: Not for now. I'm actually worried about exposing too many implementation details, we should do enough research about use cases (e.g. why do they need to access the infinite context?) before thinking about exposing those.

huozhi (approved):

infinite/index.ts
huozhi: I feel the different return type is also bit confusing for new uers (personally), but it's safe to carry on. We can revisit it later.

Would be nice if explicit return type is declared here

shuding: Oh that's a good point. We should probably only return the data key from `serialize`.

huozhi: unstable_serialize?

huozhi (approved):
LGTM

piotr-cz:
This is helpful, however there is no way to retrieve `errorKey` and `isValidatingKey` values.

I'm using [persistent storage](https://swr.vercel.app/docs/advanced/cache#localstorage-based-persistent-cache) for cache and once an item is removed, I'd like to remove it completely from persistent cache as well.

At this moment I'm using:

```js
export function deleteCacheEntry(cache: Cache, key: string): void {
  const keys = [
    key,
    '$err$' + key, // errorKey
    '$req$' + key, // isValidatingKey
  ]

  keys.forEach(key =>
    cache.delete(key)
  ))
}
```

One prefixes change, such code won't have any effect.

shuding:
I think one way to simplify all this is, using only 1 key for all states related to that resource and keep those states as object values:

```js
// old
cache.set(key, data)
cache.set(errorKey, error)
cache.set(validatingKey, isValidating)

// new
cache.set(key, { data, error, isValidating })
```

This will surely be a breaking change but it would be great to add it.

piotr-cz:
@shuding good idea

PR #1341, opened 2021-08-13
Improvement: ensure `useSWRInfinite` reuses `useSWR` cache

shuding:
Currently `useSWRInfinite` will always fetch new pages (when `setSize` is called), even if it's already cached by `useSWR` or something outside of that "pages" cache (an array of pages). With this PR we only re-fetch if the previous cached data of _that page_ has changed. 

Check the test case to see the detailed behavior.

### DIFF
--- a/test/use-swr-infinite.test.tsx
+++ b/test/use-swr-infinite.test.tsx
@@ -726,4 +726,31 @@ describe('useSWRInfinite', () => {
     await act(() => sleep(1))
     expect(renderedData).toEqual(['new'])
   })
+
+  it('should reuse cached value', async () => {
+    const key = createKey()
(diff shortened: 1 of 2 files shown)


test/use-swr-infinite.test.tsx
```diff
@@ -726,4 +726,31 @@ describe('useSWRInfinite', () => {
...
+    }
+
+    render(
+      <SWRConfig value={{ cache }}>
+        <Page />
+      </SWRConfig>
+    )
+
+    screen.getByText('data:')
+    await screen.findByText('data:response value')
```
huozhi, line 752:
```suggestion

```

`getByText` in next line can have the same effects?

huozhi (approved):

PR #1347, opened 2021-08-17
lint-staged: remove lint, fix prettier

huozhi:
* run type checking on `pre-push` git hook, to speed up local development
* fix prettier, lint command with lint-staged for changed files only


shuding:
I think we should still run linter and type checker on each commit locally, so we can fix issues immediately rather than waiting on the CI and then fix them with another commit.

huozhi:
Moved type-checking to pre-push since it requires context of all files, but eslint and prettier only requires changed files. So we can still keep them in pre-commit phase

PR #1349, opened 2021-08-18
fix: alias swr in next client bundle on dev mode

huozhi:
The swr in client bundle is not the one been aliased. reported by @promer94  

Add process to override swr assets by using `next.config.js`:

* run dev command
* create symblink for swr assets overriding
* quit dev command
* unlink custom `next.config.js` and kill dev server

### DIFF
--- a/scripts/dev-next.js
+++ b/scripts/dev-next.js
@@ -1,8 +1,8 @@
 #!/usr/bin/env node
 
-const { existsSync } = require('fs')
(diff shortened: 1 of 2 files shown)


shuding:
OK this is getting more complex 😄

huozhi:
> OK this is getting more complex 😄
yep not very ideal, might need to find some thing to replace these scripts for easier alias

PR #1354, opened 2021-08-19
Use trap to handle npm script multi tasks

huozhi:
nearly 40MB dev dep (concurrently) is erased


shuding:
I don't think `trap` is supported by Windows?

huozhi:
No it doesn't unless you use WSL or bash shell. but `rm` is also not supported...

PR #1358, opened 2021-08-20
upgrade bunchee, dedupe yarn lock

huozhi:
Around 80MB is deduped in node_modules. Now 242MB

Closes #1354 

* Upgrade bunchee for better yarn dedup
* Replace concurrently with npm-run-all (20MB reduced)
* Use `rimraf` instead of `rm -rf` for cross-platform compatible (32k)


shuding:
LGTM. `rimraf` might not be necessary since it's used by `prepublish` script and contributors won't need it usually. Will merge anyways.

PR #1360, opened 2021-08-20
Invert middleware execution order

shuding:
Middleware should be wrapped and executed from the outside to the inside.


shuding:
Will come up with a diagram in the docs to explain this carefully.

Issue #1345, opened 2021-08-14
Race condition where `data` or `error` return `1` when key is `null`


nandorojo:
# Bug report

## Description / Observed Behavior

On very rare occasions, if I return `null` in the key function of `useSWR`, my `data` and / or `error` field return `1`.

At first I thought it might be due to me calling `mutate` incorrectly somewhere. But it's happening for different calls (get user, get artists), so I no longer think that's the acse.

Plus, if the key returns `null`, shouldn't the `data` and `error` _always_ be undefined?

I did manage to wrap these calls and logged that `error` was `1` while `key` was `null`. But refreshing the page got rid of that bug. 

## Expected Behavior

Don't return `1` for `data` or `error` if the key returns `null`.

## Repro Steps / Code Example

I don't have a reproduction currently, I'm still working on that. I'm still trying to reproduce this in my own app.

So as it stands, I know that this issue might not be actionable. Hopefully this isn't a bug on the SWR side and I'll find it. However, I find it puzzling that SWR itself is returning these values.

Can any of the maintainers think of a potential reason this could be happening? Maybe an incorrect falsy check, or something like that, which results in SWR returning `1`?

## Additional Context

SWR version: `0.5.6`

This started happening after I upgraded from `0.3.2` to `0.5.6`, so I'm going to see if downgrading again solves it.

nandorojo:
I'm deep into debugging this, it's still hard for me to pinpoint.

This is the case I'm finding currently (simplified):

```js
const key = () => userId || null

const swr = useSWR(key, fetcher)
```

The `fetcher` function never gets called when there's no `userId`, which is correct.

However, this is the odd behavior: 

```js
const key = () => userId || null
const swr = useSWR(key, fetcher)

const cachedData = cache.get(key)

console.log({
  key: key(),            // null
  data: swr.data,    // 1
  error: swr.error   // 1
  cachedData        // 1
})
```

I'm not entirely sure how this is possible. I think it occurs when I am signed in and then sign out.

Basically, it seems that `cache.get(null)` is returning `1`, which is propagating to every request whose key is `null`. I'm not sure why `1` is stored for the `key = null`. The only thing I could think of is if I'm at some point storing that somewhere incorrectly. 

What's odd is, when I sign out, I call `cache.clear()`. So I'm not sure how any of my previously stored `mutate` calls would have caused this.

I'm going to set up a `cache.subscribe` so that I can see if I ever store `1` under the key `null`.

[...]

PR #1361, opened 2021-08-22
New useSWRConfig API and remove createCache

shuding:
This is an attempt to improve the UX of the new custom cache feature by addressing the following pains:
- [x] It's hard to know which `mutate` (the global one or the provider's one) to use inside a component's implementation.
- [x] It's tricky to initialize the cache only once when concurrent rendering (have to use `useState` lazy initialization). Users might put `createCache` inside render by accident.
- [x] A provider cannot access to its upper-level cache provider. This makes it impossible to implement multi-layer cache boundaries.
- [x] It's tricky to just provide some values as the fallback, instead of creating a new cache layer. This is useful for SSR/SSG cases.

### Create a new cache provider with SWRConfig
```js
return <SWRConfig value={{
  provider: () => new Map()
}}>
  <App/>
</SWRConfig>
```

### Get the current config, including cache and mutate, from useSWRConfig
```js
const { cache, mutate, ...extraConfig } = useSWRConfig()
```

_Fallback values and middlewares will be merged._

### Provide fallback values
```js
return <SWRConfig value={{ fallbackValues: { 'key': 'value' } }}>
  <App/>
</SWRConfig>
```

_This is somehow equivalent to the `initialData` option but in a provider level._

### Multiple cache layers
```js
return <SWRConfig value={{
  provider: cache => new Map()
}}>
  <App/>
</SWRConfig>
```

You can leverage the cache from the upper-level.

## Ref
- https://github.com/vercel/swr/issues/1345#issuecomment-899124645
- https://github.com/koba04/swr-devtools/issues/2


src/utils/cache.ts
```diff
@@ -34,24 +24,49 @@ function revalidateAllKeys(
...
+      get: originalProviderGet,
+      delete: provider.delete.bind(provider)
+    }
+  }
+  // If the initializer returns the parent provider with some fallback values,
+  // we need to override its `get` method.
+  if (fallbackValues) {
+    provider.get = (key: string) => {
+      const value = originalProviderGet(key)
+      if (isUndefined(value)) return fallbackValues[key]
```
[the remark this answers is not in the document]
shuding: Good catch! I'll make sure it doesn't modify the original provider object's get.

However we can't write to the cache, is is a side effect and hence this option is called "fallback", not "initial".

src/utils/provider.ts
```diff
@@ -0,0 +1,37 @@
+import { useContext, useState } from 'react'
+import { defaultConfig } from './config'
+import { wrapCache } from './cache'
+import { isFunction, UNDEFINED } from './helper'
+import { SWRConfigContext } from './config-context'
+import { SWRGlobalState, GlobalState } from './global-state'
+
+import { Cache, ProviderOptions } from '../types'
+
+export const useSWRProvider = <Data = any>(
```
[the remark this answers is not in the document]
shuding: Provider also implies providing cache or fallback values. Scope and boundary feel similar, and it's more conceptual.

src/utils/provider.ts
```diff
@@ -0,0 +1,37 @@
...
+  })
+
+  // Return the new provider if it's created, otherwise return the current
+  // provider.
+  return (
+    newProvider || {
+      cache,
+      mutate: (SWRGlobalState.get(cache) as GlobalState)[6]
+    }
+  )
```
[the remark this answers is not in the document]
shuding: I think when re-rendering this hook, `cache` won't be reassigned. That would cause a problem.

src/utils/provider.ts
```diff
@@ -0,0 +1,37 @@
...
+    | Partial<ProviderOptions>,
+  options?: Partial<ProviderOptions>
+) => {
+  const cache = useContext(SWRConfigContext).cache || defaultConfig.cache
+
+  // We use a state to keep the cache instance and make sure it's only
+  // initialized once.
+  const [newProvider] = useState(() => {
+    if (isFunction(initialize))
+      return wrapCache<Data>(initialize(cache), options)
```
huozhi, line 22: Maybe add a test case for leverage cache (`initialize(cache)`) from top context?

src/utils/provider.ts
```diff
@@ -0,0 +1,37 @@
...
+  initialize?:
+    | ((currentCache: Readonly<Cache>) => Cache)
+    | Partial<ProviderOptions>,
+  options?: Partial<ProviderOptions>
+) => {
+  const cache = useContext(SWRConfigContext).cache || defaultConfig.cache
+
+  // We use a state to keep the cache instance and make sure it's only
+  // initialized once.
+  const [newProvider] = useState(() => {
```
huozhi, line 20: This function has two responsibilities:
* create and return a scoped cache
* return a inherited cache

Consider separating them?

the creating cache part can be moved to `SWRConfig` and only keep `useSWRProvider()` for consuming purpose.

huozhi, line 20: Consider adding `cache` only to SWRConfig, and initialize/wrap cache inside SWRConfig.

```js
<SWRConfig value={{ cache: () => (new Map()) }}>
  <Page>
</SWRConfig>
```

For consuming side only use `{ mutate, cache } = useSWRProvider()`. then we could deprecate global `mutate`.

.eslintrc
```diff
@@ -43,6 +43,7 @@
     "@typescript-eslint/no-var-requires": 0,
     "@typescript-eslint/no-explicit-any": 0,
     "@typescript-eslint/explicit-module-boundary-types": 0,
+    "@typescript-eslint/ban-types": 0,
```
huozhi, line 46: [This rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md) is kind of useful, it could avoid using Function/object/Object those unsafe types/primitives.

`typeof f === 'function'` is very typing strong checking, maybe we just use `: boolean` for that helper

src/utils/provider.ts
shuding: It's possible to deprecate `mutate` in the future, but still early to say. The simplicity of the current API is great I feel if you don't customize the cache or use any concurrent rendering techniques.

shuding: > Consider adding cache only to SWRConfig, and initialize/wrap cache inside SWRConfig.

This is a good idea, I need to validate it.

.eslintrc
shuding: > maybe we just use : boolean for that helper

That doesn't work with type narrowing. For example:

```js
function (f: number | (() => void)) {
  if (isFunction(f)) {
    f()
  }
}
```

Will error if `isFunction` doesn't use the "is Function" assertion.

src/utils/provider.ts
shuding: The good part of this approach is that we can avoid the problem of using `useSWRProvider` to create new cache at a component level:

```js
function useUser() {
  const { cache } = useSWRProvider(() => new Map) // this should be avoided
  return useSWR('/api/user', { cache })
}
```

If we only allow creating the provider within `<SWRConfig>`, then it will be at the boundary level.

As a potential downside, it's tricky to pass `cache` to the SWR hook directly (_Is this still a valid case?_). Maybe we need `createCache` for it, but it is definitely not concurrent safe. I'd leave this use case for future versions.

test/use-swr-focus.test.tsx
```diff
@@ -200,9 +200,9 @@ describe('useSWR - focus', () => {
 
   it('should revalidate on focus even with custom cache', async () => {
     let value = 0
-    const { cache } = createCache(new Map())
 
     function Page() {
+      const { cache } = useSWRProvider(() => new Map())
```
huozhi, line 205:
```suggestion
      const { cache } = useSWRProvider()
```

src/utils/merge-config.ts
```diff
@@ -1,16 +1,23 @@
...
-  const m2 = b && b.middlewares
+  if (!b) return v
+
+  const { middlewares: m1, fallbackValues: f1 } = a
+  const { middlewares: m2, fallbackValues: f2 } = b
   if (m1 && m2) {
     v.middlewares = m1.concat(m2)
   }
+  if (f1 && f2) {
+    v.fallbackValues = mergeObjects(f1, f2)
```
huozhi, line 19: consider use `fallback` instead of `fallbackValues`? it's aligned with React naming. Will be easy for devs to understand

src/use-swr.ts
```diff
@@ -196,20 +74,23 @@ export const useSWRHandler = <Data = any, Error = any>(
 
   // Get the current state that SWR should return.
   const cachedData = cache.get(key)
-  const data = isUndefined(cachedData) ? initialData : cachedData
+  const fallbackData = isUndefined(initialData)
+    ? fallbackValues[key]
```
huozhi, line 78: I feel it would be wired if it's used as option for a hook. consider following

```js
useSWR(key, fn, {
   fallbackValues: { [key]: fallback }
})
```
This would look pretty duplicated for usage of key.

I feel the fallback should be different for hook and SWRConfig:
 
```js
<SWRConfig value={{ fallbacks: { [k1]: v1, [k2]: v2 } }}>
   <Page/ >
</SWRConfig>
```

```js
useSWR(key, fn, { fallback })
```

shuding: A use case is to provide some global mockup data:

```js
const mockupData = { '/api/user': ..., '/api/items': ... }

useSWR(key, fn, { fallbackValues: mockupData })
```

Or, we can rename `initialData` to `fallbackValue` but also keep `fallbackValues` :D

src/utils/merge-config.ts
shuding: Yeah I'm struggling with the naming here. `fallback` is great but it's singular, and `fallbacks`/`fallbackMap` feel weird...

pacocoursey: imo `fallback` doesn't have to imply singular, especially if the examples showed it was an object with multiple keys

src/use-swr.ts
```diff
@@ -196,20 +74,23 @@ export const useSWRHandler = <Data = any, Error = any>(
...
+    : initialData
```
pacocoursey, line 79: Are there plans to remove `initialData` entirely in v1? This seems like a good opportunity, replace it with `fallback`/`fallbackValues`

src/utils/cache.ts
```diff
@@ -31,37 +21,61 @@ function revalidateAllKeys(
...
-  const EVENT_REVALIDATORS = {}
-  const opts = { ...defaultProvider, ...options }
+  options?: Partial<ConfigOptions>
+): [Cache<Data>, ScopedMutator<Data>] | undefined {
+  // The global state for a specific provider will be used to deduplicate
+  // requests and store listeners. As well as a mutate function that bound to
+  // the cache.
+
+  // Provider's gloabl state might be already initialized. Let's try to get the
+  // global state assoicated with the provider first.
```
pacocoursey, line 33:
```suggestion
  // global state associated with the provider first.
```

src/utils/config-context.ts
```diff
@@ -1,16 +1,48 @@
...
 }> = ({ children, value }) => {
   // Extend parent context values and middlewares.
-  value = mergeConfig(useContext(SWRConfigContext) || {}, value)
-  return createElement(SWRConfigContext.Provider, { value }, children)
+  const extendedConfig = mergeConfigs(useContext(SWRConfigContext), value)
+
+  // Should not use the inherited provider.
+  const provider = value && value.provider
+
+  // Use a lazy initialized state to create the cache on first access.
```
pacocoursey, line 28: Might be unexpected, as a user, that the cache isn't initialized immediately

src/use-swr.ts
shuding: Yeah its use case is already covered completely by `fallbackValues`, but I'm thinking about deprecating it with `fallbackValue` for better fit:

```js
// we can put multiple values here, but it's longer
useSWR(key, fetcher, { fallbackValues: { [key]: 'value' } }) 

// same as initialData, but the name is clearer
useSWR(key, fetcher, { fallbackValue: 'value' })
```

Maybe we should deprecating `initialData` in v1, otherwise we'll have to wait for v2.

src/utils/config-context.ts
shuding: It will be initialized immediately when rendering the `SWRConfig` component. This is critical for React 18+ with concurrent rendering.

src/utils/merge-config.ts
shuding: Which name would you prefer as an API option (since you're a native speaker)?

My original design was:
- `fallbackValues` for a object with multiple keys
- `fallbackValue` for a single value (same as initialData)

If we choose `fallback`, we can have both of these^.

src/utils/config-context.ts
pacocoursey: Oh I misunderstood the comment here 👍 

> create the cache on first access

src/utils/config.ts
```diff
@@ -29,40 +34,45 @@ function onErrorRetry(
...
 }
 
-// Default config
-const defaultConfig: Configuration = {
-  // events
-  onLoadingSlow: noop,
-  onSuccess: noop,
-  onError: noop,
-  onErrorRetry,
+export const defaultProvider = initCache(new Map())!
```
huozhi, line 44: fix `Forbidden non-null assertion` lint warning

src/utils/config-context.ts
```diff
@@ -1,16 +1,48 @@
...
-  value: SWRConfiguration
+  value?: SWRConfiguration &
+    Partial<ConfigOptions> & {
+      provider?: (cache: Readonly<Cache>) => Cache
+    }
 }> = ({ children, value }) => {
   // Extend parent context values and middlewares.
-  value = mergeConfig(useContext(SWRConfigContext) || {}, value)
-  return createElement(SWRConfigContext.Provider, { value }, children)
+  const extendedConfig = mergeConfigs(useContext(SWRConfigContext), value)
```
huozhi, line 23:
```suggestion
  const [cache, mutate] = defaultProvider
  const extendedConfig = mergeConfigs({ cache, mutate }, useContext(SWRConfigContext), value)
```

src/utils/config-context.ts
```diff
@@ -1,16 +1,48 @@
...
+
+  // Should not use the inherited provider.
+  const provider = value && value.provider
+
+  // Use a lazy initialized state to create the cache on first access.
+  const [cacheAndMutate] = useState(() =>
+    provider
+      ? initCache(provider(extendedConfig.cache || defaultProvider[0]), value)
+      : UNDEFINED
```
huozhi, line 32:
```suggestion
    provider
      ? initCache(provider(extendedConfig.cache), value)
      : defaultProvider
```

src/utils/config-context.ts
```diff
@@ -1,16 +1,48 @@
...
+  )
+
+  // Override the cache if a new provider is given.
+  if (cacheAndMutate) {
+    extendedConfig.cache = cacheAndMutate[0]
+    extendedConfig.mutate = cacheAndMutate[1]
+  }
```
huozhi, line 39:
```suggestion
  extendedConfig.cache = cacheAndMutate[0]
  extendedConfig.mutate = cacheAndMutate[1]
```

huozhi (approved):
lgtm, let's tackle the naming of `fallbackValues` stuff in next PRs

PR #1368, opened 2021-08-24
Rename middlewares option to use

huozhi:
middleware can be singular or plural. rename it to `use` as short and clear. (Aother popular library like expressjs choosed `router.use()` api)


shuding:
Thank you!

PR #1370, opened 2021-08-24
Rename fallbackValues and initialData

shuding:
_Note: we need to update README (#656) and docs_.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -24,10 +24,10 @@ export interface PublicConfiguration<
   revalidateWhenStale: boolean
   shouldRetryOnError: boolean
   suspense?: boolean
-  initialData?: Data
+  fallbackData?: Data
   fetcher?: Fn
   use?: Middleware[]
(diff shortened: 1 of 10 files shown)


src/types.ts
```diff
@@ -24,10 +24,10 @@ export interface PublicConfiguration<
   revalidateWhenStale: boolean
   shouldRetryOnError: boolean
   suspense?: boolean
-  initialData?: Data
+  fallbackData?: Data
   fetcher?: Fn
   use?: Middleware[]
-  fallbackValues: { [key: string]: any }
+  fallback: { [key: string]: any }
```
huozhi, line 30:
```suggestion
  fallback?: { [key: string]: any }
```

shuding: It is always provided in the default config (as `{}`), so marked as always existing here to make hooks use it confidently. For public config, we have `Partial<>` so they're all optional.

huozhi (approved):

PR #1377, opened 2021-08-24
fix: cache is missing on server side

### LINKED ISSUE #1376: immutable (beta): SWRGlobalState.get(cache) is undefined when using SSG
# Bug report

I encounter this error on build when using useSWRImmutable on a Server-Side Generated page (getStaticProps)

```typescript
const { data } = useSWRImmutable(key, fetcher)
```

> `TypeError: Cannot read property '0' of undefined`

This code seems to be the cause:

https://github.com/vercel/swr/blob/master/src/use-swr.ts#L46

```typescript
const [
  EVENT_REVALIDATORS,
  STATE_UPDATERS,
  MUTATION_TS,
  MUTATION_END_TS,
  CONCURRENT_PROMISES,
  CONCURRENT_PROMISES_TS
] = SWRGlobalState.get(cache) as GlobalState
```

`SWRGlobalState.get(cache)` seems to be undefined in this case

## Additional Context

swr@1.0.0-beta.13

huozhi:
Fixes #1376

### DIFF
--- a/src/utils/cache.ts
+++ b/src/utils/cache.ts
@@ -41,20 +41,20 @@ export function initCache<Data = any>(
       Data
     >
 
+    // Update the state if it's new, or the provider has been extended.
+    SWRGlobalState.set(provider, [
+      EVENT_REVALIDATORS,
+      {},
+      {},
+      {},
+      {},
+      {},
+      mutate
+    ])
+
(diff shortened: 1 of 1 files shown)


shuding:
Thanks!

PR #1379, opened 2021-08-24
[Infinite] Return current data if failed to resolve pages from cache when setting size

### LINKED ISSUE #1362: infinite (beta): data becomes empty when using setSize and initialData
# Bug report

useSWRInfinite behaves weirdly when there is some initialData and when I use setSize()
- it refetches the first page
- it sets data to an empty array

## Expected Behavior

It should not

## Repro Steps / Code Example

Here is my code: 

```typescript
function getKey(page: number, previous: Data | null){
  console.log("getKey", "page", page, "previous", previous)
  console.log("url", "/api/data?page=" + page)
  return "/api/data?page=" + page
}

const pages = useSWR<Data>(getKey,
  url => {
    console.log("fetching", url)
    return fetcher(url) 
  }, 
  { initialData: props.pages })

console.log("data", pages.data, "size", pages.size)

function loadMore(){
  console.log("loading more")
  pages.setSize(size => size + 1)
}

return <>
  ...
  <button onClick={loadMore}>
    load more
  </button>
</>
```

(props.pages is given by Next.js SSR and never changes between renders)

I get the following logs:

```
getKey page 0 previous null
url /api/data?page=0
data [{...}] size 1
```
... then after I click "load more"
```
loading more
getKey page 0 previous null
url /api/data?page=0
getKey page 0 previous null
url /api/data?page=0
fetching /api/data?page=0      <---------- this is the first issue
getKey page 0 previous null
url /api/data?page=0
data [] size 2                 <---------- this is the second issue
getKey page 1 previous {...}
url /api/data?page=1
fetching /api/data?page=1
getKey page 0 previous null
url /api/data?page=0
data [{…}, {…}] size 2
```

As you can see, data becomes an empty array for a short moment
This is a problem because the UI doesn't show anything within that moment

## Additional Context

I use swr@1.0.0-beta.12

shuding:
Fix #1362.

### DIFF
--- a/infinite/index.ts
+++ b/infinite/index.ts
@@ -162,7 +162,7 @@ export const infinite = ((<Data, Error>(useSWRNext: SWRHook) => (
 
   const mutate = useCallback(
     (
-      data: Data[] | Promise<Data[]> | MutatorCallback<Data[]>,
+      data: Data[] | undefined | Promise<Data[]> | MutatorCallback<Data[]>,
       shouldRevalidate = true
     ) => {
       // It is possible that the key is still falsy.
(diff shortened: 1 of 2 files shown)


huozhi (approved):

huozhi:
Awesome, this is the right fix! Tested with the sandbox link provided

PR #1381, opened 2021-08-25
Rename revalidateWhenStale to revalidateIfStale

shuding:
Both have the same meaning, but "if" is close to the HTTP cache control naming conventions ("only-if-cached", "stale-if-error").


huozhi (approved):
LGTM! "If" sounds better to me since "when" is more implying a moment

PR #1389, opened 2021-08-27
chore: remove isomorphic-unfetch from examples

koba04:
I've noticed that Next.js started providing a built-in polyfill of `fetch` since v9.1.7, so `isomorphic-unfetch` is no longer required.
https://nextjs.org/blog/next-9-1-7#new-built-in-polyfills-fetch-url-and-objectassign


huozhi (approved):
Good to see examples become simpler, nice work!

PR #1396, opened 2021-08-28
fix: isValidating when an error occurred

### LINKED ISSUE #1395: isValidating is still true on error
# Bug report

isValidating is still true even if the request failed

Capture

## Expected Behavior

I expect isValidating to be false if the request ended, even if it failed

## Repro Steps / Code Example

```typescript
const res = useSWR("/api/hello", 
() => { throw new Error() })

console.log(res)
```

## Additional Context

swr@1.0.0

koba04:
Fixes #1395.
I guess SWR should update the cache value for `isValidating` when an error occurred.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -254,6 +254,7 @@ export const useSWRHandler = <Data = any, Error = any>(
         }
       } catch (err) {
         cleanupState()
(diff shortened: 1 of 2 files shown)


shuding:
Good catch, thank you!

koba04:
Thank you!

PR #1397, opened 2021-08-28
Chore: always use config ref

huozhi:
* use config ref in `execute` function
* shorten generated code (8016 -> 8582)


shuding:
This seems like a good idea! Unrelated: I think we should make clear about when to use the latest config and when use let snapshot version.

huozhi:
Like use ref for function calls and config for some primitive values?

shuding:
I meant when to use `configRef.current.a` and when to use `config.a`. Most of the options should be good to use the latest value, but for some we need to use the snapshot for correctness.

PR #1398, opened 2021-08-28
ci: setup size bot

huozhi:


huozhi:
tried few ways to write up the glob pattern but no luck to hide LICENSE file.

koba04:
Great👍 This is very useful! 
The script is failing in the build of pushing the master branch.
https://github.com/vercel/swr/runs/3451484703

I'm not sure the reason, but does the script has to run on the build of pushing the master branch?
It seems to be better off separating as another GitHub action because it's not for testing and linting.

PR #1402, opened 2021-08-29
feat: add a warning for invalid arguments with suspense mode

### LINKED ISSUE #1359: SWR should throw an error when suspense is enabled but key is not ready
Related to #339. When the `suspense` option is enabled, SWR can not return any value to continue rendering, neither throwing any promise if it can't fetch. This includes 2 situations:
1. `suspense: true` and `key` is falsy.
2. `suspense: true` and `fetcher` is `null`.

In both cases SWR should throw an error.

koba04:
Fixes #1359.
I've added a warning for invalid arguments with suspense mode.
The check (`!key || !fn`) is the same validation used in other places.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -78,6 +78,10 @@ export const useSWRHandler = <Data = any, Error = any>(
   const data = isUndefined(cached) ? fallback : cached
   const error = cache.get(keyErr)
 
+  if (suspense && (!key || !fn)) {
+    throw new Error('useSWR requires either key or fetcher with suspense mode')
+  }
+
   // A revalidation must be triggered when mounted if:
   // - `revalidateOnMount` is explicitly set to `true`.
   // - Suspense mode and there's stale data for the initial render.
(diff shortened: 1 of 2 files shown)


shuding:
Thanks! 🎉

shuding:
Gonna merge this for now, we can fix the CI later.

koba04:
Thank you!

karol-majewski:
With that change, how are we supposed to fetch conditionally with suspense?

According to [the docs](https://swr.vercel.app/docs/suspense#note-with-conditional-fetching), a falsy value should be used when SWR is not supposed to make a call. I did exactly that, and now in v1.0.1 my app doesn't work anymore.

koba04:
@karol-majewski Thank you for your feedback, I missed the document and that makes sense for me. I think this is a breaking change with the document and I'll create a PR to revert this.
There are some discussions about how `suspense` mode behaves with nullish values for `key` and `fetcher`, so we should continue discussing about this.

karol-majewski:
@koba04 Thank you for such a quick reaction! 🙇

PR #1403, opened 2021-08-29
ci: separate size bot job

huozhi:
Previously running report together with test job for speed up, but seems pkg-size action doesn't work on "push to master" phase. So separate it for safety and independence

### DIFF
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -1,0 +1,27 @@
+name: Report
+
+on:
+  pull_request:
+
+jobs:
+  pkg-size-report:
+    name: Package Size
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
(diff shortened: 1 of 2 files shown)


shuding:
I’m not sure about the cause here, is it gonna be fixed with this PR?

> but seems pkg-size action doesn't work on "push to master" phase

huozhi:
That's sth expected but the error trace is not very clear. pkg-size action use `generateSizeReport({ pr, ...})` to create size report based on PR ([details](https://github.com/pkg-size/action/blob/cc5eef94b3703031cfc048ee8acbcf9bfbd35510/src/index.js#L18-L19)). A "push" doesn't have `base` and `head` at the same time like a PR does.

koba04:
Currently, the pkg-size action seems not to be able to post a comment on pull requests from forked repositories.

> Resource not accessible by integration

https://github.com/vercel/swr/pull/1404/checks?check_run_id=3454711722

huozhi:
Oops didn't notice the authentication error.. Let's [revert size bot](https://github.com/vercel/swr/pull/1405) for now, I'll try to work from my forked repo later on this

PR #1421, opened 2021-09-01
test: add tests for a global cache

koba04:
I've refactored tests for errors at #1404 to use a custom cache for each test, and will refactor other tests as well. Before doing that, I've added tests for the global cache not to break the behaviors of the global cache.

### DIFF
--- a/test/use-swr-cache.test.tsx
+++ b/test/use-swr-cache.test.tsx
@@ -415,11 +415,7 @@ describe('useSWR - global cache', () => {
       return <div onClick={() => setIndex(1)}>{data}</div>
     }
 
-    function App() {
-      return <Section />
-    }
-
-    const { container } = render(<App />)
+    const { container } = render(<Section />)
     await screen.findByText(fetcher(keys[0]))
 
     expect(cache.get(keys[1])).toBe(undefined)
@@ -441,11 +437,7 @@ describe('useSWR - global cache', () => {
       return <div>data:{data}</div>
     }
 
-    function App() {
-      return <Page />
-    }
-
-    render(<App />)
+    render(<Page />)
     screen.getByText('data:')
     await act(() => mutate(key, 'mutated value', false))
     await screen.findByText('data:mutated value')
@@ -460,10 +452,7 @@ describe('useSWR - global cache', () => {
       })
       return <>{String(data)}</>
     }
-    function Page() {
-      return <Foo />
-    }
-    render(<Page />)
+    render(<Foo />)
     screen.getByText('undefined')
 
     await screen.findByText('0')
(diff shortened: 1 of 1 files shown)


test/use-swr-cache.test.tsx
```diff
@@ -426,3 +386,115 @@ describe('useSWR - cache provider', () => {
...
+      mutate = mutateWithCache
+      const { data } = useSWR(key, null)
+      return <div>data:{data}</div>
+    }
+
+    function App() {
+      return <Page />
+    }
+
+    render(<App />)
```
huozhi, line 448:
```suggestion
    render(<Page />)
```

test/use-swr-cache.test.tsx
```diff
@@ -426,3 +386,115 @@ describe('useSWR - cache provider', () => {
...
+      const { data } = useSWR(keys[index], fetcher)
+
+      return <div onClick={() => setIndex(1)}>{data}</div>
+    }
+
+    function App() {
+      return <Section />
+    }
+
+    const { container } = render(<App />)
```
huozhi, line 422:
```suggestion
    const { container } = render(<Section />)
```

test/use-swr-cache.test.tsx
```diff
@@ -426,3 +386,115 @@ describe('useSWR - cache provider', () => {
...
+    function Foo() {
+      const { data } = useSWR(key, () => value++, {
+        dedupingInterval: 0
+      })
+      return <>{String(data)}</>
+    }
+    function Page() {
+      return <Foo />
+    }
+    render(<Page />)
```
huozhi, line 466: Or just rename Foo -> Page

```suggestion
    render(<Foo />)
```

huozhi (approved):

huozhi:
Thanks!

huozhi:
@koba04 How you feel about using jest `setupFiles` to reset cache in jest setup configuration? IMO that would be a shorter way to clean cache for each test case if it's possible

koba04:
Thank you!

@huozhi 
Yes, you are right. resetting a global cache in jest's setup file is the easiest way to solve problems caused by sharing a global cache.
Currently, the `Cache` interface doesn't have the `clear` method, so we would have to add it to the interface to do that. 

Personally, I prefer to create a custom cache for each test even though it requires more works because I'd like to avoid using any shared store between tests and SWR now provides the way. As the blog post for v1 mentioned, this is a valid use case of the custom cache feature and that makes tests more reliable than depending on a global cache.
But I don't have a strong opinion for that, so I'm ok that maintainers choose to keep using a global cache.

huozhi:
@koba04 thanks! It was just an idea comming to my head to save efforts, not 100% sure if it's easy and possible. Feel free to choose they way you think make much more sense! 👍

koba04:
@huozhi Thank you for your feedback!

PR #1430, opened 2021-09-05
Fix revalidateIfStale behavior when the key changes

shuding:
With the added test, currently 4 requests will be made because a SWR hook will always revalidate when the key changes. However the correct behavior should be treating a key change as a "remount", and using the `shouldRevalidateOnMount` condition to decide if a revalidation is necessary.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -400,7 +400,7 @@ export const useSWRHandler = <Data = any, Error = any>(
     }
 
     // Trigger a revalidation.
-    if (keyChanged || shouldRevalidateOnMount()) {
+    if (shouldRevalidateOnMount()) {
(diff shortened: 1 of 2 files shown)


shuding:
This bug was originally reported with https://codesandbox.io/embed/happy-nova-kzbzg?file=/src/App.js&codemirror=1.

huozhi (approved):

whj1995:
its breaking change. key has change, never revalidate.

shuding:
@whj1995 Are you using `revalidateIfStale: false`? Can you share an example?

whj1995:
> @whj1995 Are you using `revalidateIfStale: false`? Can you share an example?

https://codesandbox.io/s/goofy-silence-4nqf7?file=/src/App.js

PR #1431, opened 2021-09-06
Feature: make return state(data, error and isValidating) readonly

### LINKED ISSUE #1386: Make returned state (data, error, isValidating) readonly
We can use mark the properties of the SWR return state readonly, to avoid modifying them. For example it's dangerous to directly modify the `data` object, which could cause an infinite loop of rerender.

anirudh1713:
closes #1386 

make return state `Readonly`


shuding:
Thank you!

huozhi:
We might need to revert this since this effect all data/error usage across swr source and test files. Currently the type checking didn't capture the typing error thrown from it

PR #1440, opened 2021-09-09
Fix initial isValidating when isPaused() returns true

anothertempore:
This is a bugfix. When config `isPaused()` to returns `true`, the `isValidating` supposed to be `false`, but currently it is `true`.

Can refer to this example to reproduce the bug: https://codesandbox.io/s/elastic-cdn-pufkd?file=/src/App.js


shuding (approved):
Thanks!

PR #1446, opened 2021-09-10
build: no emit on error

huozhi:
if there's error bunchee should stop building

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -275,7 +275,7 @@ export const useSWRHandler = <Data = any, Error = any>(
           // Keep the stale data but update error.
           setState({
             isValidating: false,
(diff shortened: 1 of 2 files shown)


src/use-swr.ts
```diff
@@ -275,7 +275,7 @@ export const useSWRHandler = <Data = any, Error = any>(
           // Keep the stale data but update error.
           setState({
             isValidating: false,
-            error: err
+            error: err as Error
```
huozhi, line 278: type guard for error

image

not sure why I'm having this...since we're still not ts v4. Can also be disabled by `"useUnknownInCatchVariables": false` when upgrading to v4

shuding (approved):

Issue #157, opened 2019-11-27
Improving immediate mutate + later revalidation


Daiz:
The documentation for `mutate` talks about local mutation for faster feedback, but the documented way to use it only mutates after a promise has resolved - in other words, *not immediately*.

As a result, I've found myself writing code in this kind of pattern:

```typescript
mutate(path, { ...data, patch }, false); // mutate immediately, don't revalidate
await patchData(patch); // await the actual API call
trigger(path); // trigger revalidation afterwards
```

Technically, if `patchData` also returns the full new data, you could also do it like this:

```typescript
mutate(path, { ...data, patch }, false); // mutate immediately locally, don't revalidate
mutate(path, patchData(patch), false); // update with API-backed result from patchData
```

Which is actually also present in the documentation. But it is kinda janky to need two subsequent mutate calls for this...

One backwards-compatible way to deal with this could be to extend `mutate` to accept options as the third parameter, and introduce `initialData` much like present in `useSWR`. Then you could do something like this:

```typescript
// mutate immediately with `initialData`,
// then replace with result from patchData when it resolves
mutate(path, patchData(patch), { initialData: { ...data, patch }); 
```

In case the patchData function (or equivalent) *doesn't* return the full data, the third argument could also be augmented to accept a `Promise`, eg.

```typescript
// wait until the Promise from patchData resolves before triggering revalidate
mutate(path, { ...data, patch }, patchData(patch))
```

I feel both these changes would make using mutate a lot more ergonomic overall.

In addition, I haven't checked the code to see how easy this would be to implemented, but another nice possibility would be to keep the local data (ie. non-promise second argument if third arg is a promise, anything in `initialData`) provided to `mutate` in some sort of "unverified cache", such that in case the provided promises throw, SWR could automatically roll the data back to "last known good value" (ie. the last real API response for that path).

johot:
I also find this very odd. Like you say the whole idea seems to be to mutate directly, then revalidate at a later time to make sure everything wen't like it should.

If I have a slow API and use mutate like in the documentation samples, I still get a big delay in my UI unlike doing like @Daiz.

Any reason the samples look like they do?

shuding:
This is a nice-to-have feature! I personally like the second API more (the 3rd argument can be a function or a promise):

```js
// wait until the Promise from patchData resolves before triggering revalidate
mutate(path, { ...data, patch }, patchData(patch))

// async function
mutate(path, { ...data, patch }, async () => { return patchedData })

// no revalidation
[...]

PR #1450, opened 2021-09-12
feat: Hook for remote mutations

shuding:
Closes #1262.

## Motivation

There are a lot of existing discussions and feature requests about doing requests with side effects on the remote side manually in SWR as well as a better mutation API design (e.g. #93, #116, #157, #550, #1082), and we do see some valid use cases for it.

As a conclusion, we need something that:

- Stops revalidate automatically (mount, focus, reconnect, retry...)
- Returns a callback function for triggering the mutation manually
- Prevents potential race conditions
- Revalidates the resource after mutation
- Backfills the return value to the cache
- Deriving: Backfills the cache based on the response, and then triggers a revalidation
- Expose `isMutating` for `useSWR`
- Supports optimistic UI functionalities (#1745)

## Breakings

- This PR doesn't affect other behaviors, but it is on top of #1864.

## `useSWRMutation`

```jsx

async function getData(url, { arg: token }) {
  ... // Fetcher implementation.
      // The extra argument will be passed via the `arg` property of the 2nd parameter.
}

// A useSWR + mutate like API, but it will never start the request.
const { data, error, trigger, reset, isMutating } = useSWRMutation('/api/user', getData, {
  // options
  onError,
  onSuccess,
  revalidate = true, // auto revalidate after mutation
  populateCache = false, // write back the response to the cache after mutation
  optimisticData,
  rollbackOnError
})
```

For `revalidate`, `populateCache`, `optimisticData` and `rollbackOnError`, check https://github.com/vercel/swr/releases/tag/1.2.0 and successor releases.

Unlike `useSWR`, there is no more `isValidating` and `mutate` returned from this hook.

## `trigger`

```jsx
// Trigger it manually, the extra argument will be passed to the mutator as `{ arg }`
try {
  const result = await trigger('my_token')
} catch (error) {
  // ...
}

// override options of `useSWRMutation`
await trigger('my_token', { revalidate: false, populateCache: true })

// in-place event handler
await trigger('my_token', { onError: () => {}, onSuccess: () => {} })
```

During a mutation, the `isMutating` state will be `true`, and any race conditions of `useSWR` and `useSWRMutation` will be carefully avoided.

## `reset`

```jsx
// reset the state (data, error, isMutating)
reset()
```

## To-dos
- [x] Finalize the name of the hook.
- [x] Sometimes, mutation APIs (POST, PUT, DELETE) return different payload from the GET API. It might be an issue if we write back to the cache with the same key. This behavior needs to be reconsidered.
- [x] If the mutation APIs' results cannot be cached, that normally means that we need to revalidate those keys again if there's a `useSWR` hook with the same key. This can be another option to consider.
- [x] Ensure callbacks and `isMutating` state are working as expected.
- [x] Ensure with a test case that under trigger mode there's no deduplication.
- [x] Ensure with a test case that `trigger` prevents race conditions with normal GET requests.
- [x] Tests for `reset`.
- [x] Potential race conditions of `reset` and `trigger` calls.
- [x] `populateCache` (#1729).
- [x] #157 (#1745).
- [x] ~~Polish event callbacks including `onDiscarded` for this mutation hook.~~
- [ ] (docs) Have a specific list of options, that can be used in `useSWRMutation` and ignore the others.


promer94:
How about
```typescript
const { mutate, trigger, reset, isMutating } = useSWRMutation('/api/user', (extraArgument) => {})
// fetcher only receive extraArgument from trigger
trigger(extraArgument)
```
trigger would have batter typescript intellisense and for some non-restful endpoints, they actually may not need the key args

mutation/index.ts
```diff
@@ -0,0 +1,115 @@
...
+    ditchMutationsTilRef.current = getTimestamp()
+    setState({ data: UNDEFINED, error: UNDEFINED, isValidating: false })
+  }, [])
+
+  useIsomorphicLayoutEffect(() => {
+    keyRef.current = key
+  })
+
+  return {
+    mutate,
```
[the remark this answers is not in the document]
shuding: Oh great catch, thank you!

shuding:
>  fetcher only receive extraArgument from trigger

Yeah that makes the design so much simpler and I have thought about it. But I wasn't sure, and this design is mostly for consistency.

promer94:
IMO The benefit of receiving keys is that user could configure a general fetcher in context rather than writing it at every hook instance. 

But in useSWRMutation user cannot set a general mutate function in context (or should we allow user to do that ?). 

The mutate function could access the keys directly in every hook instance instead of receiving it from its arguments.

shuding:
I feel that it's still able to reuse the mutator function:

```js
const { trigger } = useSWRMutation('/api/user', post)

trigger({
  name: 'yo'
})
```

And `post` is defined as:

```js
function post(API, data) {
  return fetch(API, {
    method: 'POST',
    body: JSON.stringify(data)
  }).then(res => res.json())
}
```

Which is a general utility to post data.

promer94:
I am sorry. my last reply is somehow misleading.

## My expectation
```typescript
// if user already have a mutator function like this
interface User {
  name: string
  age: number
}
const updateUser = (param: User) => post('/api/user', param)

const { trigger } = useSWRMutation('/api/user', updateUser)
// ts error here since age is string
trigger({ name: '123' , age: '12'})
```
It would be batter if swr could make use of user-defined types and let typescript do type check on trigger.


## Possible solution

Since we would not use SWRConfig to set a general mutator function
```typescript
<SWRConfig 
  value={{
    fetcher: post
   }}>
</SWRConfig>
```
The key does not have to be passed to mutator function

We could improve the type of trigger by letting mutator function only receive extraArgument. In this way it doesn't need to upgrade the typescript version.

shuding:
> Since we would not use SWRConfig to set a general mutator function

This is a good point! This hook shouldn’t using the fetcher from the context. Will get it fixed.

> We could improve the type of trigger by letting mutator function only receive extraArgument.

Yeah this can maybe fix the typing problem, but my point was that this breaks the consistency in API design. `useSWR` passes the key to fetcher and it doesn’t feel great to not do so for the mutator.

promer94:
>  Does newer TS support this feature?

It could do the trick. But it would require user to do some extra work

```typescript
interface User {
  name: string
  age: number
}
const updateUser = (param: User) => post('/api/user', param)

// developer need to type the mutator function explicitly so that trigger could get correct inferred type
const { trigger } = useSWRMutation('/api/user', (key: string, extraArguments: User) => updateUser(extraArguments))
trigger({ name: '123' , age: '12'})
```

FYI: [typescript-playground](https://www.typescriptlang.org/play?ts=4.0.5#code/C4TwDgpgBAysBOBLAdgcwNIRFAvFAzgiqlAD5TICuANtQFCiRQAqlY1Em2eA2oUmgA0UAHRjKyANbIA9gHdkPALpKG4aADFkXXFAAUASlwA+WETQ7yrdpyxqmOvHAEYsZFmw6WoWrnXvQAEoQ+DTAADzMuhLS8simeFHkAArwMgC2iPgQkab+jNAAZhDAAMYAFlrhACIAhsC10VKyCsIAgvCo+FAQAB7AEMgAJt2OUFzCAKL98LUdXbq1yCAJUHRQUPPdfQPD3XzmqMJiIijF8ONKUAD8axsbenontZ34AFwHLsdi6FMzc68lEYcKZgqFqBE6g1jEY3nd7vctj1+oMRvpDCYoJ9iN9TshzpcjLd1giEY9nq8PvwcaIfn8EACukDMWCwjV6rUYVA4STSQikTtUd1qWh3FRaDd4XzSU8xC8unDsUIoNMGVtmSCoKyIezodypdLSQKUXt0cDTCKSORxdQiQbDTL5e8CIdzVqQmyoZz9byHYbZSInYrLfTZuq3drIRzjP5-IUJKVgIgZMgoJRsjAAOqBACylAaSZTusaeBiLWQgmNuzRYwmKv+SLwSxWekkWDhW2EhWQcOKZUqyGL7VeocZ+BhcIA3iSiKhUBB4HC9PK4aqw683V66ABfacbYDlNJyCgQY+TeBpeB6ADkADkZMAoIh0jZ0oMBkNrwYd-4UAN4IUtSlNAACq2QXHuFC1G+cKWiStTznCVDpAARguP6lCmhBpmAQz1BAYELroehgC80FwoR8Bun2FQ3gA9LUYCIHR6YLtewiTlAb4HjIQxwteyQAPIwMw7FQChvEgHCABSMCCbeIiWoghQgCRZHpAY27fnQdF0VAQwQAAbhA1AyJAFzIBAEBDFAwAyLZ6i2eU0DpPm9QyBc8bIImyapn07CIKUiDANQ2D4PZB71LZSBzkRmE0DZ86PphF4QImT74gu8DWQ5kB0JhyDYZxs7zoutkAIxQNu0QZtmeYFr59GMcxrHwGJrbti6XzImqnSUG+yDAM6lFumweEDJReg7OuqD9e++AGN+wDlXonHINBED8eVABMADM15QMICGbVA147deWmxgVRXRYgsVlcA21VTVEBZrmbmFsgeg8NeDFMSx4FiZOtRwtt25KMIy6dAADLBhxHa85VTiDFD9WhFzboI02MnNg3DeBo24fhk3Y-MuNDYtDDbatUEwadO37YdUDHVt20Xdp-jXY+rH1fhcLhG0wjVMYHVSeMWBdj2+grpsm5QG6xUxaVS4rm0RiYtU1V4BimqQQeR4nmeF4eTe96Ps+r7vtZX4-vlWGPord3K7Zu3PaW2S8wMTX-a17FjcTBNUzT6102de0HUdiH02zWlAA)

anothertempore:
A question: do you consider supporting configure onSuccess, onError, etc globally? Like what useSWR does.

shuding:
@anothertempore Yes those should be supported!

mutation/index.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (isUndefined(options.populateCache)) {
+      options.populateCache = false
+    }
+
+    // Trigger a mutation, also track the timestamp. Any mutation that happened
+    // earlier this timestamp should be ignored.
+    const mutationStartedAt = getTimestamp()
+    ditchMutationsTilRef.current = mutationStartedAt
+
+    try {
```
huozhi, line 69: maybe move `setState` and `args.push` calls outside of try-catch scope? like you did in previous refactor PRs

mutation/index.ts
```diff
@@ -0,0 +1,132 @@
...
+import { withMiddleware } from '../src/utils/with-middleware'
+import { useIsomorphicLayoutEffect } from '../src/utils/env'
+import { isUndefined, UNDEFINED } from '../src/utils/helper'
+import { getTimestamp } from '../src/utils/timestamp'
+
+import { SWRMutationConfiguration, SWRMutationResponse } from './types'
+
+const mutation = <Data, Error>() => (
+  key: Key,
+  fetcher: Fetcher<Data>,
```
huozhi, line 21:
```suggestion
  fetcher: Required<Fetcher<Data>>,
```

mutation/index.ts
```diff
@@ -0,0 +1,132 @@
...
+      const serializedKey = serialize(keyRef.current)[0]
+      return mutate(serializedKey, newData, shouldRevalidate)
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  )
+
+  const trigger = useCallback(async (extraArg, opts) => {
+    if (!fetcher) {
+      throw new Error('Can’t trigger the mutation: missing fetcher.')
```
huozhi, line 53: Do you prefer type-hint rather than error message here? If it's not defined it will crash anyway and we can hint user by `Required<Type>` type in the function declaration

mutation/index.ts
```diff
@@ -0,0 +1,132 @@
+import { useCallback, useRef } from 'react'
+import useSWR, {
+  useSWRConfig,
+  Middleware,
+  Key,
+  Fetcher,
+  SWRResponse
+} from 'swr'
+
+import { serialize } from '../src/utils/serialize'
```
huozhi, line 10: Non-blocking one

we can import `unstable_serialize` here from swr instead of importing source code from src folder.
currently the type files (.d.ts) aren't able to be bundled. so it will emit some useless typing files in dist which has the similar structure of source code like `src/utils/serialize.d.ts`.

we can look for another way to optimize those imports in the future

mutation/index.ts
shuding: It’s still possible that the fetcher comes from the config arg or context.

mutation/index.ts
shuding: Yeah that would be better!

mutation/index.ts
```diff
@@ -0,0 +1,132 @@
...
+      throw new Error('Can’t trigger the mutation: missing fetcher.')
+    }
+
+    const [serializedKey, args] = serialize(keyRef.current)
+    const options = Object.assign({}, config, opts)
+
+    // Disable cache population by default.
+    if (isUndefined(options.populateCache)) {
+      options.populateCache = false
+    }
```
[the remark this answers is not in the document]
shuding: Very nice, thank you!

mutation/index.ts
shuding: Just realized that we can't use `unstable_serialize`, since here we are using the `args` as well (but `unstable_serialize` only gives the key):

```js
const [serializedKey, args] = serialize(keyRef.current)
```

mutation/index.ts
```diff
@@ -0,0 +1,125 @@
...
+      const mutationStartedAt = getTimestamp()
+      ditchMutationsTilRef.current = mutationStartedAt
+
+      setState({ isMutating: true })
+      args.push(extraArg)
+
+      try {
+        const data = await mutate(
+          serializedKey,
+          (fetcher as any)(...args),
```
MaxiMir, line 70: How about deleting line 65 and writing here like this:

```tsx
(fetcher as any)(...args.concat([extraArg])),
```

mutation/index.ts
```diff
@@ -0,0 +1,125 @@
...
+        if (ditchMutationsTilRef.current <= mutationStartedAt) {
+          setState({ data, isMutating: false })
+          options.onSuccess && options.onSuccess(data, serializedKey, options)
+        }
+        return data
+      } catch (error) {
+        // If it's reset after the mutation, we don't broadcast any state change.
+        if (ditchMutationsTilRef.current <= mutationStartedAt) {
+          setState({ error: error as Error, isMutating: false })
+          options.onError && options.onError(error, serializedKey, options)
```
MaxiMir, line 84: It is possible to shorten:

```tsx
options?.onError(error, serializedKey, options)
```

mutation/index.ts
shuding: Good call, thanks.

src/utils/state.ts
```diff
@@ -73,9 +74,17 @@ export const useStateWithDeps = <Data, Error, S = State<Data, Error>>(
     []
   )
 
-  // Always update the state reference.
   useIsomorphicLayoutEffect(() => {
-    stateRef.current = state
+    // Always update the state reference if it's not used as the initial state
+    // only.
+    if (!asInitialState) {
```
huozhi, line 80: Do you feel it make sense to move this auto update of `stateRef.current` outside of `useStateWithDeps` ?

We do auto update in use-swr.ts but not in mutation.ts

mutation/index.ts
```diff
@@ -0,0 +1,125 @@
...
+            options.onSuccess?.(data as Data, serializedKey, options)
+          }
+          return data
+        } catch (error) {
+          // If it's reset after the mutation, we don't broadcast any state change.
+          if (ditchMutationsUntilRef.current <= mutationStartedAt) {
+            setState({ error: error as Error, isMutating: false })
+            options.onError?.(error as Error, serializedKey, options)
+          }
+          throw error
```
huozhi, line 82: Should we throw error aligning with calling onError? if the ignored `mutate` is erroring, the error will still be thrown but callback is not triggered

```
initial ---> mutate -----> ignore patch
initial --------> reset --> reset to initial
```

src/utils/state.ts
shuding: Yeah makes sense!

huozhi (approved):
Looks great! Amazing work

PR #1451, opened 2021-09-12
Chore: refactor code related to unsubscribing global listeners

shuding:
- Fix typo (`unscubscibe`)
- `unsubscribe` is always a function so no need to check again
- Other tweaks

### DIFF
--- a/src/utils/cache.ts
+++ b/src/utils/cache.ts
@@ -1,6 +1,6 @@
 import { defaultConfigOptions } from './web-preset'
 import { IS_SERVER } from './env'
-import { UNDEFINED, mergeObjects } from './helper'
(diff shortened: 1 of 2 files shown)


src/utils/cache.ts
```diff
@@ -80,6 +80,6 @@ export const initCache = <Data = any>(
     // We might want to inject an extra layer on top of `provider` in the future,
     // such as key serialization, auto GC, etc.
     // For now, it's just a `Map` interface without any modifications.
-    return [provider, mutate, unscubscibe]
+    return [provider, mutate, unsubscribe]
```
huozhi, line 83: Nice catch 😂 thx

huozhi (approved):

PR #1457, opened 2021-09-13
test: refactor concurrent rendering tests

koba04:
This is a follow-up PR of #1435.
concurrent rendering tests are different from other tests because they render components with `ReactDOM.createRoot`, so I didn't use `renderWithConfig` in `test/utils.tsx`.

I've also noticed that the render function of `testing-library` has the `wrapper` option to wrap a target component. 
https://testing-library.com/docs/react-testing-library/setup#custom-render
With the option, we no longer require to override `rerender` function to wrap a target component with `SWRConfig`. It's less hacky than before, so I've refactored to use it.


shuding:
Nice find about the `wrapper` option, thanks!

koba04:
Thank you!

Issue #1412, opened 2021-08-30
Better Typing support when suspense is enabled


huozhi:
Some restrictions can be applied to swr when suspense is enanbled:
* , `null` is not allowed in `Key` and `Fetcher`
* when

FDiskas:
Already fixed in https://github.com/vercel/swr/releases/tag/1.0.1-beta.0

shuding:
No it’s not added yet. I don’t think this can be handled by TypeScript because configs can be inherited from `SWRConfig`. What we can do temporarily is to support the per-hook configuration.

Svish:
My suggestion would be to just separate the two. Either a separate hook, or just a separate import like from `swr/suspense` or something like that.

That way you could simplify the regular hook by removing all suspense related stuff from there, and the suspense hook could be written with perfect types and no non-suspense related stuff.

karol-majewski:
What should happen if `useSWR` from `swr/suspense` was used alongside a global config contradicting it?

```tsx
<SWRConfig value={{ suspense: false }}>
  <App />
</SWRConfig>
```

Svish:
Can't you already do that? Doesn't useSWR take suspense as an option, which overrides the default config?

shuding:
By definition, `useSWRSuspense` should override the global config of `suspense: false`.

Also, it would be great to cover more use cases such as waterfall (#5, #168) and dependents with this suspense hook. 

Waterfall:

```js
useSWRSuspense(key1)
useSWRSuspense(key2)
```

Dependent (should it fallback to the suspending state if the first resource changes upon focus?):

```js
const { data } = useSWRSuspense(key1)
useSWRSuspense(data.id)
```

karol-majewski:
OK, so your proposal is for the new `useSWR` to override default settings with `{ suspense: true }`. That makes sense. 

How should then conditional fetching be solved? Suspense users should also be allowed to make conditional requests, therefore `key` cannot be required to be truthy. If `key` can be falsy, then that implies `data` will sometimes be defined, sometimes not. This doesn't leave much room for improvement on the type level.

We can, however, leverage `safestring` to determine when Suspense-powered requests are guaranteed to deliver a response every time — which is when `key` is guaranteed to be truthy. Suggested improvement (simplified for brevity):

```ts
declare function useSWR<T>(key: string): { data: T  };
declare function useSWR<T>(key: string | null): { data: T | undefined };

[...]

PR #1460, opened 2021-09-14
Revert "feat: add a warning for invalid arguments with suspense mode …

koba04:
Feedback from https://github.com/vercel/swr/pull/1402#issuecomment-918375939.

The PR breaks the behavior with https://swr.vercel.app/docs/suspense#note-with-conditional-fetching.

I've noticed that the comment `useSWR requires either key or fetcher with suspense mode` was wrong 🙏
This should be `useSWR requires the both key and fetcher with suspense mode`.

How does the suspense mode work with a nullish value for key and fetcher? Does SWR allow to return `undefined` as `data`?
ref. #1412, https://github.com/vercel/swr/pull/357#issuecomment-627089889.

I'll add some tests for the behavior as another PR.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -79,10 +79,6 @@ export const useSWRHandler = <Data = any, Error = any>(
   const data = isUndefined(cached) ? fallback : cached
   const error = cache.get(keyErr)
 
-  if (suspense && (!key || !fn)) {
-    throw new Error('useSWR requires either key or fetcher with suspense mode')
-  }
-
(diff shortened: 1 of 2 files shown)


huozhi (approved):
Thanks, let's continue thinking of the better approach for this case

shuding:
Thank you! As Suspense is still experimental, I don't mind changing the behavior, or introducing a better way to cover these cases...

koba04:
Thank you!

PR #1463, opened 2021-09-15
fix: Support IE11

### LINKED ISSUE #1453: Not compatible ie11
Screenshot 2021-09-13 at 4 30 59 PM

ENG Translate
SCRIPT5007: Cannot get property '__handlers' that is undefined or null reference.

I'm using the versions of the list below.
next.js 11.1.2
swr: 1.0.0

 I checked similar issues and tried all the settings at the bottom of the issue, but the same error continues to occur.
ISSUE : https://github.com/vercel/swr/issues/278

I would appreciate it if you could let me know which part to check or try.

amannn:
Closes #1453

`addEventListener` needs to be called from `document` in IE11. Otherwise you'll run into a `Invalid calling object` error.

I've tested the proposed changes in IE11 and couldn't find any other errors.

### DIFF
--- a/src/utils/web-preset.ts
+++ b/src/utils/web-preset.ts
@@ -12,7 +12,10 @@ let online = true
 const isOnline = () => online
 
 // For node and React Native, `add/removeEventListener` doesn't exist on window.
-const onWindowEvent = (hasWindow && addEventListener) || noop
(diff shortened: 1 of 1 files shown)


shuding:
Thanks for the PR!

PR #1480, opened 2021-09-20
fix: Add clear method to Cache interface

altruisticsoftware:
https://swr.vercel.app/docs/advanced/cache#access-to-the-cache

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -190,4 +190,5 @@ export interface Cache<Data = any> {
   get(key: Key): Data | null | undefined
   set(key: Key, value: Data): void
   delete(key: Key): void
+  clear(): void
 }


shuding:
Thanks for the PR! `clear` is currently not a strong requirement for the cache provider, as we don't intend to require every method of Map to be implemented as the custom cache but only the very basic ones. I think a type-safe way to do this is:

```js
if (cache instanceof Map) cache.clear()
```

piotr-cz:
Anyway it's described in [Cache docs](https://swr.vercel.app/docs/advanced/cache#access-to-the-cache):

> cache.clear()  // ⚠️ Clear all the cache. SWR will revalidate upon re-render.

To maintain backward compatibility with custom Cache Provider implementations (which are not an instances of Map), I'd suggest adding the `clear` method` as an optional method:

```ts
export interface Cache<Data = any> {
    // ...
    clear?(): void;
}
```

PR #1482, opened 2021-09-21
chore: Code refactoring

shuding:
- Renames `shouldRevalidate` to just `revalidate`
- Binds event listener registers
- Changes the `shouldDedupe` logic as `shouldStartNewRequest`

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -125,7 +125,11 @@ export const useSWRHandler = <Data = any, Error = any>(
       let startAt: number
       let loading = true
       const opts = revalidateOpts || {}
-      const shouldDedupe = !isUndefined(CONCURRENT_PROMISES[key]) && opts.dedupe
+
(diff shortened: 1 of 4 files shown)


src/use-swr.ts
```diff
@@ -141,29 +145,22 @@ export const useSWRHandler = <Data = any, Error = any>(
         delete CONCURRENT_PROMISES_TS[key]
       }
 
-      // start fetching
-      try {
-        cache.set(keyValidating, true)
-        setState({
-          isValidating: true
-        })
-        if (!shouldDedupe) {
```
huozhi, line 150: now it will broadcast anyway, will this results into more revalidations?

shuding: Good catch!

shuding: Fixed

huozhi (approved):
Looks good!

PR #1486, opened 2021-09-21
Upgrade typescript to 4.4.3

huozhi:
Upgrade ts to v4.4.3

There's a problem that we have to resolve tslib to 2.3.0 for both bundling and linting. The dep `tsutil` of typescript eslint is still resolving tslib v1, which will break the linting.


shuding (approved):
Thank you!

PR #1492, opened 2021-09-23
ci: multiple dir type checking

huozhi:
Current `types:check` script will skip sub folders like `inifinite` etc. tweaks some ts configuration include sub folders

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -12,7 +12,7 @@
     "noUnusedParameters": true,
     "strictBindCallApply": true,
     "outDir": "./dist",
-    "rootDir": "src",
+    "rootDir": ".",
     "strict": true,
     "target": "es5",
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -24,8 +24,15 @@
     },
     "typeRoots": ["./src/types", "./node_modules/@types"]
   },
-  "include": ["src"],
-  "exclude": ["dist", "node_modules"],
+  "include": [
+    "src", 
+    "immutable",
```
shuding, line 29: Will this also change the bundled files to include everything?

huozhi: No, bundler collect files from entry. but change `tsc` behavior like if you run it directly it will compile all these folders. That's why I use them for typing check as well

PR #1493, opened 2021-09-23
fix: Refactor core revalidate function and fix isValidating state bug

shuding:
First of all this PR refactors the internal `revalidate` implementation, which is the most important part of the core. Since both the success and error branches share the same logic to update states like `isValidating` and `error`, they're moved to the same places before/after fetching. The code is also cleaner.

Correspondingly 3 new tests are added while the last one `should sync all loading states if errored but paused` will fail with the latest release.


src/use-swr.ts
```diff
@@ -235,70 +237,56 @@ export const useSWRHandler = <Data = any, Error = any>(
...
-            getConfig().onErrorRetry(err, key, config, revalidate, {
-              retryCount: (opts.retryCount || 0) + 1,
-              dedupe: true
-            })
+          // Error event and retry logic.
+          if (isCallbackSafe()) {
+            getConfig().onError(err, key, config)
+            if (config.shouldRetryOnError) {
+              // When retrying, dedupe is always enabled
+              getConfig().onErrorRetry(err, key, config, revalidate, {
```
huozhi, line 269: the retry is now before state update, will this lead to any inconsistence?

shuding: Should be good because state update will be scheduled til next tick, and this was called synchronously.

huozhi (approved):
Nice work

PR #1477, opened 2021-09-19
Type useSWR fetcher

promer94:
- [x] improve the type check process of the project.  Currently we only do type check in src and test folder.
- [x] support ```useSWR({a: number, b: string}, (args) => args)```
- [x] support ```useSWR(cond() ? {a: number, b: string} : null , (args) => args)```
- [x] fix current build error 
- [x] find a batter way to test type

## Playground Link
[...]

PR #1496, opened 2021-09-26
chore: Refine type interfaces

shuding:
Based on #1477, this PR refines some of the internal and exported type names and interfaces:
- Exported `ValueKey` → `Arguments` for accuracy
- Exported `Result` → removed, as it's a simple `T | Promise<T>` wrapper
- New exported `InfiniteKeyLoader` from `swr/infinite`, we can deprecate `KeyLoader` of `swr` later
- Rename internal type name `Args` to `SWRKey` for accuracy
- Other code refactoring

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,35 +1,33 @@
 import * as revalidateEvents from './constants/revalidate-events'
 
-export type Result<T = unknown> = T | Promise<T>
+export type FetcherReturnValue<Data = unknown> = Data | Promise<Data>
 
-export type Fetcher<Data = unknown, Args extends Key = Key> =
+export type Fetcher<Data = unknown, SWRKey extends Key = Key> =
   /**
    * () => [{ foo: string }, { bar: number }] | null
-   *
    * () => ( [{ foo: string }, { bar: number } ] as const | null )
    */
-  Args extends (() => readonly [...infer K] | null)
-    ? ((...args: [...K]) => Result<Data>)
+  SWRKey extends (() => readonly [...infer Args] | null)
+    ? ((...args: [...Args]) => FetcherReturnValue<Data>)
     : /**
      * [{ foo: string }, { bar: number } ] | null
-     *
      * [{ foo: string }, { bar: number } ] as const | null
      */
-    Args extends (readonly [...infer K])
-    ? ((...args: [...K]) => Result<Data>)
+    SWRKey extends (readonly [...infer Args])
+    ? ((...args: [...Args]) => FetcherReturnValue<Data>)
     : /**
      * () => string | null
      * () => Record<any, any> | null
      */
-    Args extends (() => infer T | null)
-    ? (...args: [T]) => Result<Data>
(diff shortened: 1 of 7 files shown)


src/types.ts
```diff
@@ -1,35 +1,33 @@
 import * as revalidateEvents from './constants/revalidate-events'
 
-export type Result<T = unknown> = T | Promise<T>
+export type FetcherReturnValue<Data = unknown> = Data | Promise<Data>
```
huozhi, line 3: how about name `FetcherReturnValue` as `FetcherResponse`?

src/types.ts
```diff
@@ -94,55 +91,54 @@ export interface PublicConfiguration<
...
   key: Args,
   fetcher: Fetcher<Data, Args> | null,
   config: SWRConfiguration<Data, Error>
 ) => SWRResponse<Data, Error>
 
-export type Middleware = (useSWRNext: SWRHook) => SWRHookWithMiddleware
-export type TupleKey = [any, ...unknown[]] | readonly [any, ...unknown[]]
-export type ValueKey = string | null | TupleKey | Record<any, any>
-export type Key = ValueKey | (() => ValueKey)
+type ArgumentsTuple = [any, ...unknown[]] | readonly [any, ...unknown[]]
```
huozhi, line 139: small Q: is readonly here necessary?

huozhi (approved):
2 small non-blocking comments, LGTM

src/types.ts
```diff
@@ -94,55 +91,54 @@ export interface PublicConfiguration<
...
-        ]
+      | [SWRKey]
+      | [SWRKey, Fetcher<Data, SWRKey> | null]
+      | [SWRKey, SWRConfiguration<Data, Error, SWRKey> | undefined]
       | [
-          Args,
+          SWRKey,
           Fetcher<Data, Key> | null,
-          SWRConfiguration<Data, Error, Args, Fetcher<Data, Args>>
+          SWRConfiguration<Data, Error, SWRKey>
```
promer94, line 125:
```suggestion
SWRConfiguration<Data, Error, SWRKey> | undefined
```

src/types.ts
shuding: Sounds great! 👍

src/types.ts
shuding: Yeah I think they're different for TS tuples, cc @promer94 who initially implemented it.

promer94: Great catch ! It could be removed.
```typescript
// the origin implementation looked like this
type Fetcher<Args extends Key> = Args extends (() => [...infer K] | readonly [...infer K])
  ? ((...args: [...K]) => Result<Data>) : ...

// But then shu made me realize the it could have batter implementation
// So the Args does not have to satisfy [any, ...unknown[]] | readonly [any, ...unknown[]] any more
type Fetcher<Args extends Key> = Args extends (() => readonly [...infer K] | null)
  ? ((...args: [...K]) => Result<Data>) : ...
```

PR #1497, opened 2021-09-26
chore: Improve project file structure

shuding:
- Add `cache: 'yarn'` to the GitHub workflow.
- Move `jest-setup.ts` into `scripts`.
- Move `test-type` into `test`.
- Create `tsconfig.check.json` only for type checking since `include: ["test/type"]` will also output the type of the test file in `dist`.


huozhi (approved):

tsconfig.check.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "./tsconfig.json",
+  "include": ["src", "immutable", "infinite", "test/type"]
```
huozhi, line 3: I think we can just add `test/type` to tsconfig.json in rootDir. no need to have an extra config

shuding: As mentioned in the description including `"test/type"` in tsconfig.json will also output the type of the test file in dist and get distributed to npm.

huozhi: I c, sounds good to merge!

huozhi: I tried locally again didn't see the output of type. did you run `tsc` without `--noEmit` so that the types are generated?
filed #1507 to merge them

PR #1498, opened 2021-09-26
fix: Invalidate the key when mutating without revalidating

### LINKED ISSUE #828: Clear dedupingInterval of an unmounted key

shuding:
Fixes #828, closes #1139.

There're some scenarios that a revalidation **won't** happen when mutating a key:
- When `mutate(key)` but there is no mounted `useSWR(key)` hook. A revalidation might not happen after mounting hook if it's still within the dedupling interval.
- When `mutate(key, data)` but there is no mounted `useSWR(key)` hook. Same reason as above.

When it doesn't revalidate immediately, it should skip deduplication until the next revalidation time. This PR fixes the bug.

Worth mentioning that the fix was originally proposed by @MoonBall, thank you!


src/use-swr.ts
```diff
@@ -262,7 +261,8 @@ export const useSWRHandler = <Data = any, Error = any>(
           broadcastState(cache, key, newData, newState.error, false)
         }
       } catch (err) {
-        cleanupState()
+        // Reset the state immediately.
+        cleanupState(startAt!)
```
huozhi, line 265: Maybe give startAt a default value -1 or disable the lint rule `Forbidden non-null assertion`

shuding: Sure!

shuding: I decided to disable that line only since the rule is overall helpful.

huozhi (approved):

Issue #1502, opened 2021-09-27
Bug: cache object doesn't have the method `clear`


Ne0xygene:
# Bug report

## Description / Observed Behavior

While cache seems to be an instance of Map, I cannot access to its methods as the following example : 
`cache.clear()`
While it stills work, my typescript linter isn't happy about it.

## Expected Behavior

`cache.clear()` should be availaible and shouldn't return a typescript error.

SWR version 1.0.1. 

Link to the discussion : 
https://github.com/vercel/swr/discussions/1494

kelvin9314:
@huozhi @shuding  Hello guys,

we know that the return of ```cache```, it actually is  a Map instance
shall we just add a clear function in the interface , like 

```
export interface Cache<Data = any> {

    get(key: Key): Data | null | undefined;

    set(key: Key, value: Data): void;

    delete(key: Key): void;
    
    clear(): void;

}
```

then we might pass the typescript linter error ??

allenchuang:
following because i see the same issue.

AdamKwoka:
Me too I still have problem with cache.clear();

huozhi:
Please checkout https://github.com/vercel/swr/issues/1887#issuecomment-1171269211 Thanks!

PR #1505, opened 2021-09-28
fix: add `clear` method to Cache interface

Ne0xygene:
Refer to issue #1502


huozhi:
Hi @Ne0xygene we're not tend to add clear as a required api on cache. It could not list all the methods of provider.
I filed #1506 to make the cache type can be inferred by provider. Then you're able to use any method of provider itself in code

PR #1506, opened 2021-09-28
types: infer provider type as cache type

### LINKED ISSUE #1502: Bug: cache object doesn't have the method `clear`

huozhi:
* make `Cache` a type that extends from required provider interface
* make `useSWRConfig` generic with provider type `useSWRConfig<Provider = Cache>`
* add typing test

Closes #1502
Closes #1505


src/types.ts
```diff
@@ -234,8 +236,16 @@ export type StateUpdateCallback<Data = any, Error = any> = (
...
 ) => void
 
-export interface Cache<Data = any> {
+export interface ICache<Data = any> {
   get(key: Key): Data | null | undefined
   set(key: Key, value: Data): void
   delete(key: Key): void
 }
+
+export type DefaultProvider = Map<string | null, any>
```
shuding, line 245: Why `string | null` instead of just `string`?

shuding:
I wonder if there's a way to automatically infer the type, without requiring a generic type to be passed.

src/utils/config-context.ts
```diff
@@ -8,33 +14,38 @@ import {
...
 
-const SWRConfig: FC<{
-  value?: SWRConfiguration &
-    Partial<ProviderConfiguration> & {
-      provider?: (cache: Readonly<Cache>) => Cache
-    }
-}> = ({ children, value }) => {
+type SWRConfigValue = SWRConfiguration &
+  Partial<ProviderConfiguration> & {
+    provider?: CacheProvider<any>
```
promer94, line 25: This does not look correct to me.
```typescript
export type Cache<T = DefaultProvider, Data = any> = T extends ICache<Data>
  ? T
  : never

export type CacheProvider<Data = any> = (cache: Cache<Data>) => Cache<Data>

CacheProvider<any> = (cache: Cache<any>) => Cache<any> = (cache: any) => any
```
We need provider to return a Cache here

promer94:
I think the actual problem here is that we can't safely type the cache without runtime checking.

```const { cache } = useSWRConfig()```

The cache which is returned from the hook probably does not have the clear interface.

Currently if users want to extend the cache interface. They have to write a type assertion. 

```typescript
// CustomCache.tsx
import type { Cache } from "swr"
import { SWRConfig, useSWRConfig } from "swr"
type MyCustomCache = Map<string, any>

const customCache: MyCustomCache = new Map()

const CustomCacheProvider = ({ children }) => {
  return <SWRConfig value={{ provider: () => cache }}>{children}</SWRConfig>
}
// This is just a example. The actual runtime check might be more complex
function checkCache(cache: Cache): cache is MyCustomCache {
  return cache === c
}

const useCustomCache = () => {
  const { cache } = useSWRConfig()
  if (checkCache(cache)) return cache
  throw new Error("customCache can only be used within CustomCacheProvider")
}
```

```typescript
// App.tsx
const ClearButton = () => {
  const cache = useCustomCache()
  return (
    <button
      onClick={() => {
        // type safe and runtime safe
        cache.clear()
      }}
    ></button>
  )
}

const App = () => {
  return (
    <>
      <CustomCacheProvider>
        <ClearButton />
      </CustomCacheProvider>
      {/* <ClearButton/>*/}
    </>
  )
}
```

We need to find a batter way to help user to extend current cache interface (runtime checking is still required) and provider batter type for default cache.

src/types.ts
huozhi: because `swr/infinite` has some case that cache key is `string | null`

like

```ts
  const resolvePageSize = useCallback((): number => {
    const cachedPageSize = cache.get(pageSizeCacheKey)
    return isUndefined(cachedPageSize) ? initialSize : cachedPageSize
```

src/utils/config-context.ts
huozhi: I didn't get what do you mean here. Can you rephrase it more specificly?

promer94: ## Previous
```typescript
type Provider: (cache: Readonly<Cache>) => Cache
const cache = new Map()
<SWRConfig value={{ provider: () => cache }}>{children}</SWRConfig>
// This will throw type error since { a: foo } does not satisfy the requirement
<SWRConfig value={{ provider: () => {a: foo } }}>{children}</SWRConfig>
```

## Current
```typescript
type Provider = CacheProvider<any> = (_: any) => any
const cache = new Map()
<SWRConfig value={{ provider: () => cache }}>{children}</SWRConfig>
// This will not throw type error any more
<SWRConfig value={{ provider: () => {a: foo } }}>{children}</SWRConfig>
```

src/types.ts
shuding: That feels like a bug (facepalm)

PR #1507, opened 2021-09-28
chore: remove duplicate configs

huozhi:
`test/type` files are included in `test/tsconfig.json`. so we can run two project for both src and tests


shuding:
This will generate an extra typing file of `test/type` in `dist`...

huozhi:
missed it in the first checking..let's figure out a better way to not generate the unless typing

shuding:
Wow good idea!

PR #1508, opened 2021-09-29
chroe: use yalc for local development

### LINKED ISSUE #1427: Better way of debugging with local swr assets
Currently use scripts to install node_modules and next inside examples, and run them with require hooks to debug with local swr assets. It's not deal since it's too complicated and not scalable.

### Goal
* Find a better way to debug with example / test case with local swr
* Less dependencies and easy setup

promer94:
close #1427


shuding:
Thanks!

PR #1509, opened 2021-09-29
typing: fix-type-generation

promer94:
https://codesandbox.io/s/type-fetcher-forked-38g6f?file=/src/App.tsx

# 1.0.1
build

# Current
miss-index d ts


huozhi (approved):
Good catch! Looks like we have to keep the `tsconfig.check.json` for a while...

PR #1514, opened 2021-09-30
chore: Mark returned states and mutate callback argument as readonly

### LINKED ISSUE #1386: Make returned state (data, error, isValidating) readonly

shuding:
Closes #1386.

### DIFF
--- a/infinite/index.ts
+++ b/infinite/index.ts
@@ -41,7 +41,7 @@ export const infinite = ((<Data, Error, Args extends Arguments>(
 ): SWRInfiniteResponse<Data, Error> => {
   const rerender = useState({})[1]
   const didMountRef = useRef<boolean>(false)
-  const dataRef = useRef<Data[]>()
+  const dataRef = useRef<Readonly<Data[]>>()
 
   const {
     cache,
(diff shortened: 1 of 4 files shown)


infinite/index.ts
```diff
@@ -167,7 +167,12 @@ export const infinite = ((<Data, Error, Args extends Arguments>(
 
   const mutate = useCallback(
     (
-      data: Data[] | undefined | Promise<Data[]> | MutatorCallback<Data[]>,
+      data?:
+        | Data[]
+        | Readonly<Data[]>
+        | Promise<Data[]>
+        | Promise<Readonly<Data[]>>
+        | MutatorCallback<Data[]>,
```
huozhi, line 175: do we need `MutatorCallback<Readonly<Data[]>>` here?

huozhi (approved):
left 1 comment, rest LGTM

infinite/index.ts
shuding: Nope I think, `MutatorCallback<Data = any>` already wraps readonly inside.

PR #1515, opened 2021-10-01
chore: Add test to ensure mutate accepts undefined

### LINKED ISSUE #1481: mutate(undefined) doesn't work
# Bug report

```typescript
const [fallback, setFallback] = useState<Data>(...)
const res = useSWR<Data>(..., { fallbackData: fallback })

console.log("fallback", fallback)
console.log("data", res.data)

function unset(){
  setFallback(undefined)
  res.mutate(undefined, false)
}
```

With this code, I expect res.data to be undefined on next render when calling unset()
However, it is still the same value even though the fallback has been set to undefined

before unset:
```
fallback {...}
data {...}
```

after unset:
```
fallback undefined
data {...}
```

## Additional Context

swr@1.1.0-beta.0

shuding:
Closes #1481.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -340,12 +340,13 @@ export const useSWRHandler = <Data = any, Error = any>(
             error: updatedError,
             isValidating: updatedIsValidating
           },
-          // if data is undefined we should not update stateRef.current.data
-          !compare(stateRef.current.data, updatedData)
-            ? {
+          // Since `setState` only shallowly compares states, we do a deep
+          // comparison here.
+          compare(stateRef.current.data, updatedData)
+            ? UNDEFINED
+            : {
                 data: updatedData
               }
-            : null
         )
       )
     }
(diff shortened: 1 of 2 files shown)


src/use-swr.ts
```diff
@@ -306,8 +306,9 @@ export const useSWRHandler = <Data = any, Error = any>(
   // Similar to the global mutate, but bound to the current cache and key.
   // `cache` isn't allowed to change during the lifecycle.
   const boundMutate: SWRResponse<Data, Error>['mutate'] = useCallback(
-    (newData, shouldRevalidate) =>
-      internalMutate(cache, keyRef.current, newData, shouldRevalidate),
+    // By using `bind` we don't need to modify the size of the rest arguments.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
huozhi, line 310: lint warning:
```
src/use-swr.ts#L315
React Hook useCallback received a function whose dependencies are unknown. Pass an inline function instead
```

guess you need to put it before line of deps.

We had issue with bind functions before, will it still have the similar issue then?

shuding: > We had issue with bind functions before, will it still have the similar issue then?

It should be OK since here we're always referring to the latest key. We've also added test for it.

huozhi:
LGTM (late 😂)

PR #1548, opened 2021-10-13
feat: using latest version of `axios` with fetch interceptor

danestves:
In the previous example, if we update to the latest version of `axios` we will have this error:

image

I search for a workaround and I found this old issue https://github.com/axios/axios/issues/1510#issuecomment-396894600 I ended up with this solution

image

And the TypeScript error has gone


shuding:
Thank you!

PR #1556, opened 2021-10-14
fix: change config alteration process in withMiddleware to be immutable

### LINKED ISSUE #1555: withMiddleware causes hook execution order change when using a shared config object
# Bug report

> What kind of issues did you encounter with SWR?

Using useSWRInfinite with a shared config object will cause the original config object to be mutated, causing other hooks that share the config object to have altered hook execution order(A.k.a. violating the React rule of hooks)

## Expected Behavior

> How did you expect SWR to behave here?

useSWR, useSWRInfinite should not have altered hook execution order even when a shared config object is used.

Either of the solutions should work.
1. Change withMiddleware to immutably copy the config object rather than altering the original (personally recommended)
  - [The culprit code](https://github.com/vercel/swr/blob/ecba927ab79c9efffa98a8e5f4db8fadb5a61ac8/src/utils/with-middleware.ts#L18)
  - [PR for a fix](https://github.com/vercel/swr/pull/1556)
2. Add guide in the documentation to prevent users from sharing the same config object across hooks

## Repro Steps / Code Example

[CodeSandbox](https://codesandbox.io/s/with-middleware-mutable-config-bug-b65cb?file=/src/App.js)

## Additional Context

> SWR version.

v1.0.1 

> Add any other context about the problem here.

junsushin-dev:
- Enable users to use shared config objects across useSWR, useSWRInfinite without the React rule of hooks error

closes #1555

### DIFF
--- a/src/utils/with-middleware.ts
+++ b/src/utils/with-middleware.ts
@@ -15,7 +15,7 @@ export const withMiddleware = (
       | [Key, Fetcher<Data> | null, SWRConfiguration | undefined]
   ) => {
     const [key, fn, config] = normalize(args)
-    config.use = (config.use || []).concat(middleware)
-    return useSWR<Data, Error>(key, fn, config)
+    const middlewares = (config.use || []).concat(middleware)
+    return useSWR<Data, Error>(key, fn, { ...config, use: middlewares })
   }
 }


shuding:
Thank you for fixing it!

junsushin-dev:
@shuding 
Thanks for the feedback :) Thumbs up for the maintainers keeping this project running. 
Been using SWR since start of 2021, a great library indeed. Glad to contribute!

PR #1569, opened 2021-10-16
test: speed up with @swc-node/jest

### LINKED ISSUE #1568: Switch to @swc-node/jest
https://www.npmjs.com/package/@swc-node/jest

huozhi:
Resolves #1568 

* Replace `ts-jest` with `@swc-node/jest`
* Remove module mapper and leverage tsconfig paths by swc

On mac m1 time reduced from 10s to 5s. blazing fast!

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,5 +1,4 @@
 module.exports = {
-  preset: 'ts-jest',
   testEnvironment: 'jsdom',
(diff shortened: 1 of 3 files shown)


huozhi:
looks like if we have syntax error in test cases swc-jest won't capture it unlike previous ts-jest

shuding:
👏

PR #1596, opened 2021-10-27
enhance: bind removeEventListener

huozhi:
* bind removeEventListener to window, align with addEventListener
* fix lint in test


huozhi:
Add some test but pretty hard to mock the defined value like `hasDocument = typeof document !== 'undefined'`, changed them to function for convenience

shuding:
👍

src/utils/helper.ts
```diff
@@ -13,5 +13,5 @@ export const isFunction = (v: any): v is Function => typeof v == 'function'
 export const mergeObjects = (a: any, b: any) => OBJECT.assign({}, a, b)
 
 const STR_UNDEFINED = 'undefined'
-export const hasWindow = typeof window != STR_UNDEFINED
-export const hasDocument = typeof document != STR_UNDEFINED
+export const hasWindow = () => typeof window != STR_UNDEFINED
```
tunguyen-ct, line 16: @huozhi Could you tell me why do we need to use a function here? I'm curious to learn.

nguyenyou, line 16: @huozhi Could you tell me why do we need to use a function? I'm curious to learn.

huozhi: It's for both source code and tests, if we make it a function, then in the test case we could switch environment between node and jsdom so that the evaluation is different when runtime changes.

I didn't find a good way to keep it a pre-evaluated variable while switching runtime. If you have any good idea to keep it simple that would be great 🙏

shuding: It would be great to add a comment in the source code ;)

nguyenyou: Thank you 🙏 I think this is the simplest way to achieve it 👍

Issue #1617, opened 2021-11-05
create-react-app doesn't support mjs


Kamahl19:
Updating to 1.1.0.beta-8 doesn't work on CRA based apps because of `Can't import the named export 'createContext' from non EcmaScript module (only default export is available)`

This is not really a problem of SWR but if we want to use this package on CRA based apps, the mjs has to be reverted. 

Additional context https://github.com/facebook/create-react-app/issues/10356

Other libs have the same issue https://github.com/formatjs/formatjs/issues/1395

Issue #1614, opened 2021-11-04
[React Native] [1.1.0-beta.8] `main` module field that could not be resolved


beetlebum:
error: Error: While trying to resolve module `swr` from file `/project/src/core/api/common/ApiConfig.tsx`, the package `/project/node_modules/swr/package.json` was successfully found. However, this package itself specifies a `main` module field that could not be resolved (`/project/node_modules/swr/dist/index.mjs`. Indeed, none of these files exist:

  * /project/node_modules/swr/dist/index.mjs(.native|.ios.js|.native.js|.js|.ios.json|.native.json|.json|.ios.ts|.native.ts|.ts|.ios.tsx|.native.tsx|.tsx)
  * /project/node_modules/swr/dist/index.mjs/index(.native|.ios.js|.native.js|.js|.ios.json|.native.json|.json|.ios.ts|.native.ts|.ts|.ios.tsx|.native.tsx|.tsx)
    at DependencyGraph.resolveDependency (/project/node_modules/metro/src/node-haste/DependencyGraph.js:311:17)
    at Object.resolve (/project/node_modules/metro/src/lib/transformHelpers.js:129:24)
    at resolve (/project/node_modules/metro/src/DeltaBundler/traverseDependencies.js:396:33)
    at /project/node_modules/metro/src/DeltaBundler/traverseDependencies.js:412:26
    at Array.reduce (<anonymous>)
    at resolveDependencies (/project/node_modules/metro/src/DeltaBundler/traverseDependencies.js:411:33)
    at processModule (/project/node_modules/metro/src/DeltaBundler/traverseDependencies.js:140:31)
    at async addDependency (/project/node_modules/metro/src/DeltaBundler/traverseDependencies.js:230:18)
    at async Promise.all (index 0)
    at async processModule (/project/node_modules/metro/src/DeltaBundler/traverseDependencies.js:198:5)

huozhi:
Looks like metro currently doesn't support `.mjs` as source file extension
https://github.com/facebook/metro/issues/535

PR #1618, opened 2021-11-05
Revert "chore: Add type: "module" and use .mjs extension (#1604)"

huozhi:
Fixes: #1614
Fixes: #1617

Revert "chore: Add type: "module" and use .mjs extension (#1604)"

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,28 +2,27 @@
   "name": "swr",
   "version": "1.1.0-beta.8",
   "description": "React Hooks library for remote data fetching",
-  "type": "module",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -4,26 +4,26 @@
   "description": "React Hooks library for remote data fetching",
   "type": "module",
   "main": "./dist/index.js",
-  "module": "./dist/index.mjs",
+  "module": "./dist/index.js",
```
shuding, line 7: Module and main are pointing to the same file here.

Issue #1621, opened 2021-11-05
TS throws after changing useSWR typings


Kamahl19:
# Bug report

## Description / Observed Behavior

```
Type '(org: string, repo: string) => Promise<any>' is not assignable to type '((args_0: string) => FetcherResponse<{ name: string; description: string; }>) | ((args_0: any, ...args_1: unknown[]) => FetcherResponse<{ name: string; description: string; }>) | ((args_0: Record<...>) => FetcherResponse<...>) | ((args_0: string | ... 2 more ... | Record<...>) => FetcherResponse<...>) | null'.ts(2769)
```

This error occurres when i use multiple arguments and generics for data

## Repro Steps / Code Example

https://codesandbox.io/s/create-react-app-typescript-forked-yjc49?file=/src/index.tsx

## Additional Context

Happening since 1.1.0-beta-3 , even on beta 7

shuding:
Closed by #1640.

Issue #1518, opened 2021-10-01
Key type inference doesn't work if the data generic is provided


shuding:
# Bug report

## Description / Observed Behavior

```ts
useSWR<Data>('/api', (endpoint) => sendRequestFor(endpoint))
                      ﹌﹌﹌﹌﹌ "Parameter 'endpoint' implicitly has an 'any' type.ts(7006)"
```

## Expected Behavior

It would be great if we can workaround this.

## Repro Steps / Code Example

N/A

## Additional Context

SWR 1.1.0-beta.3.

shuding:
Update: we need partial inference for this, AFAIK it's still not supported by TS yet (https://github.com/microsoft/TypeScript/issues/26242).

shuding:
Closed by #1640.

PR #1640, opened 2021-11-16
Tweak types

promer94:
* make ```useSWR<{ id: number }>('/api/', (key) => ({ id: 123 }))``` still works as before 
so users dont need to update their current code
* add test case for #1586 
* remove some comments (replace with ts playground link later)

### DIFF
--- a/infinite/types.ts
+++ b/infinite/types.ts
@@ -1,4 +1,4 @@
-import { SWRConfiguration, SWRResponse, Arguments } from 'swr'
+import { SWRConfiguration, SWRResponse, Arguments, OriginFetcher } from 'swr'
 
 type FetcherResponse<Data = unknown> = Data | Promise<Data>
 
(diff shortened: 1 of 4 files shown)


huozhi (commented):
> make useSWR<{ id: number }>('/api/', (key) => ({ id: 123 })) still works as before
Which issue is this item trying to resolve? Could you elabrate?

What's the purpose of exporting `OriginFetcher` and when will user use it?

promer94:
> Which issue is this item trying to resolve? Could you elabrate?

Because the limitation of typescript. It could not support partial type argument inference

 ```typescript
// These two will works
useSWR<{ id: number }>('/api/', (key) => ({ id: 123 }))
useSWR<{ id: number }, any, string>('/api/', (key) => ({ id: 123 }))


useSWR<{ id: number }>('/api/', (key) => ({ id: 123 }))
// this would make fetcher's type fallback to 
// ((args_0: string) => FetcherResponse<boolean>) | ((args_0: any, ...args_1: unknown[]) => FetcherResponse<boolean>) | ((args_0: Record<any, any>) => FetcherResponse<...>) | ((args_0: string | ... 2 more ... | Record<...>) => FetcherResponse<...>) | null'
// it would cause type error

```
 
 

related issue https://github.com/vercel/swr/issues/1518#issue-1013082584  
real world problem https://github.com/vercel/swr/pull/1477#issuecomment-961845025 https://github.com/vercel/swr/issues/1621#issue-1045895828



> What's the purpose of exporting OriginFetcher

useSWRInfinite also uses this type

shuding (commented):
Thanks, this looks like a great workaround! Can you add a test for the `Data` inference case you mentioned?

Also regarding `OriginFetcher` I think that `BareFetcher` might be a better name.

shuding:
❤️

huozhi:
Nice work! Thanks

PR #1658, opened 2021-11-24
type: disbale inference when data generic is provided

promer94:


infinite/types.ts
```diff
@@ -3,8 +3,8 @@ import { SWRConfiguration, SWRResponse, Arguments, BareFetcher } from 'swr'
 type FetcherResponse<Data = unknown> = Data | Promise<Data>
 
 export type InfiniteFetcher<
-  KeyLoader extends InfiniteKeyLoader = InfiniteKeyLoader,
-  Data = any
+  Data = any,
```
huozhi, line 6: Is it possible to keep the order of generics here like before?

infinite/index.ts
```diff
@@ -32,12 +31,11 @@ export const unstable_serialize = (getKey: InfiniteKeyLoader) => {
...
 }
 
-export const infinite = ((<Data, Error, Args extends Arguments>(
-  useSWRNext: SWRHook
-) => (
+export const infinite = ((<Data, Error>(useSWRNext: SWRHook) => (
   getKey: InfiniteKeyLoader,
-  fn: Fetcher<Data> | null,
-  config: typeof SWRConfig.default & SWRInfiniteConfiguration<Data, Error, Args>
+  fn: BareFetcher<Data> | null,
```
huozhi, line 36: Is Inifnite not be able to adopt Fetcher?

promer94: This type is only used internally here.  It won't affect type inference

Issue #1657, opened 2021-11-24
Support in SSR of IE11


Sh031224:
# Bug report

Support in SSR of IE11.

## Description / Observed Behavior

`Object doesn't support property or method 'attachShadow'.` error when running [this repository code](https://github.com/Sh031224/swr-next-ie11) in IE11.

![image](https://user-images.githubusercontent.com/51149996/143204756-6cb21d88-a34f-42af-b4c0-43dc5532f34e.png)

This problem is also registered in the [nextjs issue.](https://github.com/vercel/next.js/issues/13231)
So I tried [this solution](https://github.com/vercel/next.js/issues/13231#issuecomment-711484680), but got another error as below.

![image](https://user-images.githubusercontent.com/51149996/143204858-f466ec69-e110-4591-a078-3d891522d9b8.png)

You can check the above error is [here](https://github.com/Sh031224/swr-next-ie11/tree/polyfills)

Of course, if i just configure CSR without adding SSR settings, it works fine in IE11.

## Expected Behavior

It should work without errors like any other browser.

## Repro Steps / Code Example

[default](https://github.com/Sh031224/swr-next-ie11)
[polyfills](https://github.com/Sh031224/swr-next-ie11/tree/polyfills)

## Additional Context

SWR version. 1.1.0-beta.10
Next version. 12.0.4

PR #1659, opened 2021-11-24
fix: supports ssr in ie11(#1657)

Sh031224:
Fixes: #1657 

It works fine without the polyfill settings mentioned in the issue above.

Instead of importing the rAF function at the moment of import, the function that meets the condition is imported when necessary.


shuding:
Thanks!

PR #1672, opened 2021-11-26
Directly use useReducer for rerender purpose

nguyenyou:
Since the `useState` hook is a wrapper around the `useReducer` hook, directly using the `useReducer` hook for rerender purposes would save some function calls.


shuding:
By the way, would you mind sharing some more information or proof of that useReducer is more performant? Curious to learn!

nguyenyou:
I'm not pretty sure using useReducer will gain some performance boost, based on my naive thought that useState itself using useReducer internally, by directly using useReducer, like a shorter path to achieve the same goal.

PR #1674, opened 2021-11-27
chore: upgrade eslint to v8 and prettier to v2

huozhi:
please review this change with disabling the whitespace changes

* upgrade eslint to v7 and related packages
* upgrade prettier to v2


.prettierrc
```diff
@@ -2,5 +2,7 @@
   "semi": false,
   "singleQuote": true,
   "useTabs": false,
-  "tabWidth": 2
+  "trailingComma": "none",
```
huozhi, line 5: for compatibale with v1

.eslintrc
```diff
@@ -46,6 +45,7 @@
     "@typescript-eslint/ban-types": 0,
     "react-hooks/rules-of-hooks": 2,
     "react-hooks/exhaustive-deps": 1,
-    "react/prop-types": 0
+    "react/prop-types": 0,
+    "testing-library/no-unnecessary-act": 0
```
huozhi, line 49: it gives few wrong error for act

infinite/types.ts
```diff
@@ -6,13 +6,13 @@ export type InfiniteFetcher<
   Data = any,
   KeyLoader extends InfiniteKeyLoader = InfiniteKeyLoader
 > = KeyLoader extends (...args: any[]) => any
-  ? ReturnType<KeyLoader> extends (
+  ? ReturnType<KeyLoader> extends
       | readonly [...infer K]
       | null
       | false
-      | undefined)
-    ? ((...args: [...K]) => FetcherResponse<Data>)
```
huozhi, line 14: I wonder if we should keep the parentheses? @shuding wdyt

test/use-swr-concurrent-rendering.test.tsx
```diff
@@ -29,6 +29,7 @@ describe('useSWR - concurrent rendering', () => {
 
     renderWithConfig = (element, config) =>
       act(() =>
+        // eslint-disable-next-line testing-library/no-render-in-setup
```
huozhi, line 32: it's a wrong error, we didn't call this function actually

test/use-swr-integration.test.tsx
```diff
@@ -28,7 +28,10 @@ describe('useSWR', () => {
 
   it('should allow functions as key and reuse the cache', async () => {
     function Page() {
-      const { data } = useSWR(() => sharedKey, () => 'SWR')
+      const { data } = useSWR(
```
huozhi, line 31: maybe set the width to larger to keep them as before?

infinite/types.ts
shuding: Yeah it’s easier to read, let’s keep it!

test/use-swr-integration.test.tsx
shuding: I think 80-char is the standard though.

shuding (approved):

infinite/types.ts
huozhi: Seems they'll be remove anyway, no options for parentheses here but only in arrow functions..

PR #1690, opened 2021-12-03
feat: refreshInterval as a function

### LINKED ISSUE #1683: refreshInterval as function
### Discussed in https://github.com/vercel/swr/discussions/1630

<sup>Originally posted by **correttojs** November 11, 2021</sup>
currently the `refreshInterval` option can only be a number. In some cases this number should be dynamic, dependent on the updated fetched data.
For example, the fetcher returns a list of items and a value corresponding to the next expected update. </div>

Andrewnt219:
## Issue
Closes #1683 

## Description

In addition to passing a number, `refreshInterval` can also receive a function that returns a number as an interval. The function has access to the updated data.

Do I need to add anything to the examples folder? I'd like to update the docs as well, but it's in a different repo.

## Test plan

- Function can be passed to `refreshInterval`
- Data passed to `refreshInterval`  is updated data

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -3,7 +3,13 @@ import { defaultConfig } from './utils/config'
 import { SWRGlobalState, GlobalState } from './utils/global-state'
 import { IS_SERVER, rAF, useIsomorphicLayoutEffect } from './utils/env'
 import { serialize } from './utils/serialize'
-import { isUndefined, UNDEFINED, OBJECT, mergeObjects } from './utils/helper'
+import {
+  isUndefined,
+  UNDEFINED,
+  OBJECT,
+  mergeObjects,
+  isFunction
+} from './utils/helper'
 import ConfigProvider from './utils/config-context'
 import { useStateWithDeps } from './utils/state'
 import { withArgs } from './utils/resolve-args'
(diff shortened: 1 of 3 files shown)


src/use-swr.ts
```diff
@@ -444,7 +450,12 @@ export const useSWRHandler = <Data = any, Error = any>(
       // - `force` is true, which is the start of polling
       // - or `timer` is not 0, which means the effect wasn't canceled
       if (refreshInterval && timer !== -1) {
-        timer = setTimeout(execute, refreshInterval)
+        // Use the passed interval
+        // ...or invoke the function with the updated data to get the interval
+        const interval = isFunction(refreshInterval)
+          ? refreshInterval(data)
+          : refreshInterval
+        timer = setTimeout(execute, interval)
```
shuding, line 458: If `interval` is `0` we don't want to continue polling:

```suggestion
        if (interval) {
          timer = setTimeout(execute, interval)
        }
```

Andrewnt219: Ah that's right. I should move the logic outside of the current if to prevent the double check.

Andrewnt219:
Should I add a test for function returns 0?

Andrewnt219:
There's a flaw in the test for updated data, I'll update it soon.

src/use-swr.ts
shuding: Yep that would be great!

shuding (approved):
Perfect, thank you!

shuding:
I think we can add a test for the case of returning `0` (can be another PR of course), this is already looking good to me. Thanks a lot for your contribution!

Andrewnt219:
No problem, I have used swr a lot and it feels great to give back. I'll have a follow up PR for the test soon. 

Do I have to update somewhere in the docs site as well?

shuding:
Thank you! You can update the description for that option here: https://github.com/vercel/swr-site/blob/master/pages/docs/options.en-US.mdx (and update other languages in English too, other contributors can pick up the translation work) 🎉

PR #1691, opened 2021-12-04
test: add refreshInterval function returns 0

Andrewnt219:
## Issues
Related to #1690 

## Description
Add test when `refreshInterval` function returns 0 for the interval. The `refreshInterval` should be disable, which is the same as using `refreshInterval` with a number.

I keep it simple because there are tests for updated data and passing function as an interval. Alternatively, I can make the count goes down from 2 to 0 to demo it is working initially before being disabled.


shuding:
Thank you!

PR #1697, opened 2021-12-08
fix: If `newData` is deeply to the latest state, broadcast the latest state

### LINKED ISSUE #1695: re-render when using fetcher with useSWR

When using useSWR with a fetcher the component will re-render even if the data does not change. 

This is an extension of https://github.com/vercel/swr/discussions/1694. The more I research it is looking like there is room for an improvement.

## Expected Behavior

Component does not re-render if useSWR data does not change.

## Repro Steps / Code Example

fork the Next.js example here https://swr.vercel.app/examples/ssr

change `const { data, error } = useSWR(API);` to `const { data, error } = useSWR(API, fetcher);`

Notice that `console.log("Is data ready?", !!data);` logs to the console twice instead of once.

## Additional Context

If this is intended, maybe a note can be added to the docs stating that using a fetcher will cause a re-render even if data does not change. This seems to be in contrast to the performance docs on Deep Comparison here https://swr.vercel.app/docs/advanced/performance#deep-comparison

icyJoseph:
Attempt to fix #1695

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -263,6 +263,12 @@ export const useSWRHandler = <Data = any, Error = any>(
         // For local state, compare and assign.
         if (!compare(stateRef.current.data, newData)) {
           newState.data = newData
+        } else {
+          // data and newData are deeply equal
+          // it should be safe to broadcast the stale data
+          newState.data = stateRef.current.data
+          // At the end of this function, `brocastState` invokes the `onStateUpdate` function,
+          // which takes care of avoiding the re-render
         }
 
         // For global state, it's possible that the key has changed.
--- a/test/use-swr-loading.test.tsx
+++ b/test/use-swr-loading.test.tsx
@@ -1,5 +1,5 @@
 import { act, screen, fireEvent } from '@testing-library/react'
-import React from 'react'
+import React, { useEffect } from 'react'
 import useSWR from 'swr'
 import {
   createResponse,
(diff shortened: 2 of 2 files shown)


src/use-swr.ts
```diff
@@ -263,6 +263,12 @@ export const useSWRHandler = <Data = any, Error = any>(
         // For local state, compare and assign.
         if (!compare(stateRef.current.data, newData)) {
           newState.data = newData
+        } else {
+          // data and newData are deeply equal
+          // it should be safe to broadcast the stale data
+          newState.data = stateRef.current.data
+          // At the end of this function, `brocastState` invokes the `onStateUpdate` function,
+          // which takes care of avoiding the re-render
         }
```
shuding, line 272: This is a good fix! And I just took a deeper look and it seems the problem is caused by here: https://github.com/vercel/swr/blob/4663154e23a11572733af84b9ef46f8b82644574/src/use-swr.ts#L372

When they're not equal, we broadcast the new data even if the new data is not defined. So another way to fix it is adding:

```js
isUndefined(updatedData) ||
```

to that line of code.

icyJoseph: Do you have any preference? I guess one `if` branch vs `if-else` is more terse?

Also, I commented heavily here, but perhaps can revert the comment?

icyJoseph: I may have misunderstood but

```js
if (isUndefined(newData) || !compare(stateRef.current.data, newData)) {
          newState.data = newData
 }
```

Still requires the `else` block right (the test still fails otherwise)? We need to bind `newState.data` to  `stateRef.current.data`.

shuding: Oh sorry I meant adding it [here](https://github.com/vercel/swr/blob/4663154e23a11572733af84b9ef46f8b82644574/src/use-swr.ts#L372): 

```diff
      setState(
        mergeObjects(
          {
            error: updatedError,
            isValidating: updatedIsValidating
          },
          // Since `setState` only shallowly compares states, we do a deep
          // comparison here.
+         isUndefined(newData) || compare(stateRef.current.data, updatedData)
            ? UNDEFINED
            : {
                data: updatedData
              }
        )
      )
```

I prefer this for now because it is more universal, so we have a way to call `broadCastState()` but skip updating the data, if it's undefined.

icyJoseph: I tried that too, and it caused a couple other tests to failed, iirc. 

Not on my laptop now, but I'll retry first thing in the morning if that's ok. 

I also agree that placing the check there seems more correct.

icyJoseph: The test cases that stop working are:

```shell
● useSWR - local mutation › should be able to mutate data to undefined
● useSWR - local mutation › should be able to mutate data to undefined asynchronously
```

The test description kinda makes it clear why, because when mutating to `undefined`, with the proposed changed we'd be broadcasting the stale state instead.

I tried:

```ts
 isUndefined(updatedData)
            ? { data: UNDEFINED }
            : compare(stateRef.current.data, updatedData)
            ? UNDEFINED
            : {
                data: updatedData
              }
```
But this breaks two tests :sweat_smile: the first one is introduced on this MR, it implicitly reintroduces the initial issue.

```shell
● useSWR - loading › should avoid extra rerenders is the data is the `same` 
● useSWR - configs › should read the config fallback from the context
```

I think it'd be an ok trade-off to patch this now, and investigate further.

shuding: Thank you, yeah that makes total sense!

PR #1698, opened 2021-12-09
fix: Return `undefined` data if key's falsy under suspense mode

### LINKED ISSUE #1660: Suspense and conditional fetching causing infinite rerenders
### Discussed in https://github.com/vercel/swr/discussions/1651

<sup>Originally posted by **rikutiira** November 22, 2021</sup>
Hey, and first of all thanks for the great work on the library! 

I'm new to SWR and I have been reading discussion & GH issues around this, and I'm still a bit unsure about how suspense and conditional fetching should work with the latest version (1.1.0-beta.9 at the time of writing this) so I was hoping to get some clarity around the issue 😄.

I'm having an infinite rerender issue when using suspense with conditional fetching from version 1.0.0 onwards (including current beta).

The [SWR documentation on Suspense](https://swr.vercel.app/docs/suspense#note-with-conditional-fetching) says that with conditional fetching, `data` should be `undefined` but with version 1.0.0+, it seems to fallback to the `Suspense` component instead and causes an infinite render loop where the `useSWR` hook is called. I'm new to suspense, but I suppose this is not expected behaviour?

1) Am I doing something wrong in the code?
2) Is the behaviour expected? If it is, how are you supposed to deal with conditional data fetching? Is there any additional documentation on the subject somewhere?

Here is a Codesandbox where you can test the issue: https://codesandbox.io/s/cold-dust-4dvzl?file=/src/App.js (uncomment `response` to run it). Here's the code as well:

```
import React, { useState, Suspense } from "react";
import useSWR from "swr";
import "./styles.css";

const fetcher = (str) => {
  return new Promise((resolve) => {
    setTimeout(() => {
      resolve(str);
    }, 1000);
  });
};

const Test = (props) => {
  console.log("renders");

  const response = useSWR(props.str ?? null, fetcher, { suspense: true });

  return (
    <>
      <div>Data:</div>
      <pre>{JSON.stringify(response, null, 2)}</pre>
    </>
  );
};

  const [str, setStr] = useState();
  return (
    <div className="App">
      <h1>SWR test</h1>
      <Suspense fallback={<div>loading...</div>}>
        <Test str={str} />
      </Suspense>
      <button onClick={() => setStr("hello world")}>Clicky</button>
    </div>
  );
}
```

Thanks! 🙇 </div>

icyJoseph:
Attempting to fix #1660

### DIFF
--- a/test/use-swr-config.test.tsx
+++ b/test/use-swr-config.test.tsx
@@ -1,12 +1,7 @@
 import { act, screen, fireEvent } from '@testing-library/react'
 import React, { useEffect, useState } from 'react'
 import useSWR, { SWRConfig, useSWRConfig, Middleware } from 'swr'
-import {
-  sleep,
-  renderWithConfig,
-  createKey,
-  renderWithGlobalCache
-} from './utils'
+import { renderWithConfig, createKey, renderWithGlobalCache } from './utils'
 
 describe('useSWR - configs', () => {
   it('should read the config fallback from the context', async () => {
@@ -30,8 +25,7 @@ describe('useSWR - configs', () => {
     await screen.findByText('data: 0')
 
     // wait for the refresh interval
-    await act(() => sleep(INTERVAL * 1.5))
-    screen.getByText('data: 1')
+    await screen.findByText('data: 1')
   })
 
   it('should stop revalidations when config.isPaused returns true', async () => {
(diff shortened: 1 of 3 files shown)


test/use-swr-config.test.tsx
```diff
@@ -30,8 +25,7 @@ describe('useSWR - configs', () => {
     await screen.findByText('data: 0')
 
     // wait for the refresh interval
-    await act(() => sleep(INTERVAL * 1.5))
```
icyJoseph, line 33: This was misbehaving locally (flaky test, failed sometimes, passed others).

Would this be an acceptable swap? it doesn't really check anymore that we are over the INTERVAL time though, which might not be desirable.

shuding: Sounds reasonable!

shuding:
Thank you!

PR #1707, opened 2021-12-15
Make polyfill for `requestAnimationFrame` more robust

thomaspaulmann:
Many @raycast community members use `swr` to have a great caching solution for their extensions. Unfortunately they ran into a few issues that they patch in their extensions codebase right now. After chatting to @shuding, we thought it's best to fix this directly in the `swr` library to guarantee easy adoption. This shouldn't impact any other JS environments.

**Problem**
The polyfill for `requestAnimationFrame` relies on a full implementation of the `window` object. Not every JS environment provides this. F.e. we use Node.js in Raycast. In production builds, it doesn't offer a `window` object. However, in `development` builds we polyfill add a minimal window object to add support for `react-devtools`. The `window` object doesn't inlcude an implementation for `requestAnimationFrame` which SWR assumes.

**Solution**
I added an additional check to make sure the `requestAnimationFrame` is available in the JS environment. If not, it gets polyfilled with `setTimeout`. I don't check if `requestAnimationFrame` is a function. I took this as given.

_Note:_ I tried to follow the coding style and added a new utility to check for the availability of `requestAnimationFrame` which is then used in the `env.ts` file. 

Also, I'm not too familiar with the `swr` codebase. Please let me know if you think there would be a better way to fix it.


FezVrasta:
For the swr maintainers: is `requestAnimationFrame` required or would a microtask (`Promise.resolve()`) work too?

shuding:
Thank you @thomaspaulmann for opening this PR, and the fix looks good to me! 

@FezVrasta The best option here would be `requestIdleCallback`, but it's not that widely usable so we went with `requestAnimationFrame` and fallback to `setTimeout`. A microtask works too, but `setTimeout` still feels better to me. Because the intention is to not occupy any resource and and wait for React to render as well as finish state updates (which sometimes can be a microtask), before SWR revalidate the data. That's just because we already got something to display, and React's tasks are more important at this moment. So I'd still recommend using `setTimeout` as the fallback instead of `Promise.resolve()`.

shuding:
Thanks again for contributing, it's released as `swr@1.1.2-beta.0` — give it a try.

thomaspaulmann:
Thanks a lot for the super quick merge and the clarification on the microtask.

Issue #1703, opened 2021-12-11
State for last update time


shuding:
It would be helpful to introduce a new state `updatedAt`, which is a timestamp representing the last time we get the data from the fetcher or mutator.

## Use Case

The typical use case for this feature will be showing a last refreshed time of the data in the UI which is critical for its realtimeness.

Without this new state one can only implement it with an effect hook to monitor the data change. However due to the deep comparison of SWR that doesn't work when data remains the same.

icyJoseph:
`updatedAt` then should be equal to `Date.now` right after the concurrent promise for this key resolves?

```ts
      startAt = CONCURRENT_PROMISES_TS[key]
      newData = await CONCURRENT_PROMISES[key]
      newState.updatedAt = Date.now()
```

Of course keeping the `updatedAt` as part of the tracked dependencies state.

And is anyone actively working on this?

PR #1708, opened 2021-12-16
feat: Keep track `updatedAt` as part of the SWR state

icyJoseph:
Seeding work to keep track of the `updatedAt` time.  #1703 

At the top of my head I can think of a few things to do:

- [  ] Are tests exhaustive enough?
- [  ] Updates to other parts of the SWR API (?)
- [  ] Documentation

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -203,6 +203,7 @@ export const useSWRHandler = <Data = any, Error = any>(
         // considered here.
         startAt = CONCURRENT_PROMISES_TS[key]
         newData = await CONCURRENT_PROMISES[key]
+        newState.updatedAt = Date.now()
 
         if (shouldStartNewRequest) {
           // If the request isn't interrupted, clean it up after the
@@ -515,6 +516,10 @@ export const useSWRHandler = <Data = any, Error = any>(
     get isValidating() {
       stateDependencies.isValidating = true
       return isValidating
+    },
+    get updatedAt() {
+      stateDependencies.updatedAt = true
+      return stateRef.current.updatedAt
     }
   } as SWRResponse<Data, Error>
 }
--- a/test/use-swr-updated-at.test.tsx
+++ b/test/use-swr-updated-at.test.tsx
@@ -1,0 +1,98 @@
+import { act, fireEvent, screen } from '@testing-library/react'
+import React, { useEffect } from 'react'
+import useSWR from 'swr'
+import { createKey, renderWithConfig, sleep } from './utils'
+
+describe('useSWR - updatedAt', () => {
+  it('should be initially undefined', async () => {
+    const key = createKey()
+
+    const fetcher = () => {
+      return 'data'
+    }
+
+    function Page() {
+      const { mutate, updatedAt } = useSWR(key, fetcher, {
+        revalidateOnMount: false
+      })
+
+      return <button onClick={() => mutate()}>data: {updatedAt}</button>
+    }
+
+    renderWithConfig(<Page />)
+
+    screen.getByText('data:')
+  })
+
+  it('should not trigger re-render if not consumed', async () => {
+    const key = createKey()
+
+    const fetcher = () => {
+      return 'data'
+    }
+
+    const renderSpy = jest.fn()
+
+    function Page() {
+      const { mutate } = useSWR(key, fetcher, {
+        revalidateOnMount: false
+      })
+
+      renderSpy()
+
+      return <button onClick={() => mutate()}>data</button>
+    }
+
+    renderWithConfig(<Page />)
+
+    screen.getByText('data')
+
+    fireEvent.click(screen.getByRole('button'))
+
+    expect(renderSpy).toHaveBeenCalledTimes(1)
+  })
+
+  it('should eventually reflect the last time the fetcher was called', async () => {
+    const key = createKey()
+
+    let fetcherCallTime: number
+
+    const fetcher = () => {
+      fetcherCallTime = Date.now()
+      return 'data'
+    }
+
+    const updateSpy = jest.fn()
+
+    function Page() {
+      const { mutate, updatedAt } = useSWR(key, fetcher)
+
+      useEffect(() => {
+        updateSpy(updatedAt)
+      }, [updatedAt])
+
+      return <button onClick={() => mutate()}>data</button>
+    }
+
+    renderWithConfig(<Page />)
+
+    screen.getByText('data')
+
+    fireEvent.click(screen.getByRole('button'))
+
+    expect(updateSpy).toHaveBeenCalledTimes(1)
+
+    expect(updateSpy.mock.calls[0][0]).toBeUndefined()
+
+    await act(async () => {
+      await sleep(0)
+    })
+
+    expect(updateSpy).toHaveBeenCalledTimes(2)
+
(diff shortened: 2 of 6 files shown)


shuding:
Thank you @icyJoseph — the implementation looks great! One more thing I think we should include in the test is reflecting the `updatedAt` state change in other hooks, that are using the same key.

And also, I'd like to hear your idea about the name "updatedAt" here. Should we call it "fetchedAt"? Simply because I've been thinking about this "updatedAt" name for a while and it feels that the `mutate()` calls should be counted as "updates" too. But in some cases their updates are not that easy to track.

icyJoseph:
> Thank you @icyJoseph — the implementation looks great! One more thing I think we should include in the test is reflecting the `updatedAt` state change in other hooks, that are using the same key.

Right, if we have two or more hooks looking at the same key, `updatedAt` ought to be the same in those. I'll write a test for it.

> And also, I'd like to hear your idea about the name "updatedAt" here. Should we call it "fetchedAt"? Simply because I've been thinking about this "updatedAt" name for a while and it feels that the `mutate()` calls should be counted as "updates" too. But in some cases their updates are not that easy to track.

I see, mutate calls can take many shapes, from a simple call to fetch fresh data, to a controlled call where we are in total control of the new data.

I do agree that `fetchedAt` would be more appropriate if the intention is to tell the user when was the last time the fetcher was called, rather than the last time the data was `mutated`. 

Here, `broadcastState` is called twice, `src/utils/mutate.ts` wouldn't it be enough to get `Date.now()` before calling broadcast state in these cases?

shuding:
> Here, broadcastState is called twice, `src/utils/mutate.ts` wouldn't it be enough to get `Date.now()` before calling broadcast state in these cases?

It's a bit tricky, since the `updatedAt` state only exists locally in the hook, not in the cache. So when a `useSWR` hook mounts, it can't know what's the last time the data gets updated.

After some research, I think we should change the structure of the cache into "key → { data, error, isValidating, ... }". Currently it is "data_key → data", "error_key → error", ... which is not very extendable.

icyJoseph:
I agree with your last point there.

Just as an update, over the holidays I wrote a test to check that other hooks consuming `key` should update their `updatedAt` key to the latest. 

For better or worse, the test failed, so I've been making some changes to turn it green.

While I am at, I've come to believe that either`lastFetchedAt` or `fetchedAt` make much more sense.

src/use-swr.ts
```diff
@@ -324,7 +329,17 @@ export const useSWRHandler = <Data = any, Error = any>(
...
+        updatedAtRef.current = Date.now()
+        broadcastState(
+          cache,
+          key,
+          newState.data,
+          newState.error,
+          false,
+          false,
+          undefined,
+          getUpdatedAt()
```
icyJoseph, line 341: Yeah, this turned out to look rather ugly, but I didn't want to change the existing order of parameters, though, updatedAt, arguably has more importance than the rest, as it is consumed by users.

test/use-swr-updated-at.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+    })
+
+    await act(async () => {
+      await new Promise<void>(resolve => window.queueMicrotask(resolve))
+    })
+
+    // Assert that `second`, eventually agrees with `zero`
+    expect(screen.getByTestId('second')).toHaveTextContent(
+      screen.getByTestId('zero').textContent
+    )
```
icyJoseph, line 193: Consider asserting: Both should not be undefined

test/use-swr-updated-at.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+    expect(updateSpy).toHaveBeenCalledTimes(2)
+
+    const updatedAt = updateSpy.mock.calls[1][0]
+    expect(updatedAt).toBeDefined()
+
+    expect(updatedAt).toBeGreaterThanOrEqual(fetcherCallTime)
+  })
+
+  it('should be consistent in all hooks using the same key', async () => {
```
icyJoseph, line 103: Rather long test.

- Four hooks using the same key are at play
- Initially, Page (zero), a Dashboard mock (first), and a mutator are present.
- During initial rendering, `updatedAt` is undefined.
- Eventually, call mutate from Page.
- Expect Page and Dashboard to agree on `updatedAt`
- I am not 100% sure why, but after this I ought to wait on the deduping interval to settle
- After that, mount the second dashboard, configured to revalidate on mount
- Assert that `updatedAt` for `second` is undefined
- Eventually `second` agrees with `zero`
- Trigger an update from the `mutator`
- Values of `updatedAt` should increase
-All hooks should agree on `updatedAt`

test/use-swr-updated-at.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+      await new Promise<void>(resolve => window.queueMicrotask(resolve))
+    })
+
+    // not empty, and not the same as before
+    expect(screen.getByTestId('zero')).not.toBeEmptyDOMElement()
+    expect(screen.getByTestId('zero')).not.toEqual(firstUpdatedAt)
+
+    // sanity check, the value showed by `first` and `second` has increased
+    expect(Number(screen.getByTestId('first').textContent)).toBeGreaterThan(
+      Number(firstUpdatedAt)
```
icyJoseph, line 207: And also greater than the value calculated when `second` is mounted

icyJoseph:
Now I am back from my break. I see I've been making an error on my tests. Mostly the call to mutate, and expecting that calls to it should reflect similar `updatedAt` in the rest of hooks.

> One more thing I think we should include in the test is reflecting the updatedAt state change in other hooks, that are using the same key.

In this case, we are talking purely about calls to `revalidate`, for instance, when a hook mounts with `revalidateOnMount, or when the window is refocused, right?

I'll rewrite the two tests that use `mutate`. I think we are nearly there with this feature.

icyJoseph:
@shuding how do you feel about the PR now? I think we just need to decide a name for the prop:

> It would be helpful to introduce a new state updatedAt, which is a timestamp representing the last time we get the data from the fetcher or mutator.

- `updatedAt`: As you said before, this implies the mutator changes, and those, at the moment, are hard to catch. I did give it a go at some point, and the code got out of control. I believe a global map/table would be necessary to achieve this.
- `lastFetchedAt`: In the current implementation, whenever the re-validate function actually starts a new request, we call `Date.now()` right after the `await` statement, update `this` hook, and broadcast this property to other hooks. 

Other names could be `fetchedAt`, `revalidatedAt` (though the name doesn't imply fetch that happens on mount).

shuding (commented):
I’m back too and thank you @icyJoseph, the implementation now looks excellent to me as well as those tests. 

Regarding the name I think `fetchedAt` is still better. Or, I’d like to give it a try to also include the update time of mutations, and keep the name `updatedAt` (really want to make it polished before releasing as a stable API).

To make that happen, I’ll come up with a PR to change the cache structure to keep all states by one single key. 

Does the plan sound good to you?

icyJoseph:
@shuding Alright, then should we wait on your PR first, and then adapt this branch to that?

nilskaspersson-imtf:
What happened to this PR? I was thinking of writing some similar functionality myself.

benevbright:
sad that it hasn't been implemented.

PR #1714, opened 2021-12-21
type(infinite): export SWRInfinteHook and InfiniteKeyLoader from infinite

houkanshan:
Hi, this PR simply exports two types: `SWRInfinteHook` and `InfinteKeyLoader`.

For `InfiniteKeyLoader` I have this use case.

```ts
const keyLoader: InfiniteKeyLoader = (index, previousPageData) => { /* ... */ };
export const useList = () => {
  const { data } = useSWRInfinite(keyLoader, fetcher);
  // ...
};
export const revalidateList = () => mutate(unstable_serialize(keyLoader));
```

I saw `SWRHook` is exported in `swr` so I added `SWRInfinteHook` as well. Thanks!

### DIFF
--- a/infinite/index.ts
+++ b/infinite/index.ts
@@ -17,23 +17,23 @@ import {
   SWRInfiniteConfiguration,
   SWRInfiniteResponse,
   SWRInfiniteHook,
-  InfiniteKeyLoader,
-  InfiniteFetcher
+  SWRInfiniteKeyLoader,
+  SWRInfiniteFetcher
 } from './types'
 
 const INFINITE_PREFIX = '$inf$'
 
-const getFirstPageKey = (getKey: InfiniteKeyLoader) => {
+const getFirstPageKey = (getKey: SWRInfiniteKeyLoader) => {
   return serialize(getKey ? getKey(0, null) : null)[0]
 }
 
-export const unstable_serialize = (getKey: InfiniteKeyLoader) => {
+export const unstable_serialize = (getKey: SWRInfiniteKeyLoader) => {
   return INFINITE_PREFIX + getFirstPageKey(getKey)
 }
 
 export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
   (
-    getKey: InfiniteKeyLoader,
+    getKey: SWRInfiniteKeyLoader,
     fn: BareFetcher<Data> | null,
     config: Omit<typeof SWRConfig.default, 'fetcher'> &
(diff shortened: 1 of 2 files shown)


huozhi (approved):

infinite/index.ts
```diff
@@ -267,4 +267,10 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
   }) as unknown as Middleware
 
 export default withMiddleware(useSWR, infinite) as SWRInfiniteHook
-export { SWRInfiniteConfiguration, SWRInfiniteResponse, InfiniteFetcher }
+export {
+  SWRInfiniteConfiguration,
+  SWRInfiniteResponse,
+  SWRInfiniteHook,
+  InfiniteKeyLoader,
```
huozhi, line 274: Maybe we should rename to `SWRInfiniteKeyLoader` and `SWRInfiniteFetcher `?

shuding:
Thank you @houkanshan and glad to see you here! Didn't know you are a SWR user :)

infinite/index.ts
houkanshan: thanks! updated in aeaeeaa , this could break the code below, not sure if it's a concern tho.

```ts
import { InfiniteKeyLoader } from 'swr/infinite/dist/infinite/types';
```

houkanshan:
> Thank you @houkanshan and glad to see you here! Didn't know you are a SWR user :)

Thank you for making this library, isn't everyone using SWR? 😛

infinite/index.ts
huozhi: Thanks, we don't guarantee the subpath imports like this work as expected since it could change during iterations. Allowed subpaths exports are marked in package.json exports field.

houkanshan: gotcha, thx!

houkanshan:
@huozhi thanks for editing the PR message. I didn't realise this PR actually did 2 things:
- Export `SWRInfinteHook`, and `SWRInfinteKeyLoader`.
- Rename `InfiniteFetcher` to `SWRInfiniteFetcher`. This will be a breaking change because `InfiniteFetcher` is exported since 1.1.0. I believe it needs to be mentioned in the changelog.

shuding:
Thanks for the heads up! For the breaking part I’ll open a PR to revert it. We can do it in 2.0.

houkanshan:
Sound good, thank you!

shuding:
Fixed with #1728 👍

PR #1729, opened 2021-12-23
feat: Add `populateCache` option to `mutate`

shuding:
This is cherry-picked from #1450. Currently we only support 2 types of `mutate` calls:

1. `mutate()`
2. `mutate(data, revalidate = true)`

This PR extends the second parameter with another option:

```js
mutate(data, {
  populateCache: boolean = true
})
```

If disabled, `data` will not be written to the cache. The `revalidate` behavior can also be controlled here too:

```js
mutate(data, {
  revalidate: boolean = true,
  populateCache: boolean = true
})
```

_Note that this is the initial step for #1450, and other future improvements for optimistic UI (#1721)._

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -142,13 +142,19 @@ export type MutatorCallback<Data = any> = (
   currentValue?: Data
 ) => Promise<undefined | Data> | undefined | Data
 
+export type MutatorOption = {
+  revalidate?: boolean
+  populateCache?: boolean
+}
+
 export type Broadcaster<Data = any, Error = any> = (
   cache: Cache<Data>,
   key: string,
   data: Data,
   error?: Error,
   isValidating?: boolean,
-  shouldRevalidate?: boolean
+  revalidate?: boolean,
+  populateCache?: boolean
 ) => Promise<Data>
 
 export type State<Data, Error> = {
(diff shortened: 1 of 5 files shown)


huozhi (approved):

src/types.ts
```diff
@@ -142,13 +142,19 @@ export type MutatorCallback<Data = any> = (
   currentValue?: Data
 ) => Promise<undefined | Data> | undefined | Data
 
+export type MutatorOption = {
```
huozhi, line 145: export this in core?

huozhi (approved):

huozhi:
Does `revalidate: false + populate: false` equal passing `false` directly? I feel like the `{}` option is similar to `true`. I also didn't come up the case when `revalidte` is `false` but `populateCache` is true.

Is that possible to only have `populateCache` as option?

shuding:
@huozhi The existing behavior doesn't change, so `mutate(data, false)` is equivalent to `mutate(data, { revalidate: false, populateCache: true })`, which disables the revalidation.

And since both `revalidate` and `populateCache` default to `true`, `mutate(data, {})` is the same as `mutate(data, { revalidate: true, populateCache: true })`.

PR #1745, opened 2021-12-26
feat: Optimistic mutation with error rollback

### LINKED ISSUE #157: Improving immediate mutate + later revalidation

shuding:
## Quick Overview

This PR a adds two more options to `mutate`, `rollbackOnError` and `optimisticData`:

```js
mutate(updateUser(user), {
  revalidate: true,
  populateCache: true,
  rollbackOnError: true,
  optimisticData: user,
})
```

Here `updateUser(user)` is a promise or async function to actually do the **remote mutation**, and `optimisticData` is the value for immediate and **optimistic update**.

When `populateCache` is disabled, the result of that remote mutation will not be written to the cache. 

And if `rollbackOnError` is enabled, and the remote mutation errors (the promise rejects), the previously updated optimistic data will be rolled back.

And overall, using the mutate API can avoid race conditions from normal `useSWR` requests.

A detailed example:  https://codesandbox.io/s/swr-basic-forked-k5hps?file=/src/App.js.

## Documentation 

https://github.com/vercel/swr-site/pull/202

## Related Discussions

Resolves #157. Related to #1450.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -142,9 +142,11 @@ export type MutatorCallback<Data = any> = (
   currentValue?: Data
 ) => Promise<undefined | Data> | undefined | Data
 
-export type MutatorOptions = {
+export type MutatorOptions<Data = any> = {
   revalidate?: boolean
   populateCache?: boolean
+  optimisticData?: Data
+  rollbackOnError?: boolean
 }
 
 export type Broadcaster<Data = any, Error = any> = (
@@ -167,7 +169,7 @@ export type Mutator<Data = any> = (
   cache: Cache,
   key: Key,
   data?: Data | Promise<Data> | MutatorCallback<Data>,
-  opts?: boolean | MutatorOptions
+  opts?: boolean | MutatorOptions<Data>
 ) => Promise<Data | undefined>
 
 export interface ScopedMutator<Data = any> {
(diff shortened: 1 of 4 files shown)


SevenOutman:
Hi @shuding, if I'm getting it correct, the CodeSandbox demo shows that the mutation error is retrieved from the `error` property of `useSWR` result, which happens to be the same place to retrieve the fetching error. I find it kind of ambiguous because populating `error` is not like populating `data` which means "`data` will be this value when revalidation completes".

On the other hand, users may tend to display fetching error and mutation error separately, because fetching/revalidating is like a "natural" process of hydrating page content where an error is like what's going on around, while mutation is usually a user action where an error is a feedback to what user attempted to do.

shuding:
@SevenOutman Thanks, that makes perfect sense. We plan to have `onError` and `onSuccess` etc. callbacks in `mutate` with follow-up PRs.

shuding:
One side note is that this will be a breaking behavior, so we will leave it to 2.0.

PR #1773, opened 2022-01-09
Use yarn workspace

huozhi:
- [x] group core libraries
- [x] remove example development 
- [x] update contribution docs
- [x] simplify publish script
- [x] remove yarn link / yalc related tricks


yalc.lock
```diff
@@ -0,0 +1,8 @@
+{
+  "version": "v1",
+  "packages": {
+    "swr": {
+      "signature": "984b0beb41ffe50f87a7a947b8de4e6a"
+    }
+  }
+}
```
shuding, line 8: What is this file for?

test/use-swr-focus.test.tsx
```diff
@@ -195,11 +195,7 @@ describe('useSWR - focus', () => {
     await focusWindow()
     // wait for throttle intervals
     await act(() => sleep(150))
-    // trigger revalidation
-    await focusWindow()
-    await waitForNextTick()
-    await act(() => sleep(100))
-    await screen.findByText('data: 5')
+    await screen.findByText('data: 4')
```
shuding, line 198: Why changing this test?

shuding (commented):
Can we rename the “package” directory to “src”?

package.json
```diff
@@ -97,11 +66,9 @@
...
-    "typescript": "4.4.3",
-    "yalc": "1.0.0-pre.53"
-  },
-  "peerDependencies": {
-    "react": "^16.11.0 || ^17.0.0 || ^18.0.0"
+    "typescript": "4.4.3"
   },
-  "dependencies": {}
+  "dependencies": {
+    "next": "12.0.7"
```
shuding, line 72: Should this be in `devDependencies`?

huozhi:
> Can we rename the “package” directory to “src”?
@shuding Choose `package` because it contains src, package.json and dist files (`package/dist`) like the previous swr package. `src` might be a little bit inaccurate? wdyt

test/use-swr-focus.test.tsx
huozhi: This flaky test failed almost every time..

shuding:
I'm not sure if `package` is the best name here — if the goal is to make development easier with examples, I think we can organize the files in a leaner way.

PR #1775, opened 2022-01-10
Added clear to cache methods

### LINKED ISSUE #1774: Type Cache does not contain the clear method while it exists
# Bug report

I used cache.clear() in project. In the docs it says it exists (and it really does), but doesn't exist in the type definitions

## Expected Behavior

cache.clear() to exist in the type definition

## Repro Steps / Code Example
```tsx
import type { NextPage } from 'next'
import useSWR, { useSWRConfig } from 'swr'

const fetcher = async (url: string) => { return await (await fetch(url)).json() }

const Home: NextPage = () => {
  const { cache } = useSWRConfig()
  const { data } = useSWR("/api/hello", fetcher)
  console.log(data)
  console.log(cache.get("/api/hello"))
  return (
    <div>
      <button onClick={() => { cache.clear(); console.log(cache) }}>Clear Cache</button>
    </div>
  )
}

export default Home

```
Screenshot 2022-01-10 at 15 52 09

## Additional Context

SWR version. 1.1.2

Markos-Th09:
This pull request fixes #1774

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -258,4 +258,5 @@ export interface Cache<Data = any> {
   get(key: Key): Data | null | undefined
   set(key: Key, value: Data): void
   delete(key: Key): void
+  clear(key: Key): void
 }


src/types.ts
```diff
@@ -258,4 +258,5 @@ export interface Cache<Data = any> {
   get(key: Key): Data | null | undefined
   set(key: Key, value: Data): void
   delete(key: Key): void
+  clear(key: Key): void
```
icyJoseph, line 261: With this signature, what would be the difference between `delete` and `clear`? iirc, the cache used here is pretty much a Map? those have a `clear` method that takes no arguments, which removes all `key-value` pairs from the Map.

huozhi:
Thanks for the contribution, please checkout the comment [here](https://github.com/vercel/swr/pull/1480#issuecomment-923726048)

src/types.ts
Markos-Th09: > With this signature, what would be the difference between `delete` and `clear`? iirc, the default cache used here is a Map? those have a `clear` method that takes no arguments, which removes all `key-value` pairs from the Map.
> 
> Edit:
> 
> Relevant code https://github.com/vercel/swr/blob/main/src/utils/config.ts#L41 - I think it would be good to just have `clear(): void`

My bad! You are right. I fixed this with commit c54ce94.

Issue #1768, opened 2022-01-05
Better React DevTools support


golopot:
# Bug report

## Description / Observed Behavior

Hook value in dev tool looks strange.

<img width="173" alt="Observed" src="https://user-images.githubusercontent.com/16285118/148184992-e89f77ab-0ae7-48cf-8cc7-644589d51ded.png">

## Expected Behavior

<img width="120" alt="Expected" src="https://user-images.githubusercontent.com/16285118/148185030-6ed41e1e-6a6d-43c8-b075-c7fb92f34f36.png">

## Repro Steps / Code Example

```jsx
export default function Home() {
  const { data } = useSWR("123", async (_) => {
    return "done";
  });

  return <div>{String(data)}</div>;
}
```
## Additional Context

SWR version: swr@1.1.2

huozhi:
The new react devtool seems parsing nested function calls of the hook, so you might have to click inside to check out the displayed data. We can polish the name to make `eval` more readable

huozhi:
The outer function name is updated in #1779

PR #1779, opened 2022-01-10
polish: display name in devtool

huozhi:
Related #1768

react devtool now parses all the function calls so polish the anonymous function name to a readable one

After
image


shuding:
LGTM, I think in the future we might rename the `useSWRHandler` to something else (not starting with `use`), and remove that `useDebugValue`.

PR #1781, opened 2022-01-12
chore: remove unnecessary exports

koba04:
These exports are implementation details of each hook, so it would be unnecessary to export them.


huozhi (commented):
I remember there was a reason to export them to let user do `import { infinite } from 'swr/infinite'` to compose with other middleware together then pass to useSWR.

koba04:
@huozhi That makes sense, thank you!

shuding:
We sometimes keep things undocumented to avoid abusing, and we will still have a chance to polish them 😄

PR #1792, opened 2022-01-14
chore: Clean up configurations

shuding:
- .eslintignore can be merged into .eslintrc
- .npmignore is already a subset of .gitignore
- .yarnrc is not needed as SWR has only dev dependences, safe to use the `^` semver matcher
- .prettierrc can be merged into package.json

Also improved package.json a bit.

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,2 +1,0 @@
-save-prefix ""
-
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,0 @@
-dist/
-node_modules
-scripts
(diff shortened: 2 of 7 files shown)


.yarnrc
```diff
@@ -1,2 +0,0 @@
-save-prefix ""
```
huozhi, line 1: how do we deal with save exact version behavior?

shuding: We can change it manually inside package.json

huozhi (approved):

PR #1803, opened 2022-01-23
fix: use the latest reference of fetcher with suspense mode

### LINKED ISSUE #1776: `useSWRInfinite`, first update of key doesn't trigger the proper data fetching
# Bug report

I have a use case of `useSWRInfinite` where I add filters (as query params) to my paginated request.

I would require each change in filters to trigger a full revalidation of data.

Because my swr key changes when I change my filters, I expected I would require nothing else to implement the use case.

Surprisingly, `useSWRInfinite`, does not properly refetch data after changing query params in my key.
It seems to happen consistently on the first key update.

Am I missing something here?

## Expected Behavior

I expected that any update to the swr key would trigger data fetching.

## Repro Steps / Code Example

I updated the demo sandbox (outdated) that you provide in your documentation.

https://codesandbox.io/s/swr-infinite-forked-uibdc?file=/src/App.js:718-729

Here's the surprising behaviour: 

When changing a query param impacting `useSWRInfinite` - `getKey`, the first occurrence does not consistently trigger an API call

https://user-images.githubusercontent.com/3367393/148785091-0f26047b-7a9d-4616-9639-d703233b61c9.mp4

## Additional Context

SWR version: 1.1.2

koba04:
fixes #1776

This is similar to #1727, but this is for the suspense mode.
#1776 seems to be caused by the stale `getKey`, which is referenced from the stale `fetcher` function.
This mutates the fetcher and config ref objects in rendering, but I think it's not a desirable way, so I should probably find out another way.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -510,6 +510,9 @@ export const useSWRHandler = <Data = any, Error = any>(
   // If there is no `error`, the `revalidation` promise needs to be thrown to
   // the suspense boundary.
   if (suspense && isUndefined(data) && key) {
+    // Always update fetcher and config refs even with the Suspense mode.
+    fetcherRef.current = fetcher
+    configRef.current = config
     throw isUndefined(error) ? revalidate(WITH_DEDUPE) : error
   }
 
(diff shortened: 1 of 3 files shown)


icyJoseph:
Linking to #1806 which describes a similar issue

shuding:
Thank you, this is a great catch!

koba04:
Thank you!

shuding:
This fix is included in the `swr@1.2.0` stable release!

PR #1807, opened 2022-01-25
Fix/stale fetcher on suspense

### LINKED ISSUE #1806: Fetcher functions seems to be cached by useSWR(?)
### Discussed in https://github.com/vercel/swr/discussions/1805

<sup>Originally posted by **westmark** January 25, 2022</sup>
_v 1.1.2_

Hi! I have the weirdest bug. Consider this custom hook

```typescript
const useCompanyTeams = (
  companyId: Maybe<string>,
  config: SWRConfiguration = {},
) => {
  const api = useApi<CompaniesApi>('CompaniesApi');

  const rand = Math.random().toString(32).substring(2);
  console.log(rand, companyId, companyId ? companyTeamsKey(companyId) : null);

  return useSWR<Array<Team>>(
    companyId ? companyTeamsKey(companyId) : null,
    (key) => {
      console.log(
        'Calling',
        rand,
        companyId ? companyTeamsKey(companyId) : null,
        key,
      );
      return getCompanyTeams(companyId, api);
    },
    {
      suspense: true,
      ...config,
    },
  );
};
```

(`useApi` returns a generated API with all paths already baked in, and the random value is only for debugging purposes)

During page load, this hook might be called with `companyId` set to `null`, which should yield a null key leading to no data being fetched. Here is the output from the application

```
> 39ftoqmq8ng null null
> 9f0aol3gcgg undefined null
> hass14fndqg null null
> msf3pvqppk8 e6f1a6ba-febf-433c-bb94-eeeae26f6e6b /companies/e6f1a6ba-febf-433c-bb94-eeeae26f6e6b/teams
> calling bl8s3lb5j2o null  /companies/e6f1a6ba-febf-433c-bb94-eeeae26f6e6b/teams
```

The last line above is interesting. 

1. The `rand` variable is logged immediately after its creation, but `bl8s3lb5j2o` which is printed INSIDE the fetcher closure does not appear above, which it should. 
2. Furthermore, `companyId` is `null` inside the closure, but the `key` argument is valid. 

I can't explain 1) at all. This only happens if the hook is invoked with `companyId` set to null before it becomes non-null. If invoked with a non-null value the first time, everything will work and the log will be what your would expect. 

2) seems to suggest that the fetcher(i.e. the closure) is cached inside SWR. Is it?

Can anyone shine some light on this?</div>

icyJoseph:
Attempts to fix #1806 

I would have hoped the `useIsomorphicLayoutEffect` to take care of this, but it would seem that it needs to be done the good ol' way, by refreshing the ref to the fetcher on every render. 

Though, the tests pass under this change, so it might be correct?

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -69,6 +69,8 @@ export const useSWRHandler = <Data = any, Error = any>(
   // Refs to keep the key and config.
   const keyRef = useRef(key)
   const fetcherRef = useRef(fetcher)
+  fetcherRef.current = fetcher
+
   const configRef = useRef(config)
   const getConfig = () => configRef.current
   const isActive = () => getConfig().isVisible() && getConfig().isOnline()
--- a/test/use-swr-suspense.test.tsx
+++ b/test/use-swr-suspense.test.tsx
@@ -316,4 +316,57 @@ describe('useSWR - suspense', () => {
 
     await screen.findByText('SWR')
   })
+
+  it('should use the latest available fetcher closure', async () => {
+    const fetcherSpy = jest.fn()
+
+    const useData = (dataKey: string | null) => {
+      return useSWR(
+        dataKey,
+        () => {
+          fetcherSpy(dataKey)
+          return createResponse('SWR')
+        },
+        { suspense: true }
+      )
+    }
+
+    function Section({ dataKey }: { dataKey: string | null }) {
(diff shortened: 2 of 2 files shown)


icyJoseph:
I think #1803 does a more complete fix of this issue. Closing in favor of that PR.

PR #1816, opened 2022-01-27
`shouldRetryOnError` accepts a function that can be used to conditionally stop retrying  

sairajchouhan:
As discussed in #1574 `shouldRetryOnError` accepts a function and we can stop retrying if some condition fails.

This is useful in case we want to stop retrying on some logic, like if the status code is `401` but still want to have Exponential backoff algorithm to retry for other errors

Currently, if you provide `onErrorRetry` in `SWRConfig` you will essentially override the default `onErrorRetry` which will also stop using exponential backoff for retrying other errors

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -42,7 +42,7 @@ export interface PublicConfiguration<
   revalidateOnReconnect: boolean
   revalidateOnMount?: boolean
   revalidateIfStale: boolean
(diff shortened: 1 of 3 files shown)


shuding:
This is a great PR, thank you!

sairajchouhan:
Thank you

PR #1818, opened 2022-01-28
feat: Support `populateCache` as a function

### LINKED ISSUE #1817: Support `populateCache` as a function
We added the `populateCache` option (as well as other options for optimistic UI) recently, a normal use case looks like the following:

```js
mutate(addTodo(todo), {
  optimisticData: [...todos, todo],
  populateCache: true,
  rollbackOnError: true,
  revalidate: false,
})
```

Say that the `addTodo` calls the `POST /api/todos` endpoint, which returns the full, updated list.

## Transform Before Populating the Cache

However, it's also common that the `POST /api/todos` endpoint only returns the new added item, not the full list. So naturally ​the next step is to support `populateCache` as a synchronous function that transforms the returned data before writing back to the cache:

```js
mutate(addTodo(todo), {
 ​optimisticData: [...todos, todo],
 ​populateCache: addedTodo => [...todos, addedTodo],
 ​rollbackOnError: true,
 ​revalidate: true,
})
```

Note that `revalidate: true` is needed now. 

Also `populateCache` is synchronous only to avoid possible race conditions.

## Opt-out

With this design, if an error is throw when calling `populateCache`, we can skip populating the cache for this specific mutation.

This enables a way to conditionally populating the cache.

shuding:
This PR adds the support of using a function as `populateCache`, to transform the mutation result to the new data for cache population:

```js
mutate(mutation, {
  populateCache(mutationResult, currentData) {
    return updatedData
  }
})
```

For example, if the `POST /todos` API only returns the successfully added item, rather than the full list, we can do this to achieve the cache population and optimistic UI:

```js
mutate(addTodo(newTodo), {
  revalidate: false,
  optimisticData: [...todos, newTodo],
  populateCache: (addedTodo, currentTodos) => {
    return [...currentTodos, addedTodo]
  }
})
```

Closes #1817.


src/utils/mutate.ts
```diff
@@ -22,7 +22,9 @@ export const internalMutate = async <Data>(
     typeof _opts === 'boolean' ? { revalidate: _opts } : _opts || {}
 
   // Fallback to `true` if it's not explicitly set to `false`
-  let populateCache = options.populateCache !== false
+  let populateCache = isUndefined(options.populateCache)
+    ? true
```
huozhi, line 26: default populateCache can be a function then we don't need to check it later

```suggestion
    ? (v => v)
```

huozhi (commented):
If `populateCache` can be a function return data itself like `v => v` then we don't need to check if it's a function any more.

The only way to disable it is setting it to `false` and passing function is just a new way to extend it

huozhi (approved):

PR #1825, opened 2022-01-29
Use cjs for commonjs exports

### LINKED ISSUE #1822: [Bug] 1.2.0 breaks Nextjs/Yarn 3 PnP/Zero Install
# Bug report

Hate to open this! I love `swr` and it usually works great. But I just did an upgrade and it broke my build 😢 [Repo branch that repros it](https://github.com/dgattey/dg/tree/broken/swr).

I upgraded from 1.1.2 to 1.2.0 and my NextJS/Yarn 3 project with PnP and zero install active broke. `yarn next dev` or `yarn next build` errors because it can't find the `swr` module at all as this shows:
image

I think I narrowed it down to [this commit](https://github.com/vercel/swr/commit/c6bf094ce3c6d3fc33ecdc693600fbf6175ada8b). It changes the exports swr provides and landed between beta 0 and beta 1 of 1.2.0. Consequently, installing version 1.2.0-beta.0 works perfectly, but version 1.2.0-beta.1 breaks it (albeit with a slightly different error, but still one about not being able to resolve swr). 

Getting this error with swr 1.2.0:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'swr' imported from /Users/dgattey/checkouts/dg/.next/server/pages/index.js
```

## Expected Behavior

For SWR to be able to be resolved into a real package location.

## Repro Steps / Code Example

Check out the [broken/swr branch of this repo](https://github.com/dgattey/dg/tree/broken/swr). Happy to give access if useful/explain anything.

huozhi:
Fixes #1822

### DIFF
--- a/immutable/package.json
+++ b/immutable/package.json
@@ -1,7 +1,7 @@
 {
   "name": "swr-immutable",
   "version": "0.0.1",
-  "main": "./dist/index.js",
+  "main": "./dist/index.cjs",
   "module": "./dist/index.esm.js",
   "types": "./dist/immutable",
   "exports": "./dist/index.mjs",
--- a/immutable/package.json
+++ b/immutable/package.json
@@ -1,10 +1,13 @@
 {
   "name": "swr-immutable",
   "version": "0.0.1",
-  "main": "./dist/index.cjs",
+  "main": "./dist/index.js",
   "module": "./dist/index.esm.js",
   "types": "./dist/immutable",
-  "exports": "./dist/index.mjs",
+  "exports": {
+    "import": "./dist/index.mjs",
+    "require": "./dist/index.cjs"
+  },
   "peerDependencies": {
     "swr": "*",
     "react": "*"
--- a/infinite/package.json
+++ b/infinite/package.json
@@ -1,10 +1,13 @@
 {
   "name": "swr-infinite",
   "version": "0.0.1",
-  "main": "./dist/index.cjs",
+  "main": "./dist/index.js",
   "module": "./dist/index.esm.js",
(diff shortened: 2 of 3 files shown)


shuding:
My only concern is if threre's some bundler that doesn't support extensions other than `.js`...

huozhi:
> My only concern is if threre's some bundler that doesn't support extensions other than `.js`...

At least we need to move forward with the new standard. Now `esm.js` is the fallback for `.mjs`, and `.js` is the fallback for `.cjs`, containing all of them in the publish assets to fit the bundler whos don't support it is wired, which makes swr keep staying behind the new standard.

There's some bundler like metro (for RN) seems doesn't support cjs nor the exports field so far, and the [issue](https://github.com/facebook/metro/issues/670) is not updated. It's uncertain if they're going to support it or not (probably not).
However expo bundler supports them pretty well so it works if you play RN with expo.

One idea is to learn from RN community:
* if there's any possibility to switch bundler?
* how much the rough percentage is metro taking so far?

Then could propbably make better decisions to support it or drop it.

shuding:
> At least we need to move forward with the new standard.

I agree this is important, however we have to maintain backward compatibility as well, if possible. I'd personally choose "not broken" over "featureful".

huozhi:
Then maybe we could test with 3 extensions together `js`, `esm.js`, `cjs`, `mj` together for now. And decide a time to drop the `esm.js` and `js` together with announcement

huozhi:
Since it's the yarn 3 issue we can close the inproper fix

PR #1837, opened 2022-02-03
refactor: revalidateIfStale has an effect on updates, not only mounting

koba04:
Please correct me if I'm wrong 🙏

The documentation of `revalidateIfStale` says that "automatic revalidation on mount even if there is stale data", but according to the current implementation, the `revalidateIfStale` option has an effect on updates, doesn't it?

https://swr.vercel.app/docs/options#options

If so, I'll send a PR for that.

In this PR,  I made three changes that I've noticed while reading the implementation.

1. The [`shouldRevalidateOnMount`](https://github.com/vercel/swr/blob/c63cafcb00f78af1d16ef5649d7c26d396c46de9/src/use-swr.ts#L446) function has an effect on updates, not only mounting. So I feel `shouldRevalidate` is the appropriate name for the function.

2. In the `shouldRevalidate` function, the `revalidateOnMount` option should be changed only when mounting, so I've added the condition.

3. The `keyChanged` variable should be `true` when `key` has been changed or initial mount. `key` is always changed when mounting, so `key !== keyRef.current` meets the condition.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -84,12 +84,15 @@ export const useSWRHandler = <Data = any, Error = any>(
   const info = cache.get(keyInfo) || {}
   const error = info.error
 
+  const isInitialMount = !initialMountedRef.current
+
   // - Suspense mode and there's stale data for the initial render.
   // - Not suspense mode and there is no fallback data and `revalidateIfStale` is enabled.
   // - `revalidateIfStale` is enabled but `data` is not defined.
-  const shouldRevalidateOnMount = () => {
+  const shouldRevalidate = () => {
     // If `revalidateOnMount` is set, we take the value directly.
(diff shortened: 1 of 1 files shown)


koba04:
I've created a demo for this.
https://codesandbox.io/s/awesome-butterfly-6gkrr

shuding (approved):
Thank you for the refactoring, and totally agreed with all the 3 points you raised! However I’m not seeing anything wrong in your demo (it should be the expected behavior for `revalidateIfStale`), maybe I’m missing something?

shuding:
Oh I just saw your PR for the docs – good catch!

koba04:
Thank you!
I've marked this as `refactor`, but I've changed a current behavior of `revalidateOnMount`. `revalidateOnMount` now affects only on mount, so it might be better to add this into the changelog of the next release.

PR #1847, opened 2022-02-08
test: add a test for the behavior of revalidateOnMount when the key has been changed

koba04:
This is a follow-up PR of #1837.
I've added a test that `revalidteOnMount: false` should call the fetcher function when the key has been changed.


shuding:
Thanks!

koba04:
Thank you!

PR #1851, opened 2022-02-10
bugfix: make suspense and revalidateIfStale work together

simowe:
# Summary
Today, `revalidateIfStale` is ignored when using `suspense: true`.

This fix makes `revalidateIfStale` behave consistently with or without `suspense`

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -97,13 +97,14 @@ export const useSWRHandler = <Data = any, Error = any>(
     // If it's paused, we skip revalidation.
     if (getConfig().isPaused()) return false
 
-    return suspense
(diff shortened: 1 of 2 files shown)


shuding (commented):
Thanks! The fix looks good to me, would you mind also adding a test case to ensure it doesn't break in the future?

simowe:
> Thanks! The fix looks good to me, would you mind also adding a test case to ensure it doesn't break in the future?

Done!

simowe:
> Looks great, thank you!

No problem!

pkellner:
Does this mean the Suspense will be supported in V2? I got the feeling that there is a lot more to implementing suspense then is indicated here based on the discussions we've had.

PR #1861, opened 2022-02-18
feat: support functional optimisticData

### LINKED ISSUE #1858: Support `optimisticData` as a function
### Discussed in https://github.com/vercel/swr/discussions/1850

<sup>Originally posted by **ludovic-montel** February  9, 2022</sup>
First of all, thanks a lot for this library. I've been using it for almost a year, and more intensively for several months, and I love it!

Il like to create optimistic user interfaces, and the new `optimisticData` option with its auto-rollback behavior is great!

I'd like to push it a little further: could you `optimisticData` as a function ? I think it would be consistent with the `mutate` first argument and with #1817.

This function would take the cached data as parameter, and would allow:
- using the global `mutate` function
- extracting and test a pure function that makes the mutation I want, depending on the current data and the operation that is processed

Again, thanks a lot for your work</div>

huozhi:
Resolves #1858

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -145,7 +145,7 @@ export type MutatorCallback<Data = any> = (
 export type MutatorOptions<Data = any> = {
   revalidate?: boolean
   populateCache?: boolean | ((result: any, currentData: Data) => Data)
-  optimisticData?: Data
+  optimisticData?: Data | ((data?: Data) => Data)
   rollbackOnError?: boolean
 }
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -145,7 +145,7 @@ export type MutatorCallback<Data = any> = (
 export type MutatorOptions<Data = any> = {
   revalidate?: boolean
   populateCache?: boolean | ((result: any, currentData: Data) => Data)
-  optimisticData?: Data | ((data?: Data) => Data)
+  optimisticData?: Data | ((currentData?: Data) => Data)
   rollbackOnError?: boolean
 }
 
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -145,7 +145,7 @@ export type MutatorCallback<Data = any> = (
 export type MutatorOptions<Data = any> = {
   revalidate?: boolean
   populateCache?: boolean | ((result: any, currentData: Data) => Data)
-  optimisticData?: Data
+  optimisticData?: Data | ((data?: Data) => Data)
```
shuding, line 148:
```suggestion
  optimisticData?: Data | ((currentData?: Data) => Data)
```

Let's name this as `currentData` so TypeScript will give better intellisense.

djfarly:
Hej @shuding @huozhi,

looks like this merge was an hour late to make it into [1.2.2](https://github.com/vercel/swr/releases/tag/1.2.2) any chance to release this? Would be greatly appreciated! ❤️

shuding:
It’s included in 1.3.0 🎉

src/types.ts
```diff
@@ -145,7 +145,7 @@ export type MutatorCallback<Data = any> = (
 export type MutatorOptions<Data = any> = {
   revalidate?: boolean
   populateCache?: boolean | ((result: any, currentData: Data) => Data)
-  optimisticData?: Data
+  optimisticData?: Data | ((currentData?: Data) => Data)
```
veewee, line 148: Shouldn't this be `((currentData?: Data) => Data|undefined)`

PR #1873, opened 2022-02-26
stopped using bind in unnecessary areas

Cut0:
Remove `bind` in unnecessary areas.


shuding (commented):
Thanks for your contribution! We use `.bind` as it's currently shorter after minification which generates smaller bundle. Any reason that you prefer anonymous functions over `.bind`?

Cut0:
Thanks for your review!
I thought anonymous functions were more recognizable, but as you said, bundle size should be a priority.
I have learned a great deal.

So, I'll close this PR!

PR #1927, opened 2022-04-16
remove mount check for react18

promer94:
https://github.com/reactwg/react-18/discussions/82

18 example
https://codesandbox.io/s/adoring-ramanujan-kpe2i4?file=/src/App.tsx

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -178,14 +178,27 @@ export const useSWRHandler = <Data = any, Error = any>(
       // new request should be initiated.
       const shouldStartNewRequest = !FETCH[key] || !opts.dedupe
 
-      // Do unmount check for calls:
-      // If key has changed during the revalidation, or the component has been
-      // unmounted, old dispatch and old event callbacks should not take any
-      // effect.
-      const isCurrentKeyMounted = () =>
-        !unmountedRef.current &&
-        key === keyRef.current &&
-        initialMountedRef.current
+      /* 
+         For React 17
+         Do unmount check for calls:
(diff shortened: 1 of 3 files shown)


src/utils/helper.ts
```diff
@@ -1,11 +1,12 @@
+import { version } from 'react'
 export const noop = () => {}
 
 // Using noop() as the undefined value as undefined can possibly be replaced
 // by something else.  Prettier ignore and extra parentheses are necessary here
 // to ensure that tsc doesn't remove the __NOINLINE__ comment.
 // prettier-ignore
 export const UNDEFINED = (/*#__NOINLINE__*/ noop()) as undefined
-
+export const IS_REACT_LEGACY = !version.startsWith('18')
```
huozhi, line 9: experimental versions starts with `0.0.0`, we could check if it has react 18 api
```suggestion
export const IS_REACT_LEGACY = !React.useTransition
```

src/utils/helper.ts
```diff
@@ -1,11 +1,12 @@
+import { version } from 'react'
```
huozhi, line 1:
```suggestion
import React from 'react'
```

src/utils/helper.ts
shuding: `useTransition` is too long, we can use `useId` for this case :)

huozhi: 😂 save byte by byte

shuding: Interestingly someone just mentioned about Preact compatibility: https://github.com/vercel/swr/discussions/1645#discussioncomment-2578618

PR #1928, opened 2022-04-16
Add `isLoading` state and refactor the core

shuding:
This PR removes the `isFallback` state added in #1925, with a new `isLoading` state which covers more common use cases. Basically it's considered as "loading" when there is no **loaded data*** for the current key yet, while we are loading new data.

_*loaded data = data returned by the fetcher, not fallback or laggy data._

Here is a diagram showing the difference between `isLoading` and `isValidating`:

Page 1


koba04:
This diagram is amazing! We should add this in the v2 docs! 😍

Issue #1906, opened 2022-04-06
Simple example with React 18 and Suspense fails with Error


pkellner:
# Bug report

## Description / Observed Behavior

React 18 and SWR fails on the simplest of examples. I've created a sandbox at this URL: https://codesandbox.io/s/long-cdn-qxkc3v?file=/pages/indexSuspense.js that if your browse (to that page), you'll get the error:

`error: The server could not finish this Suspense boundary, likely due to an error during server rendering. Switched to client rendering.`

## Expected Behavior

I expect to see a page rendered just like at the default root which is in the file `/pages/index.js`

## Repro Steps / Code Example

Run the code sandbox, browse to http://.../indexSuspense and you will get the error I observe

## Additional Context

Below is the code from code sandbox along with the associated package.json

```
import { Suspense } from "react";
import useSWR from "swr";

const fetcher = (url) => fetch(url).then((r) => r.json());

function Profile() {
  const { data } = useSWR("/api/cities", fetcher, {
    suspense: true
  });
  return <pre>{JSON.stringify(data, null, 2)}</pre>;
}

export default function Example() {
  return (
    <Suspense fallback={<div>Loading...</div>}>
      <Profile />
    </Suspense>
  );
}
```

package.json

```
{
  "name": "example-react-suspense",
  "version": "0.1.0",
  "private": true,
  "scripts": {
    "build": "next build",
    "dev": "next dev",
    "start": "next start"
  },
  "dependencies": {
    "next": "12.1.4",
    "react": "^18.0.0",
    "react-dom": "^18.0.0",
    "swr": "^1.2.2"
  }
}
```

I've also been testing with the latest canary from next and the latest pre-release from react 18.1 and getting the same errors.

shuding:
@pkellner I just checked your example, and it turns out that since React 18 now supports Suspense on the server side, that `fetch('/api/cities')` will happen on the server. So I saw this error:

> TypeError: Only absolute URLs are supported
    at getNodeRequestOptions ...

which is the root cause. You can't fetch a relative URL (`/api/cities`) on the server side.

In fact, if you change it to an absolute API like `https://api.github.com/repos/vercel/swr` it will SSR correctly.

shuding:
I will keep this issue open until we add more clarification and guidance on Suspense and SSR to the documentation. Let me know if there's any suggestion on this!

pkellner:
[...]

PR #1931, opened 2022-04-16
breaking: Throw error when using suspense on the server side without fallback in React 18

### LINKED ISSUE #1832: SWR should throw when trying to suspend on the server
# Bug report

Server-side streaming with useSWR appears to work, in that you can see the data fetched on the server and sent through script tags later on. However, it isn't currently set up to share data between the server and client, causing a client-side fetch to happen in addition to the server-side fetch. Since nothing breaks or warns, it looks like this is fully working when it's not.

## Expected Behavior

As suggested by @devknoll [here](https://twitter.com/devknoll/status/1487588880299831302), useSWR should throw a friendly error when the the `suspense` flag is on and it tries to suspend on the server, so folks know this isn't yet implemented fully. 

## Repro Steps / Code Example

1. Have the following code: 
index.js
```
<Suspense fallback={<div>Loading profile…</div>}>
    <Profile />
</Suspense>
```

profile.jsx
```
function Profile(props) {
   const data = useSWR(fetcher, '/some/profile', {suspense: true})
   return (
        <div> Profile ID: { data.id } </div>
   );
}

```
2. Turn on the `concurrentFeatures` flag.
3. Check the HTML response in the network tab and see the script tags being sent from the server with the data.
3. See that there are also client-side fetches for the same data in the network tab.

shuding:
Close #1841, close #1832. 

More discussions can be found [here](https://github.com/vercel/swr/issues/1906). In short, libraries like SWR should not support fetching data on the server side as of today.

### DIFF
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -1,7 +1,12 @@
 import { useCallback, useRef, useDebugValue } from 'react'
 import { defaultConfig } from './utils/config'
 import { SWRGlobalState, GlobalState } from './utils/global-state'
-import { IS_SERVER, rAF, useIsomorphicLayoutEffect } from './utils/env'
+import {
+  IS_REACT_LEGACY,
+  IS_SERVER,
+  rAF,
+  useIsomorphicLayoutEffect
+} from './utils/env'
 import { serialize } from './utils/serialize'
 import {
   isUndefined,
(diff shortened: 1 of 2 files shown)


src/use-swr.ts
```diff
@@ -542,6 +547,15 @@ export const useSWRHandler = <Data = any, Error = any>(
   // If there is no `error`, the `revalidation` promise needs to be thrown to
   // the suspense boundary.
   if (suspense && isUndefined(data) && key) {
+    // SWR should throw when trying to use Suspense on the server with React 18,
+    // without providing any initial data. See:
+    // https://github.com/vercel/swr/issues/1832
+    if (!IS_REACT_LEGACY && IS_SERVER) {
```
huozhi, line 553: should we throw whatever data and key is?

shuding: @huozhi As the comment and discussions mentioned, we should only throw **if we are going to** suspend on the server side.

huozhi (approved):

src/use-swr.ts
```diff
@@ -542,6 +547,15 @@ export const useSWRHandler = <Data = any, Error = any>(
...
+      throw new Error(
+        'Fallback data is required to use suspense on the server side.'
```
shuding, line 555:
```suggestion
        'Fallback data is required when using suspense in SSR.'
```

Issue #1846, opened 2022-02-08
Option for custom serialization / hashing function


wouterraateland:
# Bug report

## Description / Observed Behavior

I'm using SWR with Supabase. The fetching logic is similar for each query, so I'd like to use a single fetcher and use the PostgREST query as the SWR key as follows:
```TypeScript
useSWR(supabase.from("table").select("*").eq("id", 1), supabaseFetcher);
```

The problem with this setup, is that `supabase.from("table").select("*").eq("id", 1)` gets serialized into a different key each time. Thus, the request won't be cached and will be retried on every render...

## Possible Solution

It'd be great if the standard `serialize` or `stableHash` function would be extensible.
In this case a simple extension to `stableHash` would do the trick:
```TypeScript
const customHash = (arg: any): string => {
  if (arg instanceof PostgrestFilterBuilder) return arg.url.toString();
  else return stableHash(arg)
}
```

shuding:
How about using the `.url` value as the key?

```js
useSWR(supabase.from("table").select("*").eq("id", 1).url, supabaseFetcher);
```

wouterraateland:
In this case, using the `.url` value as the key would work, but is not ideal. The encapsulating `PostgrestFilterBuilder` provides some nice error handling. By just passing the `.url`, this error handling would have to be re-implemented.

In general, let `O` be an object with property `O.p` that can be serialized in a stable way. Then, I can imagine situations in which it would be hard to reconstruct the functionality of `O` given `O.p`.
Of course, one could implement some sort of lookup table to find `O` given `O.p`, but that would require extra code.
I therefor think that passing `O` as key, with a custom serialization function would be a cleaner solution.

PortableSteve:
> It'd be great if the standard serialize or stableHash function would be extensible.

I've got another case where this would be useful. Sometimes the fetcher returns data that includes some fields that I don't want included in the hash/compare function - for example a timestamp field that always changes. I don't want my component re-rendering if only the timestamp changes. I realise this can be achieved with a custom `compare()` function, but it would be nice to be able to make use of `stableHash()`.

One option might be to provide an optional transform function that takes the result of the fetcher, and returns the data that should be used by the hash function:

```
const transform = (data) => {
   const { timestamp, ...rest } = data;
   return rest;
};
```

Another option might be to expose `stableHash()` so that it could be called from custom `compare()` functions.

[...]

PR #1934, opened 2022-04-17
(WIP) Adding hash config

msdrigg:
Adding a function to the configuration to replace `stableHash`.

This pr is meant to resolve issues discussed in #1846. 

It adds a hash function in the configuration that defaults to stableHash.

I haven't written any tests for this, but I have tested this on my personal repository and seen that it works. Can anyone please take a look at this and give me some direction on how to make this ready to merge? 

E.g.:

-  what tests should I write?
- Am I missing any places I should put the default configuration
- This changes the signature of `serialize`. Should I make the hash function in `serialize` optional and make it defualt to `stableHash` internally?

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -8,7 +8,7 @@ export type BareFetcher<Data = unknown> = (
 export type Fetcher<
   Data = unknown,
   SWRKey extends Key = Key
-> = SWRKey extends () => infer Arg | null | undefined | false
+  > = SWRKey extends () => infer Arg | null | undefined | false
   ? (args: Arg) => FetcherResponse<Data>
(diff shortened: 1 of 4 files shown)


msdrigg:
Seeing the errors, and this needs some work. Will resolve issues next week

steinroe:
Do you need any support here? :)

msdrigg:
I have kind of stalled out here. Having a busy week with work and I am out of town this weekend. If anyone wants to take over this, I would not mind at all because I won't be able to get this resolved for at least another week.

There is a lot of work that needs to be done to change the signature of serialize. It might be a better approach to make serialize overridable rather than customHash, but I am not sure if that would work.

PR #1936, opened 2022-04-20
Add missing clear method to Cache type definition

Vinnl:
The Cache has been exposed since [1], which supposedly includes a
.clear() method [2], which in turn is not actually present in the
type definition [3]. This adds it to the type definition to align
with the implementation.

Since Map also contains a .clear() method [4], this shouldn't break
the ability to pass that in as a map.

[1] https://github.com/vercel/swr/pull/231
[2] https://github.com/vercel/swr/issues/161#issuecomment-695417107 and https://swr.vercel.app/docs/advanced/cache#access-to-the-cache
[3] https://github.com/vercel/swr/issues/161#issuecomment-1079198998
[4] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/clear

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -264,4 +264,5 @@ export interface Cache<Data = any> {
   get(key: Key): Data | null | undefined
   set(key: Key, value: Data): void
   delete(key: Key): void
+  clear(): void
 }
--- a/test/use-swr-cache.test.tsx
+++ b/test/use-swr-cache.test.tsx
@@ -260,7 +260,8 @@ describe('useSWR - cache provider', () => {
             }
             return v
           },
-          delete: k => parentCache_.delete(k)
+          delete: k => parentCache_.delete(k),
+          clear: () => parentCache_.clear()
         }
       }
     })


shuding:
We can just add this, but I'd love to hear more feedback about **why** there's a need to clear the cache (and why it can't be done via changing the cache provider). 

This is because manually updating the cache is dangerous and can lead into inconsistency in the UI, so we're very cautious here about it.

Vinnl:
@shuding Sure thing! In my case, it's because we log the user out by forgetting their auth tokens, after which future API requests fail. At that point, we want to make sure that no user data is visible anymore. (Unfortunately, this is only one of the authentication methods, used in development; we don't always have a user-specific string that we can add to the cache keys.)

babldev:
Unclear to me what the best approach is (replacing cache provider vs. a `.clear()`), but I think it's worth having some sort of documented path for "the user logged out so we want to clear cached responses." Happy to help write-up docs for that for the [Cache](https://swr.vercel.app/docs/advanced/cache) page if that is the agreed upon best practice.

ezhlobo:
The case when we log in/out is very common for apps. And usually when we logged in we work with private endpoints which leads us to the state with private data cached and accessed by users even when they are not supposed to have it. And there are more cases, for example with 'policy' pattern based on data from endpoints.

I'm fine with the idea to create and maintain global object for the app and pass it as custom provider, however as the case common for apps it would be great to have a way to do so out of the box.

So I support these changes by @Vinnl as clean and robust way to do hard reset.

beqramo:
Sometimes logout logic isn't in hooks so we need a static function/method to call.

piotr-cz:
(Re)adding method to an interface is a breaking change, I suggest this should come in v2.0

piotr-cz:
The `clear` method has been removed from docs.

But since [2.5.0](https://github.com/vercel/swr/releases/tag/v2.5.0) there is an `unload` method (A new API to clear the SWR cache)

huozhi:
Closing per https://github.com/vercel/swr/pull/1936#issuecomment-5553882006

PR #1937, opened 2022-04-20
types: loose cache types

huozhi:
Makes type of `cache` looser, so user could use any method without inferring or use type casting

x-ref: #161 
x-ref: #1775 
x-ref: #1480 
Closes: #1936


shuding (commented):
> so user could use any method without inferring or use type casting

This is like making it `any`, and the user can easily do something wrong. I think we should be more strict on APIs.

PR #1942, opened 2022-04-27
test: upgrade to jest 28

huozhi:


shuding:
This will be done in #1918 along with testing-library 👍

PR #1945, opened 2022-04-29
Improved React 18 support

shuding:


promer94:
https://github.com/vercel/swr/blob/0223e0de222e5965d14818ccfa346fc2a4b75c15/src/utils/broadcast-state.ts#L22-L25

This could be removed since stateUpdate should be handled by use-sync-external-store

Issue #1899, opened 2022-03-31
useSWRInfinite Mutate function doesn't accept the same arguments as useSWR Mutate


kristiancoulsonplt:
# Bug report

## Description / Observed Behavior

The docs (and typescript types!) show that the mutate function returned by the useSWRInfinite hook works the same as the useSWRHook
Docs:
[https://swr.vercel.app/docs/pagination#return-values](https://swr.vercel.app/docs/pagination#return-values)
Types (extends the existing hooks return values):
[https://github.com/vercel/swr/blob/main/infinite/types.ts#L37](https://github.com/vercel/swr/blob/main/infinite/types.ts#L37)

However if you dive into the code they don't work the same way at all
[https://github.com/vercel/swr/blob/main/src/utils/mutate.ts#L21](https://github.com/vercel/swr/blob/main/src/utils/mutate.ts#L21)

[https://github.com/vercel/swr/blob/main/infinite/index.ts#L186](https://github.com/vercel/swr/blob/main/infinite/index.ts#L186)

I'm assuming that at some point in the past they matched, but more recently the regular hooks mutate got changed, and the types for the inf hook do match the regular one, but when in the code editor, it wasn't able to differentiate that the interface with the new options were invalid

## Expected Behavior

I would assume they would work the same, or that the docs/types would be updated to reflect how they currently are. 

## Repro Steps / Code Example

```js
const { data, mutate } = useSWR('url', fetcher);

// does work
mutate(newData, { revalidate: false });
// does work
mutate(newData, false);
```

```js
const { data, mutate } = useSWRInfinite('url', fetcher);

// doesn't work
mutate(newData, { revalidate: false });
// does work
mutate(newData, false);
```

## Additional Context

Currently on the latest version, it took me so long to work out why this was happening, it kept returning to the old cache without any errors or anything

myxoh:
Tied to this - a problem from this is there's no way of doing an `optimisticUpdate` when using `useInfiniteSwr`

rapiz1:
OK I just realized this has not been released and I got bit by this

transitive-bullshit:
Even on `main`, `swr/infinite` doesn't pass all of the `mutate` options to the underlying `mutate`, including `optimisticData`, `populateCache`, and `rollbackOnError`.

This is exacerbated by the fact that these two versions of `mutate` have the same TS signature, but don't respond to options the same way.

PR #1947, opened 2022-05-01
fix: mutate args for useSWRInfinite hook

sanjeev29:
This PR fixes the bug mentioned in #1899. 

As mentioned in the documentation (https://swr.vercel.app/docs/pagination#return-values), mutate function returned by the useSWRInfinite hook should work the same as the useSWR hook.


shuding (commented):
Thanks! Would you mind adding a test so we won't have regression in the future?

sanjeev29:
@shuding Yes! Will do that.

koba04:
@shuding 
Should the mutate function `useSWRInfinite` returns support the following options as well as the `revalidate` option?

- `optimisticData`
- `populateCache`
- `rollbackOnError`

https://swr.vercel.app/docs/mutation#optimistic-updates

If not, it would be nice to add the supported options to the documentation.

shuding:
Yes, I think these should be supported by this hook too 👍

sanjeev29:
@shuding Cool. Let me make a push for this!

sanjeev29:
@shuding Kinda had a hard time writing a test for this using optimistic updates. Can you scope it out here?

koba04 (approved):

koba04:
The following type also should be updated, but I think it's ok to fix it with [the following PR](https://github.com/vercel/swr/pull/1947#issuecomment-1115011415)
https://github.com/vercel/swr/blob/a3221302b3558a443f8cfa5d6e3dd27bb7c8f138/infinite/index.ts#L179

sanjeev29:
@koba04 I can resolve these conflicts?

shuding:
Thanks!

Issue #1949, opened 2022-05-01
Provider used with conditional fetching causes problems


outloudvi:
# Bug report

## Description / Observed Behavior

```
Uncaught TypeError: can't access property 0, _a is undefined
    useSWRHandler index.mjs:634
    useSWRArgs index.mjs:612
    Article App.js:8
    React 9
    workLoop scheduler.development.js:266
    flushWork scheduler.development.js:239
    performWorkUntilDeadline scheduler.development.js:533
```

## Expected Behavior

No error.

## Repro Steps / Code Example

``` jsx
import useSWR, { SWRConfig } from 'swr'
import { useState, useEffect } from 'react'

const fetcher = (url) => fetch(url).then((r) => r.json())

function Article() {
  const [run, setRun] = useState(false)
  const { data } = useSWR(run ? 'https://httpbin.org/json' : null, fetcher)
  useEffect(() => {
    setRun(true)
  }, [])

  return (
    <h1>
      <div>{data?.slideshow?.author}</div>
    </h1>
  )
}

function App() {
  return (
    <SWRConfig
      value={{
        provider: () => new Map(),
      }}
    >
      <Article />
    </SWRConfig>
  )
}

export default App

```

## Additional Context

SWR version: 1.3.0
Tested browser: Firefox 101.0a1 and Chromium 101.0.4951.41 on Arch Linux

The reason why I'm using SWR in this way is that I'm using it with Next.js, and I hope to configure a LocalStorage-based persistent cache. However, SWR seems to directly return the cached value on the client initial rendering. This makes the rendering different from server side rendering since the server doesn't necessarily has the same cache as the client, and finally spawns some hydration errors. A solution I've thought of is to let the client start loading data inside `useEffect`, which leads to this case.

huozhi:
Looks like the custom cache has been recreated so that swr cannot access the original one. A workaround to get rid of this is keep same provider instance in each render. e.g. like using `useRef`

```js
function App() {
  const m = useRef(new Map())
  return (
    <SWRConfig
      value={{
        provider: () => m.current,
      }}
    >
      <Article />
    </SWRConfig>
  )
}
```

huozhi:
Yeah sure, we'll update the docs later to present it with a better approach. Thanks!

Issue #1904, opened 2022-04-05
React 18 in concurrent mode + Cache provider


piotr-cz:
# Bug report

## Description / Observed Behavior

When using React 18 in concurrent mode and a custom [cache provider](https://swr.vercel.app/docs/advanced/cache), an error is thrown:

```
Uncaught TypeError: Cannot read properties of undefined (reading '0')
    at useSWRHandler (index.esm.js:631:1)
```

This happens when using [create-react-app](https://www.npmjs.com/package/create-react-app) v4 or v5 in dev mode (`react-scripts start`), however not when running an actual build (`react-scripts build`).

## Expected Behavior

No error is thrown

## Repro Steps / Code Example

CodeSandbox based on [Basic usage](https://swr.vercel.app/examples/basic) is available here: <https://codesandbox.io/s/eager-alex-zyijly?file=/src/index.js>:

1. Update react and react-dom to 18.0 or 18.1 in dependencies, enable concurrent mode (`ReactDOM.createRoot(rootElement).render(...)`)
2. Add new provider `const provider = () => new Map();`
3. Wrap `<App />` with `<SWRConfig value={{ provider }}></SWRConfig>`

## Additional Context

SWR version: 1.1.0 ... 1.3.0 || 2.0.0-beta.0 || 2.0.0-beta.1
Latest version with no error: 1.0.1

It looks like the `SWRGlobalState` WeakMap in the `useSWRHandler` function doesn't contain custom cache provider, even that it has been successfully added in `initCache` function.

emreycolakoglu:
Hey, this breaks us pretty bad too. Had to create a workaround using middlewares for my use-case (persistance) but it's far from perfect.
Is there a workaround?

piotr-cz:
Any workarounds I know of is downgrading react to v17 or swr to v1.1.
I guess this issue will receive more attention when people will start upgrading react to v18.

shuding:
React 18 support will be our top priority for 2.0 and it’s in the works.

promer94:
This should works in latest beta version now.
[codesandbox](https://codesandbox.io/s/runtime-tree-b2fh33?file=/src/index.js)

piotr-cz:
@promer94 
I cannot completely confirm, as upgrading swr in [my sandbox](https://codesandbox.io/s/eager-alex-zyijly?file=/src/index.js) to 2.0.0-beta.3 as well as downgrading in yours yours leads to
```
ModuleNotFoundError
[...]

PR #1953, opened 2022-05-08
refactor: switch to useSyncExternalStoreWithSelector

promer94:
This PR is intented to solve the 'tearing' problem described in #794 for React 18.
Example for 'tearing' in React 18: https://codesandbox.io/s/concurrent-swr-case-iii-shu-s5fggt

It will also solve the cache initialization problem in StrictMode
#1904 https://codesandbox.io/s/swr-1904-mxfft8
#1949 https://codesandbox.io/s/swr-1949-22d6iw

## limitations
Accroding to the [useSES RFC](https://github.com/reactjs/rfcs/blob/main/text/0214-use-sync-external-store.md#drawbacks)

> to prevent data races, you can't trigger a transition using a store update — updates to the store always render synchronously.

examples:
https://codesandbox.io/s/concurrent-swr-case-i-shu-dckwqj
https://codesandbox.io/s/concurrent-swr-case-ii-shu-3j5j2e


koba04:
This would not be an appropriate place to ask, but currently SWR stores some information like a current key and store into `useRef` objects, which might be a problem with concurrent features as far as I know.

Should SWR migrate the usage of useRef objects like this PR does, or is it safe as long as `useRef` objects are updated in effects run in commit phases like `useEffect` and `useLayoutEffect`?

promer94:
> is it safe as long as useRef objects are updated in effects run in commit phases like useEffect and useLayoutEffect 

As far as i know, updating ref in effect is concurrent safe. 
and we will also use https://github.com/dai-shi/will-this-react-global-state-work-in-concurrent-rendering for testing

test/use-swr-error.test.tsx
```diff
@@ -37,7 +37,7 @@ describe('useSWR - error', () => {
         { onError: (_, errorKey) => (erroredSWR = errorKey) }
       )
       if (error) return <div>{error.message}</div>
-      return <div>hello, {data}</div>
+      return <div><>hello, {data}</></div>
```
huozhi, line 40: why do we adding Fragment here?

huozhi (commented):
Thanks and great work!
could you add more explanation about the links in PR description section?

how about move the `swr/` dir to `core/` or other dir that doesn't have the same name with repo?

test/use-swr-error.test.tsx
promer94: Screen Shot 2022-05-13 at 07 34 10


Because the data is infered as Error here.

shuding (approved):
Thanks, this is incredible work!

PR #1961, opened 2022-05-13
Fix cache types

chibicode:
With the [2.0.0 update](https://github.com/vercel/swr/releases/tag/2.0.0-beta.0) (https://github.com/vercel/swr/pull/1863), each value in `cache` now holds the current states.

### Type changes

- Reuse the existing `State` type whenever necessary
- Remove `null` from a list of possible values of the cache
- Update the type of `createCacheHelper` to be compatible with `infinite/index.ts`
- Update the default type argument of `useSWRConfig`.
- Update the type of `MutatorOptions[`populateCache`] to allow `undefined` for `currentData` on function argument.

### Non-type changes

- `use-swr.ts`: `isLoading` was not being copied over in `onStateUpdate`. This became obvious as I saw that the `State` type contained `isLoading`.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -144,7 +144,7 @@ export type MutatorCallback<Data = any> = (
 
 export type MutatorOptions<Data = any> = {
   revalidate?: boolean
-  populateCache?: boolean | ((result: any, currentData: Data) => Data)
+  populateCache?: boolean | ((result: any, currentData?: Data) => Data)
   optimisticData?: Data | ((currentData?: Data) => Data)
   rollbackOnError?: boolean
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -255,14 +255,12 @@ export type RevalidateCallback = <K extends RevalidateEvent>(
   type: K
 ) => RevalidateCallbackReturnType[K]
 
-export type StateUpdateCallback<Data = any, Error = any> = (state: {
-  data?: Data
-  error?: Error
-  isValidating?: boolean
-}) => void
+export type StateUpdateCallback<Data = any, Error = any> = (
+  state: State<Data, Error>
+) => void
 
-export interface Cache<Data = any> {
-  get(key: Key): Data | null | undefined
-  set(key: Key, value: Data): void
+export interface Cache<Data = any, Error = any> {
+  get(key: Key): State<Data, Error> | undefined
+  set(key: Key, value: State<Data, Error>): void
   delete(key: Key): void
 }
--- a/src/use-swr.ts
+++ b/src/use-swr.ts
@@ -178,7 +178,7 @@ export const useSWRHandler = <Data = any, Error = any>(
       // new request should be initiated.
       const shouldStartNewRequest = !FETCH[key] || !opts.dedupe
 
-      /* 
+      /*
          For React 17
          Do unmount check for calls:
          If key has changed during the revalidation, or the component has been
@@ -442,7 +442,8 @@ export const useSWRHandler = <Data = any, Error = any>(
         mergeObjects(
           {
             error: state.error,
-            isValidating: state.isValidating
+            isValidating: state.isValidating,
+            isLoading: state.isLoading
           },
           // Since `setState` only shallowly compares states, we do a deep
           // comparison here.
--- a/src/utils/cache.ts
+++ b/src/utils/cache.ts
@@ -8,6 +8,7 @@ import * as revalidateEvents from '../constants'
 import {
   Key,
   Cache,
+  State,
   ScopedMutator,
   RevalidateEvent,
   RevalidateCallback,
@@ -106,13 +107,9 @@ export const createCacheHelper = <Data = any, ExtendedInfo = {}>(
 ) =>
   [
     // Getter
-    () => cache.get(key) || {},
+    () => (cache.get(key) || {}) as State<Data, any> & Partial<ExtendedInfo>,
     // Setter
-    (
-      info: Partial<
-        { data: Data; error: any; isValidating: boolean } | ExtendedInfo
-      >
-    ) => {
+    (info: Partial<State<Data, any> | ExtendedInfo>) => {
       cache.set(key, mergeObjects(cache.get(key), info))
     }
   ] as const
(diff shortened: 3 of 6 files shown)


src/types.ts
```diff
@@ -255,14 +255,12 @@ export type RevalidateCallback = <K extends RevalidateEvent>(
   type: K
 ) => RevalidateCallbackReturnType[K]
 
-export type StateUpdateCallback<Data = any, Error = any> = (state: {
-  data?: Data
-  error?: Error
-  isValidating?: boolean
-}) => void
+export type StateUpdateCallback<Data = any, Error = any> = (
+  state: State<Data, Error>
```
chibicode, line 259: Reuses the `State` type, which will add `isLoading` which was missing in the original type definition.

https://github.com/vercel/swr/blob/09d80c525798e85e6fd7e1dbc939b68249ad2b29/src/types.ts#L167-L172

src/types.ts
```diff
@@ -255,14 +255,12 @@ export type RevalidateCallback = <K extends RevalidateEvent>(
...
+) => void
 
-export interface Cache<Data = any> {
-  get(key: Key): Data | null | undefined
```
chibicode, line 265: Removed `null` as it seemed like it's no longer being used.

src/types.ts
```diff
@@ -255,14 +255,12 @@ export type RevalidateCallback = <K extends RevalidateEvent>(
...
-  set(key: Key, value: Data): void
+export interface Cache<Data = any, Error = any> {
+  get(key: Key): State<Data, Error> | undefined
+  set(key: Key, value: State<Data, Error>): void
```
chibicode, line 264: From [release notes](https://github.com/vercel/swr/releases/tag/2.0.0-beta.0):

> So you will have to do the following change to your code, `get`:
> 
> ```diff
> - cache.get(key)
> + cache.get(key)?.data
> ```
> 
> And `set`:
> 
> ```diff
> - cache.set(key, data)
> + cache.set(key, { ...cache.get(key), data })
> ```

src/use-swr.ts
```diff
@@ -442,7 +442,8 @@ export const useSWRHandler = <Data = any, Error = any>(
         mergeObjects(
           {
             error: state.error,
-            isValidating: state.isValidating
+            isValidating: state.isValidating,
+            isLoading: state.isLoading
```
chibicode, line 446: `isLoading` was not being copied over originally, but I believe it should be.

src/utils/cache.ts
```diff
@@ -106,13 +107,9 @@ export const createCacheHelper = <Data = any, ExtendedInfo = {}>(
 ) =>
   [
     // Getter
-    () => cache.get(key) || {},
+    () => (cache.get(key) || {}) as State<Data, any> & Partial<ExtendedInfo>,
```
chibicode, line 110: This was necessary to make `ExtendedInfo` work:

https://github.com/vercel/swr/blob/09d80c525798e85e6fd7e1dbc939b68249ad2b29/infinite/index.ts#L67-L77

I couldn't figure out a way to do it without type assertion.

I did notice that this line had a type error (`Type 'Data' is not assignable to type 'Data[]'`), but I couldn't figure out how to resolve this.

https://github.com/vercel/swr/blob/09d80c525798e85e6fd7e1dbc939b68249ad2b29/infinite/index.ts#L147

src/utils/mutate.ts
```diff
@@ -100,7 +100,7 @@ export const internalMutate = async <Data>(
   if (populateCache) {
     if (!error) {
       // Transform the result into data.
-      if (isFunction(populateCache)) {
+      if (isFunction(populateCache) && originalData) {
```
chibicode, line 103: Without this check, there was this error:

```
Argument of type 'Data | undefined' is not assignable to parameter of type 'Data'.
```

This is because `populateCache` expects the arg to be `Data`,

https://github.com/vercel/swr/blob/dfcba4d63026bee6b2994a5c23ab6fdc2a327263/src/types.ts#L147

but `originalData` can be `undefined`, as we're getting it from the cache:

https://github.com/vercel/swr/blob/dfcba4d63026bee6b2994a5c23ab6fdc2a327263/src/utils/mutate.ts#L55

However, not sure if this will break something.

src/utils/use-swr-config.ts
```diff
@@ -5,7 +5,7 @@ import { mergeObjects } from './helper'
 import { FullConfiguration, Cache } from '../types'
 
 export const useSWRConfig = <
-  T extends Cache = Map<string, any>
+  T extends Cache = Cache
```
chibicode, line 8: Should this use `Cache<Map<string, any>>` instead? I didn't think that was necessary, but I'm not sure.

chibicode: I guess this will be a breaking change as well (for those supplying a custom type in `useSWRConfig`)? Maybe we should update release notes?

huozhi: We added it recently only for v2, so I think it's fine

chibicode: Apparently this change was added recently — not sure why it was using `Map<string, any>` https://github.com/vercel/swr/pull/1938

huozhi: The previous cache type in v1 is `Map<string, any>`, it's good to update it now 👍

huozhi (approved):

src/utils/mutate.ts
chibicode: @shuding suggested to change the argument type of `populateCache` to accept `undefined` instead, closing this comment.

src/types.ts
```diff
@@ -144,7 +144,7 @@ export type MutatorCallback<Data = any> = (
 
 export type MutatorOptions<Data = any> = {
   revalidate?: boolean
-  populateCache?: boolean | ((result: any, currentData: Data) => Data)
+  populateCache?: boolean | ((result: any, currentData?: Data) => Data)
```
chibicode, line 147: This was a suggested fix for this issue in `src/utils/mutate.ts`: https://github.com/vercel/swr/pull/1961#discussion_r872266013

PR #1962, opened 2022-05-13
Improved React 18 support

promer94:
This PR refactors the SWR core with `useSyncExternalStore` and `startTransition`, also moved utils into a shared sub-import.


shuding (approved):
The `_internal` change increases the overall bundle size a lot caused by named exports which can't be mangled. But I will address these later with a follow-up PR.

_internal/types.ts
```diff
@@ -255,12 +265,22 @@ export type RevalidateCallback = <K extends RevalidateEvent>(
...
+  set(key: Key, value: Data): void
   delete(key: Key): void
 }
+
+export interface CacheValue<Data = any, Error = any> {
+  data?: Data
+  error?: Error
+  isValidating?: boolean
+  isLoading?: boolean
+}
```
chibicode, line 279: `CacheValue` is identical to `State` (see https://github.com/vercel/swr/pull/1961/files#r872242431)

https://github.com/vercel/swr/blob/9a0328fffdeb8ca0eaf42bc88f318f7e48a3b245/_internal/types.ts#L177-L182

_internal/types.ts
```diff
@@ -255,12 +265,22 @@ export type RevalidateCallback = <K extends RevalidateEvent>(
...
-export type StateUpdateCallback<Data = any, Error = any> = (
-  state: State<Data, Error>
-) => void
-
-export interface Cache<Data = any, Error = any> {
-  get(key: Key): State<Data, Error> | undefined
-  set(key: Key, value: State<Data, Error>): void
+export interface Cache<Data = any> {
+  get(key: Key): CacheValue<Data> | undefined
+  set(key: Key, value: Data): void
```
chibicode, line 270: The type of `set` was updated in https://github.com/vercel/swr/pull/1961/files#r872245634 - seems like a regression.

_internal/utils/use-swr-config.ts
```diff
@@ -2,10 +2,10 @@ import { useContext } from 'react'
 import { defaultConfig } from './config'
 import { SWRConfigContext } from './config-context'
 import { mergeObjects } from './helper'
-import { FullConfiguration, Cache } from '../types'
+import { FullConfiguration, Cache, CacheValue } from '../types'
 
 export const useSWRConfig = <
-  T extends Cache = Cache
+  T extends Cache = Map<string, CacheValue>
```
chibicode, line 8: I believe we no longer want to use `Map`: https://github.com/vercel/swr/pull/1961/files#r872298911

test/type/config.ts
```diff
@@ -1,11 +1,14 @@
...
 interface CustomCache<Data = any> extends Cache<Data> {
   reset(): void
 }
 
 export function useTestCache() {
-  expectType<Cache>(useSWRConfig().cache)
+  expectType<Map<string, CacheValue>>(useSWRConfig().cache)
+  expectType<Map<string, CacheValue>>(
+    useSWRConfig<Map<string, CacheValue>>().cache
+  )
```
chibicode, line 12: I believe we no longer want to use Map: https://github.com/vercel/swr/pull/1961/files#r872298911

chibicode (commented):
There might be some type regression from https://github.com/vercel/swr/pull/1961/

_internal/utils/use-swr-config.ts
promer94: Yes, since the default cache is actually a `Map`.   
https://github.com/vercel/swr/blob/9a0328fffdeb8ca0eaf42bc88f318f7e48a3b245/_internal/utils/config.ts#L44

_internal/types.ts
promer94: These two types should be merged

PR #1963, opened 2022-05-13
build: Mangle internal export names

shuding:
In #1962 we added a shared internal module for all subpath imports to use. However the named exports can't be minified (by many bundlers) because they're across modules. Here we are using TypeScript enums to ensure that 1) these names can be minified, and 2) we can still import the correct utils with typings easily.

This decreases the bundle size by 800+ bytes (minified) for just `'swr'`, and will probably save more if other subpaths are used such as `'swr/infinite'`.


core/index.ts
```diff
@@ -1,11 +1,12 @@
+// @ts-nocheck
```
huozhi, line 1: why do we need this?

shuding: Good catch – it was for testing, removed!

_internal/enums.ts
```diff
@@ -0,0 +1,31 @@
+export const enum InternalUtils {
```
huozhi, line 1: maybe explain why we need to have this as comment here, to help contributors understand better this is for size compression purpose

shuding:
Closing for now as most of bundlers are already handling this today. For the other cases that modules can’t be bundled together (bundleless solutions, or ending up in different chunks somehow), we will wait and see if it has an impact or not.

PR #1967, opened 2022-05-14
build: import with-selector with extension from useESE

huozhi:
useESE 1.1.0 is missing `exports` field which will break esm module resolution

* add extension
* `./shim/with-selector` and `./with-selector` references the same file


shuding:
@huozhi swr/infinite uses the shim as well, could you also update that? Thanks!

huozhi:
@shuding the `use-sync-external-store/shim/with-selector` is ambiguous for esm loader but `use-sync-external-store/shim` import works well since it's a folder with `index.js` inside. So I only changed with-selector

shuding:
@huozhi I am running into a problem with that:

CleanShot 2022-05-14 at 13 21 16@2x

huozhi:
oh I see I changed the exports field in node_modules....

PR #1968, opened 2022-05-14
Bug fixes

shuding:
This PR gets rid of the `as SWRInfiniteResponse<Data, Error>` casting from `swr/infinite`. Also, it ensures that `setSize` and `mutate` returned from `useSWRInfinite` are strictly promises as the typings required (a bug), and fixes `MutatorOptions<Data>` which should be `MutatorOptions<Data[]>` instead.

### DIFF
--- a/infinite/index.ts
+++ b/infinite/index.ts
@@ -30,7 +30,7 @@ import type {
 import { useSyncExternalStore } from 'use-sync-external-store/shim/index.js'
 
 const INFINITE_PREFIX = '$inf$'
-const EMPTY_PROMISE = Promise.resolve(UNDEFINED)
+const EMPTY_PROMISE = Promise.resolve() as Promise<undefined>
 
 const getFirstPageKey = (getKey: SWRInfiniteKeyLoader) => {
   return serialize(getKey ? getKey(0, null) : null)[0]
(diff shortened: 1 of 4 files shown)


infinite/index.ts
```diff
@@ -30,6 +30,7 @@ import type {
 import { useSyncExternalStore } from 'use-sync-external-store/shim/index.js'
 
 const INFINITE_PREFIX = '$inf$'
+const EMPTY_PROMISE = Promise.resolve(UNDEFINED)
```
huozhi, line 33: can save 1 byte without passing UNDEFINED?

infinite/index.ts
```diff
@@ -225,7 +220,13 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
           }
         }
 
-        return args.length ? swr.mutate(data, shouldRevalidate) : swr.mutate()
+        // We are using `arguments` here because there's no other way around to
+        // tell if an argument is missing or `undefined`. It will be compiled to
+        // `arguments` anyways.
+        // @ts-ignore
+        return arguments.length
```
huozhi, line 227: Can we use casting for arguments instead of ts-ignore comment?

huozhi (approved):
non blocking comments, LGTM

infinite/index.ts
shuding: Right, although we need to `as Promise<undefined>` because it will be `Promise<void>` if missing the value.

infinite/index.ts
shuding: > casting for arguments

Can you please explain it?

PR #1970, opened 2022-05-15
Fix race conditions related to optimistic UI

### LINKED ISSUE #1965: Race condition with optimisticData and rollbackOnError
# Bug report

Hey there! First of all thank you for your great job! SWR looks very powerful to use.
However there's some weird behaviour I noticed during learning documentation and examples.
I used original example of optimistic updates and faced with unclear behaviour.
If 2 mutates are called in a short period of time, and both are failed, the "previous" optimisticData would be written into the cache.
Better to show in short [video](https://drive.google.com/file/d/1-W4hKG_xYGboqmZTbH4nOJBYcg-vOOfX/view).

## Expected Behavior

In this particular case I expect the cache shouldn't be changed. However in general it looks like there's some option is needed that allows to control the cache value in terms of the fetch is failed.

## Repro Steps / Code Example

You also able to reproduce this case in my [sandbox](https://codesandbox.io/s/swr-basic-forked-fcss5o?file=/src/App.js). Just click add button twice and you'll see there's both mutations failed, while one todo item is left in the list.

## Additional Context

SWR version: 1.3.0

shuding:
Fixes #1965.

### DIFF
--- a/_internal/utils/mutate.ts
+++ b/_internal/utils/mutate.ts
@@ -86,7 +86,7 @@ export const internalMutate = async <Data>(
   // Do optimistic data update.
   if (hasOptimisticData) {
     optimisticData = isFunction(optimisticData)
-      ? optimisticData(currentData)
+      ? optimisticData(originalData)
       : optimisticData
     set({ data: optimisticData, _o: originalData })
   }
@@ -94,7 +94,7 @@ export const internalMutate = async <Data>(
   if (isFunction(data)) {
     // `data` is a function, call it passing current cache value.
     try {
-      data = (data as MutatorCallback<Data>)(currentData)
+      data = (data as MutatorCallback<Data>)(originalData)
     } catch (err) {
       // If it throws an error synchronously, we shouldn't update the cache.
       error = err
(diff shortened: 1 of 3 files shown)


_internal/utils/mutate.ts
```diff
@@ -70,20 +79,22 @@ export const internalMutate = async <Data>(
...
-  const originalData = get().data
+  const state = get()
+  const currentData = state.data
+  const originalData = isUndefined(state._o) ? currentData : state._o
 
   // Do optimistic data update.
   if (hasOptimisticData) {
     optimisticData = isFunction(optimisticData)
-      ? optimisticData(originalData)
+      ? optimisticData(currentData)
```
shuding, line 89: Still need to discuss that should we pass the original or current data here.

_internal/utils/mutate.ts
```diff
@@ -70,20 +79,22 @@ export const internalMutate = async <Data>(
...
       : optimisticData
-    set({ data: optimisticData })
+    set({ data: optimisticData, _o: originalData })
   }
 
   if (isFunction(data)) {
     // `data` is a function, call it passing current cache value.
     try {
-      data = (data as MutatorCallback<Data>)(originalData)
+      data = (data as MutatorCallback<Data>)(currentData)
```
shuding, line 97: Same here.

_internal/utils/mutate.ts
```diff
@@ -118,15 +129,15 @@ export const internalMutate = async <Data>(
     if (!error) {
       // Transform the result into data.
       if (isFunction(populateCache)) {
-        data = populateCache(data, originalData)
+        data = populateCache(data, currentData)
```
shuding, line 132: Same here.

PR #1986, opened 2022-05-20
fix: should only update cache when actually start new request

### LINKED ISSUE #1974: Beta feedback issue: isLoading and isValidating are changing indefinitely like in a loop in 2.0.0-beta.3
# DISCLAIMER
I am using beta version, and I want to give an early feedback. Probably you will encounter the issue but I wanted to help a bit. Sorry but I do not have a time atm to dig more. Ignore this issue if you do not take bug reports for beta versions.
And thanks for SWR 2 :)

# Bug report

isLoading and isValidating is updating all the time from true to false and again in a loop even if there is no revalidation or mutation.

## Expected Behavior

If there is no mutation or revalidation, `isLoading` and `isValidating` props should NOT transition from true to false and again in a loop.

## Additional Context

It is working in SWR version: 2.0.0-beta.1
It is **NOT** working in SWR version: 2.0.0-beta.3

promer94:
close #1974

https://codesandbox.io/s/loading-nh4jtc

### DIFF
--- a/core/use-swr.ts
+++ b/core/use-swr.ts
@@ -276,10 +276,10 @@ export const useSWRHandler = <Data = any, Error = any>(
       if (isUndefined(getCache().data)) {
         initialState.isLoading = true
       }
-      setCache(initialState)
-
       try {
         if (shouldStartNewRequest) {
+          // TODO: should add a test case for this https://github.com/vercel/swr/issues/1974
+          setCache(initialState)
           // If no cache being rendered currently (it shows a blank page),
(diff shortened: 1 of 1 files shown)


shuding (commented):
Good catch! We will need a test for this

huozhi (commented):
Can we separate the bug fix with e2e testing setup in 2 PRs? feel like they're not strongly bound together.

We also need to include need to include the `test/e2e` folder to prettier included files

shuding (approved):

PR #1989, opened 2022-05-24
feat: mutate filter

### LINKED ISSUE #1946: RFC: Mutate multiple items
Currently, the global `mutate` API (imported directly from this package, or returned by `useSWRConfig()`) allows you to mutate a specific key in the SWR cache:

```js
import { mutate } from 'swr'
// or:
// { mutate } = useSWRConfig()

mutate(key, data)
```

Where `key` can be a valid SWR key argument (`string | null | undefined | false | any[] | Record<any, any>`), or a function that returns a key.

## Proposal

In this proposal, we will make it also accept a filter function, and **deprecate the case of function that returns a key**. This extends the `mutate` API to filter through all the existing keys and mutate multiple of them in one call.

```js
mutate(key => key.startsWith('/api/'), ...)
```

This doesn't change the hook-returned `mutate` API (`const { mutate } = useSWR(...)`).

### Clear All

One typical use case is to clear or revalidate all keys, this can be done by passing a `() => true` function:

```js
function onLogout() {
  // Invalidate session
  await logout()

  // Match all keys, set them to undefined, and skip revalidation.
  mutate(() => true, undefined, false)
}
```

### Revalidate All

Some dashboards have the feature of switching the scope. Usually the API URL is still the same after switching the scope, but the authentication token might change and affect the final resource.

We used to handle this kind of things with token as the second argument in keys:

```js
useSWR(['/api/user', token], fetcher)
```

If we don't include `token` in the key, when switching the scope the resource won't refresh automatically and might cause the UI being outdated (still showing the previous scope).

But then the pain comes when you want to mutate that resource: `mutate(['/api/user', ???])`. You'll have to know the token which identifies the resource. It somehow makes sense but in most cases the API is still the unique identifier of the resource here, as we can't have two tokens at the same time in one app instance.

With the extended `mutate` API, we can solve it this way:

```js
// Use the resource
useSWR('/api/user', url => fetch(url, { headers: {
  // The token can be provided by the current scope or even global,
  // not necessarily bound to the resource.
  Authentication: `Bearer ${token}`
}})...)

// After switching the scope
function changeUser() {
  refreshToken()

  // Invalidate all the resources and revalidate these are currently rendered on the screen.
  mutate(() => true, undefined)
}
```

We omit the `false` argument from `mutate`, which defaults to `true`, meaning we want to refetch these resources again from the server. And the end user will see "old user → skeleton → new user".

### Mutate Only a Subset

Another common case is to only mutate/clear authentication-required resources, but keep the public ones. This can be done by using a filter function:

```js
// Logged out, but users still can see public data.
mutate(key => !key.startsWith('/api/public'), undefined, false)
```

Also, it can be helpful to mutate resources that belong to the same category, such as a list of pages:

```js
// Uppercase all item names in `/api/items?page=<page>`
mutate(key => /\/api\/items\?page=\d+/.test(key), items =>
  items.map(item => ({ ...item, name: item.name.toUpperCase() }))
)
```

### Simplicity

The most popular use case as far as we know is to clear all the cache, which doesn't look that easy and intuitive (`mutate(() => true, undefined, false)`). One way to simplify that is to add a shorthand API specifically for that case.

### useSWRInfinite

Apparently we need to skip special keys (where we keep the pages array, size and extra context) when doing a match mutate. But how do we tell the "pages state" to revalidate? Like updating the number of pages when you mutate a page to be `undefined`.

### Batching

Consider this common dependent request scenario:

```js
const foo = useSWR('foo').data
const bar = useSWR(foo ? 'bar?foo=' + foo.id : null).data
```

If we do `mutate(() => true, undefined, true)`. And if the "mutate and revalidate" actions are batched (no render in between), then this will happen:

```
hook 1: foo → [undefined → revalidation] → foo
hook 2: bar → [undefined → revalidation*] → key changed to `null` → key changed to `bar?foo=foo.id` → revalidation
```

Which causes an extra (*) revalidation that is unnecessary.

However, if we do the mutation and trigger the revalidation in another tick, we can avoid this problem but there might be other issues yet to explore.

### Atomic

If we mutate multiple resources, do we think of them as multiple actions that are unrelated to each other, or one **atomic** action that affects multiple resources? For the latter we need to combine all revalidations and handle cache populations and error rollbacks after all of them, to keep data consistency.

### Key Serialization

SWR accepts objects as the key. For example:

```js
useSWR([ '/api/user', 'uid' ], ...)
useSWR([ getUser, 'uid' ], ...)
useSWR({ api: '/api/user', uid: 'uid' }, ...)
```

But then internally SWR serializes the object into a string. This makes it impossible to filter by key. One solution is to keep the original key object in cache with a special property name `_key`, and also add `toJSON` to keep the cache compatible with things like `localStorage`. And then, the filter function will accept the original key value:

```js
mutate(key => Array.isArray(key) && key[0] === '/api/user', ...)
```

However, it forces you to always check the type of the key first when doing a mutate action.

huozhi:
Closes #1946.

## Description

The global `mutate` API (imported directly from SWR or returned by `useSWRConfig`) can now accept a filter function, to match multiple keys for mutating:

```js
await mutate(
  key => typeof key === 'string' && key.startsWith('/api/'),
  undefined, // newData | (data => newData)
  false,     // shouldRevalidate = true
)
```

## Breaking Changes

- This deprecates the use case of accepting a function that returns a key.
- The `Cache` interface now requires a `keys()` method that returns all keys in the cache object, similar to the JavaScript `Map` instances.

## Usage

(See #1946)

### DIFF
--- a/_internal/utils/mutate.ts
+++ b/_internal/utils/mutate.ts
@@ -49,8 +49,6 @@ export async function internalMutate<Data>(
   const revalidate = options.revalidate !== false
   const rollbackOnError = options.rollbackOnError !== false
 
-  if (!_key) return
-
   const KEYS = (SWRGlobalState.get(cache) as GlobalState)[6]
 
   // If 2nd arg is key filter, return the mutation results of filtered keys
@@ -63,6 +61,7 @@ export async function internalMutate<Data>(
     return await Promise.all(matchedKeys.map(mutateByKey))
   } else {
     const [serializedKey] = serialize(_key)
+    if (!serializedKey) return
     return await mutateByKey(serializedKey)
   }
 
--- a/_internal/utils/mutate.ts
+++ b/_internal/utils/mutate.ts
@@ -12,7 +12,8 @@ import {
   Arguments
 } from '../types'
 
-type KeyFilter = (key?: string) => boolean
+type ValidKey = Omit<Arguments, 'null' | 'undefined'>
+type KeyFilter = (key?: ValidKey) => boolean
 
 export async function internalMutate<Data>(
   cache: Cache,
@@ -50,11 +51,13 @@ export async function internalMutate<Data>(
 
   if (!_key) return
 
+  const KEYS = (SWRGlobalState.get(cache) as GlobalState)[6]
+
   // If 2nd arg is key filter, return the mutation results of filtered keys
   if (isFunction(_key)) {
     const keyFilter = _key
     const matchedKeys = []
-    for (const k of cache.keys()) {
+    for (const k of KEYS.values()) {
       if (keyFilter(k)) matchedKeys.push(k)
     }
     return await Promise.all(matchedKeys.map(mutateByKey))
@@ -63,7 +66,8 @@ export async function internalMutate<Data>(
     return await mutateByKey(serializedKey)
   }
 
-  async function mutateByKey(key: string): Promise<Data | undefined> {
+  async function mutateByKey(_k: ValidKey): Promise<Data | undefined> {
+    const [key] = serialize(_k)
     const [get, set] = createCacheHelper<
       Data,
       State<Data, any> & {
--- a/_internal/utils/mutate.ts
+++ b/_internal/utils/mutate.ts
@@ -9,7 +9,6 @@ import {
   MutatorOptions,
   GlobalState,
   State,
-  Falsy,
   Arguments
 } from '../types'
 
@@ -18,26 +17,12 @@ type KeyFilter = (key?: string) => boolean
 export const internalMutate = async <Data>(
   ...args: [
     Cache,
-    KeyFilter | Arguments | Falsy,
+    KeyFilter | Arguments,
     undefined | Data | Promise<Data | undefined> | MutatorCallback<Data>,
     undefined | boolean | MutatorOptions<Data>
   ]
-): Promise<(Data | undefined)[]> => {
-  const [cache, _keyFilter, _data, _opts] = args
-  if (!_keyFilter) return Promise.resolve([])
-  let matchedKeys: string[]
-
-  let keyFilter: KeyFilter
-  if (typeof _keyFilter !== 'function') {
-    const [serializedKey] = serialize(_keyFilter)
-    matchedKeys = [serializedKey]
-  } else {
-    keyFilter = _keyFilter as KeyFilter
-    matchedKeys = []
-    for (const _key of cache.keys()) {
-      if (keyFilter(_key)) matchedKeys.push(_key)
-    }
-  }
+): Promise<(Data | undefined)[] | Data | undefined> => {
+  const [cache, _key, _data, _opts] = args
 
   // When passing as a boolean, it's explicitly used to disable/enable
   // revalidation.
@@ -51,6 +36,21 @@ export const internalMutate = async <Data>(
   let optimisticData = options.optimisticData
   const revalidate = options.revalidate !== false
   const rollbackOnError = options.rollbackOnError !== false
+
+  if (!_key) return
+
+  // If 2nd arg is key filter, return the mutation results of filtered keys
+  if (isFunction(_key)) {
+    const keyFilter: KeyFilter = _key
+    const matchedKeys = []
+    for (const k of cache.keys()) {
+      if (keyFilter(k)) matchedKeys.push(k)
+    }
+    return await Promise.all(matchedKeys.map(mutateByKey))
+  }
+
+  const [serializedKey] = serialize(_key)
+  return await mutateByKey(serializedKey)
 
   async function mutateByKey(key: string): Promise<Data | undefined> {
     const [get, set] = createCacheHelper<
@@ -175,5 +175,4 @@ export const internalMutate = async <Data>(
     if (error) throw error
     return populateCache ? res : data
   }
-  return await Promise.all(matchedKeys.map(_key => mutateByKey(_key)))
 }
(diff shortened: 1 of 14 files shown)


_internal/utils/mutate.ts
```diff
@@ -4,23 +4,40 @@ import { SWRGlobalState } from './global-state'
...
   ...args: [
     Cache,
-    Key,
+    KeyFilter | Arguments | Falsy,
     undefined | Data | Promise<Data | undefined> | MutatorCallback<Data>,
     undefined | boolean | MutatorOptions<Data>
   ]
-): Promise<Data | undefined> => {
-  const [cache, _key, _data, _opts] = args
+): Promise<(Data | undefined)[]> => {
```
shuding, line 25: We shouldn't break the existing use case where `mutate` returns just the bare `Data`. We should overload the function type to only return `Data[]` if a filter function is passed in.

_internal/utils/mutate.ts
```diff
@@ -4,23 +4,40 @@ import { SWRGlobalState } from './global-state'
...
+
+  let keyFilter: KeyFilter
+  if (typeof _keyFilter !== 'function') {
+    const [serializedKey] = serialize(_keyFilter)
+    matchedKeys = [serializedKey]
+  } else {
+    keyFilter = _keyFilter as KeyFilter
+    matchedKeys = []
+    for (const _key of cache.keys()) {
+      if (keyFilter(_key)) matchedKeys.push(_key)
```
shuding, line 38: It should pass the original key object here, but that can be a separate PR to make it easier

huozhi: I feel we need to store the original key in the cache...

_internal/utils/mutate.ts
```diff
@@ -35,130 +49,145 @@ export const internalMutate = async <Data>(
...
-        )
-      }
+  const KEYS = (SWRGlobalState.get(cache) as GlobalState)[6]
+
+  // If 2nd arg is key filter, return the mutation results of filtered keys
+  if (isFunction(_key)) {
+    const keyFilter = _key
+    const matchedKeys = []
+    for (const k of KEYS.values()) {
+      if (keyFilter(k)) matchedKeys.push(k)
```
shuding, line 61: One nice thing is to catch potential mistakes that people forget to update their code for this breaking change:

```suggestion
      const shouldMutateKey = keyFilter(k)
      if (typeof shouldMutateKey == 'boolean') {
        if (shouldMutateKey) {
          matchedKeys.push(k)
        }
      } else {
        throw new Error('It\'s not allowed to use function to return a key in `mutate` since SWR 2.0.')
      }
```

huozhi: We also have falsy value support for the swr keys, and this feel like we're doing a part of type validation with js. And the return type will become uncertain, if there's no error, it will resolve an array of values or errors, in this case will only bail with 1 error.

_internal/utils/mutate.ts
```diff
@@ -35,130 +54,136 @@ export const internalMutate = async <Data>(
...
-  const [get, set] = createCacheHelper<
-    Data,
-    State<Data, any> & {
-      // The previously committed data.
-      _c?: Data
+  // If 2nd arg is key filter, return the mutation results of filtered keys
+  if (isFunction(_key)) {
+    const keyFilter = _key
+    const matchedKeys: Key[] = []
+    for (const state of cache.values()) {
```
huozhi, line 61: Since we store the `key` in cache state, now we need to use `values()` or `entries()` method to get the result. Not sure if it's a good design for `Cache` typing.

Or use `keys`, but it requires to call `cache.get(key)` here, maybe it's a heavy call.

_internal/utils/mutate.ts
```diff
@@ -54,16 +54,20 @@ export async function internalMutate<Data>(
...
-  // If 2nd arg is key filter, return the mutation results of filtered keys
+  // If the second argument is a key filter, return the mutation results for all
+  // filtered keys.
   if (isFunction(_key)) {
     const keyFilter = _key
     const matchedKeys: Key[] = []
-    for (const state of cache.values()) {
-      if (keyFilter(state.key)) matchedKeys.push(state.key)
+    for (const key of cache.keys()) {
+      if (keyFilter((cache.get(key) as { _k: any })._k)) {
```
huozhi, line 63:
```js
function bench() {
  const map = new Map()
  for (let i = 0; i < 10000000; i++) {
    map.set(i, i)
  }

  const label1 = 'keys + get'
  console.time(label1)
  for (const k of map.keys()) {
    map.get(k)
  }
  console.timeEnd(label1)

  const label2 = 'values'
  console.time(label2)
  for (const v of map.values()) {
    v
  }
  console.timeEnd(label2)
}
```

`keys()` + `get` will be slower than values 🤔 
```
keys + get: 216.172ms
values: 79.717ms
```

shuding: It's too unrealistic to have 10 million keys in the cache though. Your app will be out of memory before that happens 😄

huozhi: Yeah right, let's keep `keys` + `get()` for orthogonality

_internal/utils/mutate.ts
huozhi: let's improve the error/warning later as enhancement

huozhi:
Thank you shu for helping fixing the issues and those changes look good

image

shuding:
(I think there’s still one last blocker which is to make useSWRInfinite work properly)

shuding:
OK this is more like a general problem today, need to think deeper about it. Meanwhile this PR can be shipped!

PR #1990, opened 2022-05-24
chore: add engine field to limit pnpm version

huozhi:
We're using pnpm 7, and pnpm v6 will generate different lock files. Add engine field to package.json for pnpm version checking

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,6 +142,10 @@
     "arrowParens": "avoid",
     "trailingComma": "none"
   },
+  "engines": {
+    "node": ">= 14",
+    "pnpm": ">= 7"
+  },
   "dependencies": {
     "use-sync-external-store": "^1.1.0"
   }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -142,6 +144,10 @@
     "arrowParens": "avoid",
     "trailingComma": "none"
   },
+  "engines": {
+    "node": ">= 14",
```
shuding, line 148: Will this make npm yell when people are installing SWR to their projects, when they have node < 14?

package.json
```diff
@@ -142,6 +144,10 @@
...
+    "pnpm": "7"
```
shuding, line 149:
```suggestion
    "pnpm": ">= 7"
```

package.json
huozhi: Yes it will. Do you have any concern about this?

huozhi: Emmm yes, probably we don't need node sincei it's not using any specific node api

PR #1991, opened 2022-05-25
chore: save some bytes

promer94:


shuding (commented):
You can use the mergeObject util inside _internal for this

huozhi (approved):

PR #1995, opened 2022-05-25
fix: useSWRMutation - clear error after successful trigger (#1973)

### LINKED ISSUE #1973: Misbehavior of the error state from useSWRMutation
I am having a little bit of trouble understanding how the error handling works with `useSWRMutation`. I copied the example from https://swr.vercel.app/examples/error-handling but updated it to use `useSWRMutation` and `trigger` instead of `useSWR` and `mutate`. Here is a code sandbox of that: https://codesandbox.io/s/exciting-ives-rvxuh2?file=/src/App.tsx

A couple things I noticed: 

* when the fetcher throws an error, it doesn't seem to be caught completely
* when there is an error followed by a successful call, the error does not clear

These two behaviors are different from `useSWR` with `mutate`. Am I thinking about `useSWRMutation` and `trigger` incorrectly?

_Originally posted by @ericponto in https://github.com/vercel/swr/discussions/1919#discussioncomment-2688176_

anirudh1713:
closes #1973

### DIFF
--- a/mutation/index.ts
+++ b/mutation/index.ts
@@ -68,7 +68,7 @@ const mutation = (<Data, Error>() =>
 
           // If it's reset after the mutation, we don't broadcast any state change.
           if (ditchMutationsUntilRef.current <= mutationStartedAt) {
-            setState({ data, isMutating: false })
+            setState({ data, isMutating: false, error: undefined })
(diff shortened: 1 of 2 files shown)


shuding:
Thanks!

Issue #1041, opened 2021-03-16
Add an equivalent of React Query's useQueries hook


gfox1984:
There does not seem to be an equivalent of the React Query's [useQueries hook](https://react-query.tanstack.com/guides/parallel-queries#dynamic-parallel-queries-with-usequeries) in SWR.

Being able to run a variable number of queries in parallel using SWR and yet benefit from caching per individual items would be a big enabler. I know that I can use a custom fetcher with SWR which returns the Promise.all of multiple requests, but that doesn't cache results individually.

A good use case is given on the link above. Imagine having to fetch multiple users at once, and then in other parts of the code, only request for one of these users. You would definitely like to avoid making a new request for the same user requested earlier.

Note: I [tried to ask the community](https://github.com/vercel/swr/discussions/1038) for a way to do that with SWR, without success.

shuding:
Thanks for the detailed description! So the solution would be something similar to the impl. of useSWRInfinite, but requests are fired together instead of paginated. 

Happy to add it if someone comes up with a PR, I think `useSWRList` would be a nice name. 👍

promer94:
I think we could use the global mutate function to cache the each item based on their key. 

```js
import { mutate } from 'swr'
function useSWRList(queryList) {
  const mutations = queryList.map(({ queryKey, queryFn }) => {
    return mutate(queryKey, () => queryFn(...queryKey))
      .then((v) => ({ data: v, error: null }))
      .catch((e) => ({ error: e, data: null }))
  })

  return useSWR(
    queryList.map((v) => v.queryKey),
    () => Promise.all(mutations)
  )
}
```


FYI: https://codesandbox.io/s/swr-list-xw58t

nandorojo:
We'd also want to catch errors for each key individually, so that some calls can fire even if others aren't ready. If the first request's key throws an error, the others should still fire.

promer94:
@nandorojo Hi, would you mind check the codesandbox again. I have updated the example.   
```js
const mutations = queryList.map(({ queryKey, queryFn }) => () =>
    mutate(queryKey, () => queryFn(...queryKey))
      // The data has been saved in cache based on key
      .then((v) => v)
      // The error has been save in cache base on key
      // so the error was handled individually
      .catch((e) => ({
        error: e,
        key: queryKey
      }))
  )
```

I feel like this pattern might be batter then  useQueries hook. It only rerenders onces when all data resolves. 
[...]

PR #2006, opened 2022-06-01
useSWRList

promer94:
RFC: #1988
Issue: #1041

Demo: https://codesandbox.io/s/useswrlist-8lrp6k

### DIFF
--- a/examples/queries/README.md
+++ b/examples/queries/README.md
@@ -1,0 +1,30 @@
+# Basic
+
+## One-Click Deploy
+
+Deploy your own SWR project with Vercel.
+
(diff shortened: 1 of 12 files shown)


promer94:
This pr intends to support both proposals metioned in RFC. 

## Proposal A: "Array of SWRs"
### API
```
const { result } = useSWRList(keys, fetcher, config)
```
```typescript
interface Result {
    data: any;
    error: any;
    mutate: KeyedMutator<any>;
    isValidating: boolean;
    isLoading: boolean;
    key: string;
    originKey: Arguments;
}
```

## Proposal B: Aggregated Return
### API
```
const { data, isLoading, isValidating, mutate } = useSWRList(keys, fetcher, config)
```
```typescript
type Data = {
    data?: any;
    error?: any;
    key: string;
    originKey: Arguments;
}[]
```

PR #2018, opened 2022-06-09
WIP: useSWRList

promer94:


0xbe1:
any plan on crossing the finish line? look forward to using this feature

Conmiro:
Would also like to utilize this! We have a workaround that involves looping thru the list of service calls, creating a component for each which uses the useSwr() hook, then utilizes a callback/reducer to send the data back to the parent which aggregates everything.

Our particular use case is we have data that's loading in different components throughout the page, but then also have a component that needs to know the status of which panels have had their data loaded already.

PR #2024, opened 2022-06-11
feat: support functional swr context config

huozhi:
### Background

`SWRConfig` now merges the config passed in `value` property with SWR config from parent context, but this bring a problem that users have to be aware of how many layers of context are existing and there's no way to ignore the config from parent context

### Feature

* Support `(parentConfig? Config) => Config` overloading for `SWRConfig` value property

Then you can pick any properties from parent config and decide how to combine with the current SWR config, or you can just ignore any config property from parent context. e.g. you have a `useSWR` hook inside very deep level of the react tree, but you want it to keep the default config, you can just ignore all configs by ignoring all the parent configs.

#### Example

```js
<SWRConfig value={config1}>
  <Page>
    <SWRConfig value={() => ({})}>
      <Main />
    </SWRConfig>
</SWRConfig>
```


shuding (approved):
Looks great, thanks

PR #2026, opened 2022-06-12
feat: swr/preload

koba04:
This is an experimental implementation to preload fetcher functions. The following is an example.

```tsx
import useSWR, { prefetch } from 'swr/preload'

// before rendering
prefetch('/api/user', fetcher);

const Page = () => {
  const { data } = useSWR('/api/user', fetcher);
  return <div>...</div>
}
```

The preload function is necessary to avoid waterfall problems with the suspense mode.
Currently, this is implemented as middleware, but the API might be changed before merging.
The limitation of the current implementation is that we have to use `useSWR` from `swr/preload`.

The following has the waterfall problem.

```tsx
import useSWR, { prefetch } from 'swr/preload'

const Page = () => {
  // this suspends the first rendering
  const { data: user } = useSWR('/api/user', fetcher, { suspense: true });
  // this start fetching only after `/api/user` has been fetched
  const { data: movies } = useSWR('/api/movies', fetcher, { suspense: true });
  return <div>...</div>
}
```

To avoid that, we should preload fetcher functions. `prefetch` is the function for it.

```tsx
import useSWR, { prefetch } from 'swr/preload'

// before rendering
prefetch('/api/user', fetcher);
prefetch('/api/movies', fetcher);

const Page = () => {
  // These fetcher functions have been already fetched by calling `prefetch`.
  const { data: user } = useSWR('/api/user', fetcher, { suspense: true });
  const { data: movies } = useSWR('/api/movies', fetcher, { suspense: true });
  return <div>...</div>
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,12 @@
       "require": "./immutable/dist/index.js",
       "types": "./immutable/dist/immutable/index.d.ts"
     },
+    "./preload": {
+      "import": "./preload/dist/index.mjs",
+      "module": "./preload/dist/index.esm.js",
+      "require": "./preload/dist/index.js",
+      "types": "./preload/dist/preload/index.d.ts"
+    },
     "./mutation": {
       "import": "./mutation/dist/index.mjs",
       "module": "./mutation/dist/index.esm.js",
@@ -51,11 +57,13 @@
     "core/dist/**",
     "infinite/dist/**",
     "immutable/dist/**",
+    "preload/dist/**",
     "mutation/dist/**",
     "_internal/dist/**",
     "core/package.json",
     "infinite/package.json",
     "immutable/package.json",
+    "preload/package.json",
     "mutation/package.json",
     "_internal/package.json",
     "package.json"
@@ -73,11 +81,13 @@
     "watch:core": "turbo run watch --filter=swr-core",
     "watch:infinite": "turbo run watch --filter=swr-inifinite",
     "watch:immutable": "turbo run watch --filter=swr-immutable",
+    "watch:preload": "turbo run watch --filter=swr-preload",
     "watch:mutation": "turbo run watch --filter=swr-mutation",
     "watch:internal": "turbo run watch --filter=swr-internal",
     "build:core": "turbo run build --filter=swr-core",
     "build:infinite": "turbo run build --filter=swr-infinite",
     "build:immutable": "turbo run build --filter=swr-immutable",
+    "build:preload": "turbo run build --filter=swr-preload",
     "build:mutation": "turbo run build --filter=swr-mutation",
     "build:internal": "turbo run build --filter=swr-internal",
     "prepublishOnly": "pnpm clean && pnpm build",
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -4,3 +4,4 @@ packages:
   - 'immutable'
   - 'infinite'
   - 'mutation'
+  - 'preload'
--- a/preload/index.ts
+++ b/preload/index.ts
@@ -1,0 +1,31 @@
+import useSWR, { Middleware, unstable_serialize, Key, BareFetcher } from 'swr'
+import { withMiddleware } from 'swr/_internal'
+
+const preloadMap = new Map<string, Promise<any>>()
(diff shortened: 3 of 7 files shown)


koba04:
@shuding Thank you! That makes sense, I'll try it 👍

koba04:
I've renamed the API from `prefetch` to `preload`.

_internal/utils/preload.ts
```diff
@@ -0,0 +1,32 @@
...
+  const keyString = serialize(key)[0]
+  preloadMap.set(keyString, promise)
+  return promise
+}
+
+export const middleware: Middleware = useSWRNext => (key, fetcher_, config) => {
+  const fetcher =
+    fetcher_ == null
+      ? null
+      : // fetcher might be a sync function, so this should not be an async function
```
koba04, line 17: If this is an async function, https://github.com/vercel/swr/blob/9263ca79f0747f9132fdcd0cf5eb0e47f7ee08a8/test/use-swr-error.test.tsx#L433 is failed.

shuding: Makes sense!

koba04:
I've added a test for resetting an error when preloading has been failed.

_internal/utils/preload.ts
```diff
@@ -0,0 +1,30 @@
+import { Middleware, Key, BareFetcher } from '../types'
+import { serialize } from './serialize'
+
+const REQUEST = new Map<string, Promise<any>>()
```
huozhi, line 4: should this be bound with cache in global state?

_internal/utils/resolve-args.ts
```diff
@@ -18,10 +19,9 @@ export const withArgs = <SWRType>(hook: any) => {
     // Apply middleware
     let next = hook
     const { use } = config
-    if (use) {
-      for (let i = use.length; i--; ) {
-        next = use[i](next)
-      }
+    const middlewares = (use || []).concat(BUILT_IN_MIDDLEWARES)
```
huozhi, line 22: middleware is uncountable noun, hence we don't add `s` for it

_internal/utils/preload.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const preload = (key_: Key, fetcher: BareFetcher<any>) => {
```
huozhi, line 6: `preload` could take generic types and passdown to Fetcher?

_internal/utils/preload.ts
```diff
@@ -0,0 +1,30 @@
...
+  const req = fetcher(key_)
+  const key = serialize(key_)[0]
+  REQUEST.set(key, req)
+  return req
+}
+
+export const middleware: Middleware =
+  useSWRNext => (key_, fetcher_, config) => {
+    const fetcher =
+      fetcher_ == null
```
huozhi, line 16: support more falsy fetcher values like `false`/`undefined`/etc?

koba04: @huozhi Thank you! 
Because of the following line, all falsy fetcher values would be `undefined`. But I've changed this to ignore all falsy values.
https://github.com/vercel/swr/blob/9263ca79f0747f9132fdcd0cf5eb0e47f7ee08a8/_internal/utils/resolve-args.ts#L27

_internal/utils/preload.ts
koba04: @huozhi Thank you!　That makes sense to me.
But does the preload function have to accept `cache` as the argument to use the global state? (It might work if we always use a **global** cache to store the value)

```jsx
// we need the cache argument to get REQUEST from SWRGlobalState
preload(key, fetcher, cache) 
```

koba04: I've used `SWRGlobalState` to store preloading requests.
https://github.com/vercel/swr/pull/2026/commits/85f160cdafebf2e88d69a72ee5b1b4503ce39681

huozhi (approved):
Looks good on my side

shuding:
Looks amazing, thank you!

koba04:
Thank you!

zelal-dev:
Great..!! It looks amazing, best wishes!

PR #2036, opened 2022-06-18
fix: Avoid preloading the resource multiple times

shuding:
In a real-world application, a “preload” action might be initiated from multiple different places and levels (global, inside component, inside hooks, inside event callbacks). Hence deduplication is also needed here.

This improves use cases such as:

```jsx
<button onHover={() => preload('/api/user?id=' + userId, fetcher)}>
```

Where you can hover the button multiple times. Also:

```jsx
function App({ userId }) {
  useEffect(() => {
    preload('/api/user?id=' + userId, fetcher)
  }, [userId])
  ...
}
```

Where there can be multiple `<App/>` elements.


koba04 (approved):
This looks good! Thank you!👍

PR #2038, opened 2022-06-18
Fix ts type generation

### LINKED ISSUE #2037: `SWRConfig`'s `value` type is `any`
# Bug report

While trying out the latest beta I've noticed that `SWRConfig`'s types are broken so I'm getting the `any` type:
image
image

In `node_modules\swr\core\dist\core\use-swr.d.ts`, it looks like `import("_internal/types")` should be `import("swr/_internal")` instead.

## Expected Behavior

No broken types.

## Repro Steps / Code Example

1. Install the latest beta
2. Check `node_modules\swr\core\dist\core\use-swr.d.ts`
3. Notice the broken type imports

## Additional Context

SWR version: 2.0.0-beta.4

- I've also noticed that the following two files are identical, maybe only one is necessary:
  - `node_modules\swr\core\dist\index.esm.js`
  - `node_modules\swr\core\dist\index.mjs`
- `node_modules\swr\core\dist\core\tsconfig.tsbuildinfo` should probably not be published in the npm package

huozhi:
Fixes #2037

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,6 +41,7 @@ jobs:
       - run: pnpm types:check
       - run: pnpm lint
       - run: pnpm build
+      - run: pnpm test-typing
       - run: pnpm test
         env:
           CI: true
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,8 @@ dist
 .eslintcache
 examples/**/yarn.lock
 package-lock.json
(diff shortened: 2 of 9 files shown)


shuding:
Worth adding a test case?

huozhi:
Current test suites are hard to cover the generated types testing. We can do it later

PR #2048, opened 2022-06-24
fix: isLoading and isValidating should always respect cache value

### LINKED ISSUE #2030: isLoading: wrong state
# Bug report

Sometimes, the value of isLoading returned by useSWR is true when there is data and no errors.

## Expected Behavior

The isLoading should be false when there is data returned from useSWR.

## Repro Steps / Code Example

## Additional Context

SWR version: [2.0.0-beta.3](https://github.com/vercel/swr/releases/tag/2.0.0-beta.3).

promer94:
fix #2030
https://codesandbox.io/s/swr-2030-fm0g36

### DIFF
--- a/test/use-swr-loading.test.tsx
+++ b/test/use-swr-loading.test.tsx
@@ -318,4 +318,23 @@ describe('useSWR - loading', () => {
     screen.getByText('loading,validating')
     await screen.findByText('ready,ready')
   })
+  it('isLoading and isValidating should always respect cache value', async () => {
+    const key = createKey()
+    const Page = () => {
+      const { data } = useSWR(key, () => createResponse('result', { delay: 10 }))
+      const { data: response } = useSWR(data, () => createResponse('data', { delay: 10 }))
+      // eslint-disable-next-line react/display-name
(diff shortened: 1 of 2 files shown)


shuding (commented):
Amazing, can you create a test case based on that reproduction?

test/use-swr-loading.test.tsx
```diff
@@ -318,4 +318,23 @@ describe('useSWR - loading', () => {
...
+      const { data } = useSWR(key, () => createResponse('result', { delay: 10 }))
+      const { data: response } = useSWR(data, () => createResponse('data', { delay: 10 }))
+      // eslint-disable-next-line react/display-name
+      const Component = ((_: any) => () => {
+        // eslint-disable-next-line react-hooks/rules-of-hooks
+        const { data: result, isLoading, isValidating } = useSWR(key, () => createResponse('result', { delay: 10 }))
+        return (
+          <div>{`result is ${result ? result : 'null'},${isLoading},${isValidating}`}</div>
+        )
+      })(response)
```
huozhi, line 333: Seems the `response` and the `_` arg are not used?

PR #2065, opened 2022-07-07
Revert swr config generic

huozhi:
### Changes

revert #1938 

After new mutate API is landed, we're able to clear the cache or partially set cache with it, like `mutate((key) => filter(key), data)`. We encourage users to use `mutate` API instead of manipulate cache provider itself directly. 

### Origin Problem
Previously users are struggling with typing with cache provider, and trying to call `cache.clear()`, which might result into inconsistence between cache provider and react state in swr. So we introduce the new mutate API to let users manipulate cache easily. The type generics for `useSWRConfig<CacheType>` is a bit redudant if you really want to forcedly type a returned cache from context. You can just use `as` type guard like below:

```ts
const cache = useSWRConfig().cache as CustomCache
```

By default the cache type is `Cache<any>`

### DIFF
--- a/_internal/types.ts
+++ b/_internal/types.ts
@@ -26,8 +26,8 @@ export type Fetcher<
   : never
 
 // Configuration types that are only used internally, not exposed to the user.
-export interface InternalConfiguration<T extends Cache = Cache> {
-  cache: T
+export interface InternalConfiguration {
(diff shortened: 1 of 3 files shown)


shuding:
Sounds good!

PR #2066, opened 2022-07-08
feat: suspense guard in data and error getters

### LINKED ISSUE #5: Avoiding waterfalls with Suspense

h3rmanj:
Alternative to #168
Fixes #5

This PR isolates the suspense guarding into a function, which is then called in the `data` and `error` getters returned from `useSWR`.

This allows you to write your suspense `useSWR`s like this to avoid waterfalls;

```js
function App () {
  const user = useSWR('/api/user')
  const movies = useSWR('/api/movies')

  return (
    <div>
      Hi {user.data.name}, we have {movies.data.length} movies on the list.
    </div>
  )
}

// OR

function App () {
  const userRequest = useSWR('/api/user')
  const moviesRequest = useSWR('/api/movies')
  const user = userRequest.data
  const movies = moviesRequest.data

  return (
    <div>
      Hi {user.name}, we have {movies.length} movies on the list.
    </div>
  )
}
```

This also allows for dependency fetching

```js
const user = useSWR('/api/user')
const posts = useSWR('/api/posts')
// Accessing .data will throw during key evaluation, letting swr know it's not ready
const movies = useSWR(() => '/api/movies?id=' + user.data.id)
```

And it would also (mostly) be backwards compatible, since most usage uses deconstruction which will automatically call the `data` and/or `error` getters, preserving earlier functionality.

I believe this is a much cleaner interface than the other ones proposed. I've yet to test this completely, so it's mostly a proposal for now, feedback needed 😄

### DIFF
--- a/core/use-swr.ts
+++ b/core/use-swr.ts
@@ -590,33 +590,37 @@ export const useSWRHandler = <Data = any, Error = any>(
   // Display debug info in React DevTools.
   useDebugValue(returnedData)
 
-  // In Suspense mode, we can't return the empty `data` state.
-  // If there is `error`, the `error` needs to be thrown to the error boundary.
-  // If there is no `error`, the `revalidation` promise needs to be thrown to
-  // the suspense boundary.
-  if (suspense && isUndefined(data) && key) {
-    // SWR should throw when trying to use Suspense on the server with React 18,
-    // without providing any initial data. See:
-    // https://github.com/vercel/swr/issues/1832
-    if (!IS_REACT_LEGACY && IS_SERVER) {
-      throw new Error('Fallback data is required when using suspense in SSR.')
-    }
+  const suspenseGuard = () => {
+    // In Suspense mode, we can't return the empty `data` state.
+    // If there is `error`, the `error` needs to be thrown to the error boundary.
+    // If there is no `error`, the `revalidation` promise needs to be thrown to
+    // the suspense boundary.
+    if (suspense && isUndefined(data) && key) {
+      // SWR should throw when trying to use Suspense on the server with React 18,
+      // without providing any initial data. See:
+      // https://github.com/vercel/swr/issues/1832
+      if (!IS_REACT_LEGACY && IS_SERVER) {
+        throw new Error('Fallback data is required when using suspense in SSR.')
+      }
 
-    // Always update fetcher and config refs even with the Suspense mode.
-    fetcherRef.current = fetcher
-    configRef.current = config
-    unmountedRef.current = false
-    throw isUndefined(error) ? revalidate(WITH_DEDUPE) : error
+      // Always update fetcher and config refs even with the Suspense mode.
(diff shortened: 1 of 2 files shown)


huozhi (commented):
I'm not sure if we're going to go with this approach for suspense, since suspense is still a ongoing discussion. And this seems to change the existing behavior of suspense?

nstepien:
So even something like `obj.data === undefined` would trigger suspense?
We'd have to be very careful in cases where we don't want to suspend the page while it's fetching, IMO it seems less ergonomic than `.read()`.

h3rmanj:
Suspense would still need to be explicitly enabled in the options. This PR only changes when it's thrown, if the option is enabled.

nstepien:
Ah I didn't notice that `suspense: true` is still needed (it's missing in the opening comment).

I think I'd still prefer `.read()`:
- the `suspense` setting wouldn't be needed anymore
- `.data` type would still be `T | undefined`
- `.read()` type would always be `T`

In our codebase we wrap swr in custom hooks like this example:
```ts
interface MyData {
  readonly id: number;
}

const key = 'key';

function useData() {
  return useSWR<MyData>(key);
}

function useDataWithSuspense() {
  const { data, ...rest } = useSWR<MyData>(key, { suspense: true });
  return { data: data!, ...rest };
}

function useDataMaybeWithSuspense(suspense = true) {
  return useSWR<MyData>(key, { suspense });
}
```

Because the `suspense` setting doesn't change the return type we always have to add type assertions when using `suspense: true`.
In case when `suspense` is controlled by the component like in `useDataMaybeWithSuspense` we have to be more careful about types.

A solution like `.read()` would be much more ergonomic for us, we'd only need the `useData()` above without having to worry about types.

Just my 2 cents.

h3rmanj:
Yep, I agree having a `read` function returned would be more explicit and probably self-documenting. I wanted to explore this option as the properties returned from useSWR already are getter functions, and would be a minimal change to a much-needed feature.

You could technically do pattern matching with the types
```ts
type SWRSuspenseHook<T> = (suspense: true) => { data: T }
type SWRNormalHook<T> = (suspense?: false) => { data?: T }

type SWRHook<T> = SWRSuspenseHook<T> | SWRNormalHook<T>
```
but good luck doing that with the complex types of the existing SWRHook type 😄

PR #2067, opened 2022-07-10
Make it possible to infer the fetcher's key

Yuddomack:
Hi!

I am using passing objects feature (https://swr.vercel.app/docs/arguments#passing-objects) like this.

image

Unfortunately, This obviously leads to incorrect execution.
But you won't know until execute project.

because this fetcher infers the type as a `bare fetcher`, it shows no error information.

### but, enable key inference with this update,

image

### can be corrected correctly

image

I looked at swr2 (also 1.3), If only a fetcher and a key exist, there is no need to infer that it is a `bare fetcher`.
so I think it would be better to inform the user that this is an incorrect usage.

Please consider this change. 
thank you for read!

### DIFF
--- a/_internal/types.ts
+++ b/_internal/types.ts
@@ -113,10 +113,6 @@ export interface SWRHook {
   <Data = any, Error = any>(key: Key): SWRResponse<Data, Error>
   <Data = any, Error = any>(
     key: Key,
-    fetcher: BareFetcher<Data> | null
-  ): SWRResponse<Data, Error>
-  <Data = any, Error = any>(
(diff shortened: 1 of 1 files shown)


promer94:
I think we might still need to keep this overload here for compatibility https://github.com/vercel/swr/pull/1640#issuecomment-974299780

Yuddomack:
Thanks for the quick feedback. I'll check it out!

phaux:
I understand the compatibility concerns but why can't it be changed in 2.0.0-beta already?

PR #2074, opened 2022-07-15
Fix: raf return type

### LINKED ISSUE #2073: vite build types error
image vite 
Error while building

**node_modules/swr/_internal/dist/_internal/utils/env.d.ts:4:76 - error TS2694: Namespace 'global.NodeJS' has no exported member 'Timeout'.

4 export declare const rAF: (f: (...args: any[]) => void) => number | NodeJS.Timeout;**

huozhi:
Fixes #2073

### DIFF
--- a/_internal/utils/env.ts
+++ b/_internal/utils/env.ts
@@ -6,7 +6,9 @@ export const IS_REACT_LEGACY = !React.useId
 export const IS_SERVER = !isWindowDefined || 'Deno' in window
 
 // Polyfill requestAnimationFrame
-export const rAF = (f: (...args: any[]) => void) =>
+export const rAF = (
+  f: (...args: any[]) => void
+): number | ReturnType<typeof setTimeout> =>
   hasRequestAnimationFrame()
     ? window['requestAnimationFrame'](f)
     : setTimeout(f, 1)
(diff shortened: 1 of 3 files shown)


_internal/utils/env.ts
```diff
@@ -6,7 +6,9 @@ export const IS_REACT_LEGACY = !React.useId
 export const IS_SERVER = !isWindowDefined || 'Deno' in window
 
 // Polyfill requestAnimationFrame
-export const rAF = (f: (...args: any[]) => void) =>
+export const rAF = (
+  f: (...args: any[]) => void
+): number | ReturnType<typeof setTimeout> =>
```
promer94, line 11:
```suggestion
export const rAF: (f: (...args: any[]) => any) => any = (f) =>
  hasRequestAnimationFrame()
    ? window['requestAnimationFrame'](f)
    : setTimeout(f, 1)
```
Since it's just a internal function. Let's keep it simple.

promer94 (changes requested):

_internal/utils/env.ts
huozhi: it doesn't hurt to give the explict type, and I feel it's simpler

promer94 (approved):

Issue #1240, opened 2021-06-24
E2E tests


shuding:
It would be great if SWR has some e2e tests covered, since SWR uses a lot of navigator specific APIs, and they varies between environments. We should explore adding e2e tests of Chrome, Firefox, Safari, Safari iOS and React Native to SWR.

Issue #350, opened 2020-04-22
Browser support information 


wanliyunyan:
For example ，Chrome > ??? ,edge >??? , firefox > ??? , it is very important

PR #2090, opened 2022-07-28
test: setup playwright component test

promer94:
## Motivation
This pr intends to setup playwright component test for swr. So we can add e2e test easily in the future. 
Also swr uses a lot of navigator specific APIs. It would be batter if we could use real browser enviorment for testing.

## Benefit
When Playwright Test is used to test web components, tests run in Node.js, while components run in the real browser. This brings together the best of both worlds: components run in the real browser environment, real events are triggered. At the same time, test can use all the powers of Node.js as well as all the Playwright Test features. 

related: #1240 #350

https://playwright.dev/docs/test-components


test/common-utils.ts
```diff
@@ -0,0 +1,19 @@
+export const createKey = () => 'swr-key-' + ~~(Math.random() * 1e7)
```
promer94, line 1: I have to create a new utils file here
otherwise the playwright will import `React Testing Library` into e2e code which would cause error

huozhi:
Can we add some description for this PR like: 
* what the e2e suite does?
* what kind of the cases are covered? 
* why we need to have this? (benefits or comparsion with current UT suite)

It's super helpful and convenient for others who are not familiar with it to get more insights

Issue #2102, opened 2022-08-03
Silent error in getKey within useSWRInfinite


onlyanegg:
# Bug report

## Description / Observed Behavior

useSWRInfinite hides errors within getKey (at least on the first page load). The error in getKey seems to cause the fetcher not to run. Because the error is hidden, it's difficult to debug.

## Expected Behavior

I expect it to raise or display the error

## Repro Steps / Code Example

```ts
import useSWRInfinite from "swr/infinite";

useSWRInfinite(                    
  () => {                          
    throw new Error();             
  },                               
  () => fetch("http://example.com")
);                                 
```

## Additional Context

swr@1.3.0

quantran020173:
This I a feature name "Conditional Fetching", not a bug https://swr.vercel.app/docs/conditional-fetching. And why do you want to throw an error in getKey?

onlyanegg:
This isn't about conditional fetching; it's about hiding an error. I don't want to raise an error in the getKey function; this was a minimal example for reproduction. When there is an error in the getKey function, it's hidden which makes it difficult to debug. I'd like to either raise that error or log it.

PR #2103, opened 2022-08-03
Fix: log error in getKey #2102

onlyanegg:
Logs caught errors in getKey within useSWRInfinite


koba04:
Throwing an error in the key function is a valid pattern in SWR, as described in https://beta-swr.vercel.app/docs/conditional-fetching. So I'd say logging errors would be annoying.

onlyanegg:
Yeah, I'm aware of that pattern. My point is that we're silently catching these errors, which makes it hard to debug.

PR #2104, opened 2022-08-05
types: add generic for SWRInfiniteKeyLoader

jason89521:
## Description

Sometimes I need to declare a `getKey` function individually, but the original type of `SWRInfiniteKeyLoader` use `any | null` for the second argument. For better type definition, I think this type should have generic. 

## Additional infomation

It seems that `"swr/_internal": ["./_internal/index.ts"]` is missing in `compilerOptions.paths` for tsconfig.ts?

### DIFF
--- a/infinite/types.ts
+++ b/infinite/types.ts
@@ -17,9 +17,9 @@ export type SWRInfiniteFetcher<
     : never
   : never
 
-export type SWRInfiniteKeyLoader = (
+export type SWRInfiniteKeyLoader<Data = any> = (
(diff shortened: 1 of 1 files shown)


huozhi (approved):
Thanks!

huozhi:
> It seems that "swr/_internal": ["./_internal/index.ts"] is missing in compilerOptions.paths for tsconfig.ts?

Nice catch! Feel free to submit a PR to add it 👍

PR #2105, opened 2022-08-05
fix: add swr/_internal to paths

jason89521:
## Description

Add the missing paths

https://github.com/vercel/swr/pull/2104#issuecomment-1206242665


huozhi (approved):
Thanks!

PR #2115, opened 2022-08-11
type: make tsconfig simpler and fix type error

promer94:


mutation/types.ts
```diff
@@ -29,14 +29,14 @@ export type SWRMutationConfiguration<
     data: Data,
     key: string,
     config: Readonly<
-      SWRMutationConfiguration<Data, Error, SWRMutationKey, ExtraArg>
+      SWRMutationConfiguration<Data, Error, ExtraArg, SWRMutationKey>
```
huozhi, line 32: Can we have a test for this?

promer94: We don't need a test here beacaue `tsc` will report this error (but only in 4.8, i find this error in editor). 

Currently we use TS 4.7, so our ci passed.

I just fix them in advance. 😁

huozhi (approved):

Issue #2117, opened 2022-08-12
clear missing on cache interface


swushi:
Hi! 👋 

In the documentation it is specified that you can clear the cache [here](https://swr.vercel.app/docs/advanced/cache#access-to-the-cache) using:

```js
cache.clear();
```

However, `clear` is not included in the `Cache` interface. This is a patch I have been using. 

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/swr/dist/types.d.ts b/node_modules/swr/dist/types.d.ts
index d97de4e..f50cdc8 100644
--- a/node_modules/swr/dist/types.d.ts
+++ b/node_modules/swr/dist/types.d.ts
@@ -104,5 +104,6 @@ export interface Cache<Data = any> {
     get(key: Key): Data | null | undefined;
     set(key: Key, value: Data): void;
     delete(key: Key): void;
+    clear(): void;
 }
 export {};
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

huozhi:
We released a new API change latest beta [v2.0.0-beta.5](https://github.com/vercel/swr/releases/tag/2.0.0-beta.5) that could let you do the following code to remove all cache data much easier:

```js
// declaration 
// mutate(keyMatcher, data, revalidate)

mutate(/* key matcher */ (key) => true, undefined, false)
```

Notice that the functional `key` is replaced by a `key filter` function. That's the better way than `cache.clear` because it could keep the state sync with the actual cache provider. We don't recommend to manipulate the cache directly in the codebase.

The v2 docs are still in progress, the more details about the mutate API change are in PR https://github.com/vercel/swr/pull/1989 or RFC https://github.com/vercel/swr/issues/1946

The idea was to leverage the existing mutate API but extend the ability to partially mutat selected keys. That example calls means: Use `() => true` as a filter is to match all the existing keys and mutate them to `undefined`, without revalidation.

We'll give more examples about that change in v2 docs later.

PR #2118, opened 2022-08-12
fix: add clear to cache interface

swushi:
#2117

### DIFF
--- a/_internal/types.ts
+++ b/_internal/types.ts
@@ -376,6 +376,7 @@ export interface Cache<Data = any> {
   get(key: Key): State<Data> | undefined
   set(key: Key, value: State<Data>): void
   delete(key: Key): void
+  clear(): void
 }
 
 export interface StateDependencies {


huozhi:
We released a new API change latest beta [v2.0.0-beta.5](https://github.com/vercel/swr/releases/tag/2.0.0-beta.5) that could let you do the following code to remove all cache data much easier:

```js
// declaration 
// mutate(keyMatcher, data, revalidate)

mutate(/* key matcher */ (key) => true, undefined, false)
```

Notice that the functional `key` is replaced by a `key filter` function. That's the better way than `cache.clear` because it could keep the state sync with the actual cache provider. We don't recommend to manipulate the cache directly in the codebase.

The v2 docs are still in progress, the more details about the mutate API change are in PR https://github.com/vercel/swr/pull/1989 or RFC https://github.com/vercel/swr/issues/1946

The idea was to leverage the existing mutate API but extend the ability to partially mutat selected keys. That example calls means: Use `() => true` as a filter is to match all the existing keys and mutate them to `undefined`, without revalidation.

We'll give more examples about that change in v2 docs later.

PR #2140, opened 2022-08-31
test: add tests for dependency collection

koba04:
This adds tests for the dependency collection behavior, which is described in the docs.
https://swr.vercel.app/docs/advanced/performance#dependency-collection

I'm working on updating the docs, so I'd like to add this to guarantee the behavior by tests.

### DIFF
--- a/test/use-swr-integration.test.tsx
+++ b/test/use-swr-integration.test.tsx
@@ -470,4 +470,103 @@ describe('useSWR', () => {
     await act(() => sleep(150))
     expect(count).toBe(2)
   })
+
+  // Test for https://swr.vercel.app/docs/advanced/performance#dependency-collection
+  it('should render four times in the worst case', async () => {
+    let isFirstFetch = true
+    const fetcher = async () => {
+      if (isFirstFetch) {
+        isFirstFetch = false
+        throw new Error('error')
+      }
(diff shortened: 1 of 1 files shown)


test/use-swr-integration.test.tsx
```diff
@@ -470,4 +470,103 @@ describe('useSWR', () => {
     await act(() => sleep(150))
     expect(count).toBe(2)
   })
+
+  // Test for https://swr.vercel.app/docs/advanced/performance#dependency-collection
+  it('should render four times in the worst case', async () => {
```
promer94, line 475: https://github.com/reactwg/react-18/discussions/17
we could use `<Profiler />` to count the number of times a component rendered. Then the test will still work in strict mode. 

https://github.com/vercel/swr/blob/8bfa5da876b4948bdaec5359479c0aa062ee74fb/test/use-swr-integration.test.tsx#L458-L465

koba04: @promer94 
Thank you! I didn't know the component 👍 

In this test, I'd like to assert the return values too, and the test requires reading return values within rendering so that SWR can mark the values as used because it's a test for dependency collection.

promer94 (approved):

PR #2141, opened 2022-08-31
test: add more isLoading tests for the case a key is null or an error

koba04:
This PR adds more tests for `isLoading`.


koba04:
This includes irrelevant diff by the formatter that runs at commits.

promer94 (approved):

PR #2149, opened 2022-09-08
memoize config values to avoid frequent context propagation

gnoff:
every time `<SWRConfig>` renders it propagates a context change even if the config did not change functionally. This can break hydration of Suspense boundaries when this component re-renders shortly after initial render which is common when there are mount effects for root level components.

This PR adds memoization so the config only changes if it's reactive dependencies change


huozhi (approved):
Thanks!

PR #2154, opened 2022-09-10
enable declarationMap flag

fibo:
Go to JS file when using IDE functions like "Go to Definition" in VSCode.

See https://www.typescriptlang.org/tsconfig#declarationMap

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
     "declaration": true,
+    "declarationMap": true,
     "esModuleInterop": true,
     "jsx": "react",
     "lib": ["esnext", "dom"],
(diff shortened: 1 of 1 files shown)


promer94:
It won't work since the source code is not included in npm packages.

fibo:
@promer94 thanks for spotting the `files` attribute content. 

> they should not be included in a published NPM package

sounds like an opinion, many developers jump to definitions in any context.

huozhi:
Thanks for the PR. We expect users to use the published and bundled types which tree-shake out the internal types and only keeps the exported types. like [types of swr core](https://unpkg.com/browse/swr@2.0.0-beta.7/core/dist/index.d.ts). Since the source files may change and may have more inrelavant or internal purposed types, we don't want users to jump between the original sources.

PR #2160, opened 2022-09-14
chore: fix missing husky warning

huozhi:
Seeing error `Can't find Husky, skipping post-checkout hook` when you commit.

Fix it by adding the `.husky` precommit and remove the legacy config in package.json according to the new tutorial in husky

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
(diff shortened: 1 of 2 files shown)


promer94 (changes requested):
I think we need also add prepare script for husky https://typicode.github.io/husky/#/?id=install

```
{
  "scripts": {
    "prepare": "husky install"
  }
}
```
Otherwise `_/husky.sh` will be not be generated

promer94 (approved):

PR #2182, opened 2022-10-04
breaking: Change the error broadcasting behavior in mutations and add `throwOnError` option

shuding:
Errors from mutations should be separated from direct fetches. Currently, the returned and shared `error` state from `useSWR` will be mutated by `mutate` calls:

```js
const { error } = useSWR('/api')

mutate('/api', mutator) → this might update `error` if it fails
```

The main use case for that is to have a global and declarative error indicator for mutations, but then it conflicts with fetching errors (e.g. resource fails to load). With the new `useSWRMutation` hook, that is no longer needed because we can have a cleaner representation for both:

```js
const { error: loadError } = useSWR('/api')
const { error: actionError } = useSWRMutation('/api')
```

And they will be separated and scoped to different places:

image

For example (1) and (2) will always share the same data and error. So if the user fails to load initially, error indicators can be shown in both places (or you can throw to trigger an upper-level error boundary). In (3) and (4) you can trigger remote mutations to update the user resource. If the mutation succeeded, (1) and (2) will be updated and rerendered with the new data. If a mutation from (3) failed, only (3) will get the error state so you can show the error indicator there, without affecting (1) (2) and (4).

## `throwOnError`

Since there is already `onError` supported in `useSWRMutation`, here we are adding a new option `throwOnError` to mutations (both `useSWRMutation` and `mutate`) to define if an error should be thrown inline or not. It defaults to `true`.

```js
// This call will not throw an error if it fails. It will fail silently.
mutate(key, fetcher, {
  throwOnError: false
})
```

```js
const { trigger, error } = useSWRMutation('/api', mutator, {
  throwOnError: false,
  onError: (err) => { ... }
})

// This call will not throw inline
trigger()

// Override the option:
try {
  await trigger(null, { throwOnError: true })
} catch (err) {
  // ...
}
```

### DIFF
--- a/mutation/index.ts
+++ b/mutation/index.ts
@@ -7,7 +7,8 @@ import {
   withMiddleware,
   useIsomorphicLayoutEffect,
   UNDEFINED,
-  getTimestamp
+  getTimestamp,
+  mergeObjects
 } from 'swr/_internal'
 import type {
   SWRMutationConfiguration,
@@ -43,11 +44,14 @@ const mutation = (<Data, Error>() =>
           throw new Error('Can’t trigger the mutation: missing fetcher.')
         }
         if (!serializedKey) {
-          throw new Error('Can’t trigger the mutation: key isn’t ready.')
+          throw new Error('Can’t trigger the mutation: missing key.')
         }
 
         // Disable cache population by default.
-        const options = Object.assign({ populateCache: false }, config, opts)
+        const options = mergeObjects(
+          mergeObjects({ populateCache: false, throwOnError: true }, config),
+          opts
+        )
 
         // Trigger a mutation, and also track the timestamp. Any mutation that happened
         // earlier this timestamp should be ignored.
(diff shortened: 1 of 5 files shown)


huozhi (approved):

mutation/index.ts
```diff
@@ -72,11 +76,14 @@ const mutation = (<Data, Error>() =>
           }
           return data
         } catch (error) {
-          // If it's reset after the mutation, we don't broadcast any state change.
+          // If it's reset after the mutation, we don't broadcast any state change
+          // or throw because it's discarded.
           if (ditchMutationsUntilRef.current <= mutationStartedAt) {
             setState({ error: error as Error, isMutating: false })
             options.onError?.(error as Error, serializedKey, options)
```
koba04, line 83: Is `onError` always called regardless of `throwOnError`?

> should call `onError` event and skip throwing the error when `throwOnError` is disabled

https://github.com/vercel/swr/pull/2182/files#diff-665896fca56ef021ec0b6b13ffb986551f53f66c78ba58ac25ab8744084d3daeR168

I guess `onError` is not called with `throwOnError: true` from the test description.

koba04 (approved):

utg1-tdawg:
@shuding Is there a way to set `throwOnError` globally? Seems like `SWRConfig` doesn't accept it currently.

silcro-123:
> @shuding Is there a way to set `throwOnError` globally? Seems like `SWRConfig` doesn't accept it as an argument.

i have same issue.

PR #2188, opened 2022-10-09
Memoize `isEqual` function 🛩️ 

sanjaiyan-dev:
Memoize `isEqual` function for minor performance improvement.

https://reactjs.org/docs/hooks-reference.html#usecallback

_Sorry if I made any mistakes :(_


shuding (commented):
This does not improve the performance, because the current `isEqual` is not causing any extra re-renders.

sanjaiyan-dev:
> This does not improve the performance, because the current `isEqual` is not causing any extra re-renders.

Sorry & Ok

PR #2189, opened 2022-10-10
Memoizing the return values 🛥️ 

sanjaiyan-dev:
Memoizing the return values using [`React.useMemo`](https://beta.reactjs.org/apis/react/useMemo) to prevent from unnecessary computations when, any unnecessary state changes occurs or when parent component re reder.

_Extremely sorry if I made any mistakes :(_

### DIFF
--- a/core/use-swr.ts
+++ b/core/use-swr.ts
@@ -627,7 +627,7 @@ export const useSWRHandler = <Data = any, Error = any>(
       stateDependencies.isLoading = true
       return isLoading
     }
-  }), [boundMutate, returnedData, error, isValidating, isLoading]) as SWRResponse<Data, Error>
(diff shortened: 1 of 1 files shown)


feledori:
Seeing that this is you second pull request with a memoization of something that probably doesn't need to be memoized, I would recommend you reading this article by Dan Abramov [Before You memo()](https://overreacted.io/before-you-memo/). This is a really nice blog post explaining how to get around memoizing and why it should be avoided until its really necessary. :)

sanjaiyan-dev:
> Seeing that this is you second pull request with a memoization of something that probably doesn't need to be memoized, I would recommend you reading this article by Dan Abramov [Before You memo()](https://overreacted.io/before-you-memo/). This is a really nice blog post explaining how to get around memoizing and why it should be avoided until its really necessary. :)

Tks a lot 🤝💪

PR #2207, opened 2022-10-21
Fix: exclude `swr` and `swr/_internal` from sub-packages

### LINKED ISSUE #2205: [BUG] `swr/core` and `swr/_internal` are incorrectly packed into sub-packages
# Bug report

After upgrading bunchee #2187. 

`swr/core` and `swr/_internal` are incorrectly packed into sub-packagaes

### Before 
* `immutable/dist/index.mjs` ~ 486 bytes
 
before

### Now
* `immutable/dist/index.mjs` ~ 70,834 bytes 
after

promer94:
close #2205 

## Problem

Since [rollup-plugin-swc 0.6.0](https://github.com/SukkaW/rollup-plugin-swc/releases/tag/0.6.0) now supports `extends` in `tsconfig.json.` 

It will forward root's `tsconfig.json`'s `path` to `@swc/core`.  So the absolute import statement will be transformed to reletive import statement, then rollup will pack unnecessary content into final output.

### Previous tsconfig which was used by `@swc/core`
```json
// swr/immutatble/tsconfig.json
{
  "extends": "../tsconfig.json",
  "compilerOptions": {
    "outDir": "./dist"
  },
  "include": ["./*.ts"]
}
```

### Current tsconfig which is used by `@swc/core`

```json
// swr/immutatble/tsconfig.json
{
  "extends": "../tsconfig.json",
  "compilerOptions": {
    "outDir": "./dist",
    "paths": {
      "swr": ["./core/index.ts"],
      "swr/infinite": ["./infinite/index.ts"],
      "swr/immutable": ["./immutable/index.ts"],
      "swr/mutation": ["./mutation/index.ts"],
      "swr/_internal": ["./_internal/index.ts"]
    },
  },
  "include": ["./*.ts"]
}
```

```ts
import useSWR from 'swr'
// will be replaced as import useSWR from './xxx'

import _internal from 'swr/_internal'
// will be replaced as import useSWR from './xxx'
```

## Solution

* remove `tsconfig.json`'s `path`. So `swc` will not replace origin import statement
* create another `tsconfig.check.json` which contains the original `path` configs as base tsconfig for editor.
* continue the work of #2137.


huozhi:
> [rollup-plugin-swc 0.6.0](https://github.com/SukkaW/rollup-plugin-swc/releases/tag/0.6.0) will forward tsconfig.json's path to @swc/core. The absolute import statement will be transformed to reletive import statement, then rollup will pack unnecessary content into final output.

@sukkaw do you feel this is the side effect of the upgration from 0.5.x to 0.6.x you did last time?

package.json
```diff
@@ -11,53 +11,63 @@
     "fetch"
   ],
   "packageManager": "pnpm@7.1.0",
-  "main": "./core/dist/index.js",
-  "module": "./core/dist/index.esm.js",
-  "types": "./core/dist/index.d.ts",
+  "main": "./dist/index.js",
+  "module": "./dist/index.mjs",
+  "types": "./dist/index.d.ts",
+  "typesVersions": {
```
promer94, line 17: It requires TS 4.8 to resolve `types` from `exports` field.

So typesVersion are used here for backward compatibility.

package.json
```diff
@@ -11,53 +11,63 @@
...
+      "require": "./dist/index.js"
     },
     "./infinite": {
-      "types": "./infinite/dist/index.d.ts",
-      "import": "./infinite/dist/index.mjs",
-      "module": "./infinite/dist/index.esm.js",
-      "require": "./infinite/dist/index.js"
+      "types": "./dist/infinite.d.ts",
+      "import": "./dist/infinite.mjs",
+      "module": "./dist/infinite.mjs",
```
promer94, line 47: I'm not sure why [bunchee](https://github.com/huozhi/bunchee) currently can't produce output with `.esm.js`  extensions.  But we could fix this later

SukkaW:
> > [rollup-plugin-swc 0.6.0](https://github.com/SukkaW/rollup-plugin-swc/releases/tag/0.6.0) will forward tsconfig.json's path to @swc/core. The absolute import statement will be transformed to reletive import statement, then rollup will pack unnecessary content into final output.
> 
> @SukkaW do you feel this is the side effect of the upgration from 0.5.x to 0.6.x you did last time?

`rollup-plugin-swc3` supports forwarding `baseUrl` and `paths` to `@swc/core` since `0.4.1`.

package.json
huozhi: "module" field with ".mjs" extension breaks react native bundler metro before, so we keep a different extension here.

huozhi:
We still want to keep the package.json for each subfolder in v2 to avoid some bundler still can't handle the "exports" field, so subpath folder package.json can still resolve the assets dynamically

Issue #2155, opened 2022-09-11
UseSWR wrapper doesnt override and routes to the wrong fetcher (SWR global config's)


ridhwaans:
# Bug report

## Description / Observed Behavior

As the title says, I have a useSharedState hook that is used to pass state between components without prop drilling. This hook is not for http resources, it doesnt have a url fetcher but has an initial state option
```
import useSWR from 'swr'

export const useSharedState = (key, initial) => {
  const { data: state, mutate: setState } = useSWR(key, null, {
    fallbackData: initial,
  })

  return [state, setState]
}
```

I also have SWR global config in my App.js with its own fetcher
```
import useSWR, { SWRConfig } from 'swr'

const App = () => {

    return (
        <React.Fragment>
            <SWRConfig
                value={{
                    fetcher: (resource, init) => fetch(`${process.env.REACT_APP_BASE_URL}/api` + resource, init).then(res => res.json())
                }}
            >
                <BrowserRouter basename={process.env.PUBLIC_URL}>
                    <Routes>
                    ...
```

When I make an `useSWR` call it adds the base url to query endpoints in components

However when I have `useSWR` and  `useSharedState` in the same component or when I call `useSharedState` directly, it goes to the wrong fetcher & url endpoint defined thats in SWRConfig

```
import useSWR from 'swr'
import { useSharedState } from "../../hooks/useSharedState"
...
function Books() {
    
    useEffect(() => {
        document.documentElement.className = "inset";
        document.body.className = "inset-body"; 
    }, [])

    const {data : recentlyAddedBooks } = useSWR('/books/recently_added');
    const [searchInput, setSearchInput] = useSharedState('searchInput') // used for getting text from a searchbox
    
    return (...)
```

it returns 
```
`GET http://localhost:5000/apisearchInput 404 (Not Found)` because its not a resource url
```

## Expected Behavior

Need all SWR calls with resource urls i.e. `useSWR('/books')` to follow the global fetcher in `<SWRConfig/>` but other SWR wrappers like `useSharedState()` that is not meant for API calls should not use global fetcher option 

What am I doing wrong, do I need to fix the hook param to override fetcher?

## Additional Context

swr version `"^1.3.0",`

koba04:
@ridhwaans 
Thank you for reporting this.
A workaround for this is to pass `fetcher: null` to the `useSWR` in `useSharedState`. This avoids overriding `useSWR`'s fetcher function by the global `fetcher` option.

```jsx
[...]

PR #2240, opened 2022-11-12
fix: pass null fetcher to middleware

### LINKED ISSUE #2217: When providing fetcher: null , middleware is defaulting it to undefined
# Bug report

We're building a custom middleware were we interact with fetcher. 
We check if fetcher === null, but we receive fetcher as undefined inside middleware callback.

```ts
 const myCustomMiddleware = (useSwrNext) =>
  (key, fetcher, config) => {
    const extendedFetcher =  fetcher === null ? null : myFetcher
    ....
```
The issue is that `fetcher` comes undefined even if we set it as null in our `useSWRImmutable("myKey", null, options)` 
Types doesn't allow us to define fetcher as undefined in `useSWRInmutable`.

## Expected Behavior

Middleware returns `fetcher` as `null` when we set it `null` in `useSWRInmutable`

SWR version. 
 "swr": "^1.3.0"

koba04:
Fixes #2217

`null` is a valid type of `fetcher` argument, but the `null` fetcher becomes `undefined` in middleware. This is an unexpected behavior for middleware authors, so I've fixed it.

### DIFF
--- a/_internal/utils/resolve-args.ts
+++ b/_internal/utils/resolve-args.ts
@@ -24,6 +24,6 @@ export const withArgs = <SWRType>(hook: any) => {
       next = middleware[i](next)
     }
 
-    return next(key, fn || config.fetcher, config)
+    return next(key, fn || config.fetcher || null, config)
   } as unknown as SWRType
 }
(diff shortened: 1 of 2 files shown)


promer94 (approved):

Dragate:
Does this also resolve #2155 ?

koba04:
@Dragate No it doesn't. But I'll consider the issue. Thank you!

shuding (approved):

PR #2249, opened 2022-11-24
Update build target to es2018

shuding:
For most use cases of SWR, users either have a modern browser (e.g. using modules in the browser + CDN), or have a bundler (esbuild, Webpack). If you still need to target old browsers, it's easy to target a lower version with the bundler.

This change makes sense for the 2.0 version. For the majority this improves the performance of SWR by a lot. According to #2095, the new bundle size will be ~80% of the `es5` target. And measured by the [SSR benchmark](https://github.com/pwrdrvr/swr-server-side), switching from `Object.assign` to the native spread operator makes SSR 4.7x faster (P50) in Node.js 16. This scales linearly with the app size, because each `useSWR()` call contains around 3 `Object.assign` operations.

### DIFF
--- a/_internal/package.json
+++ b/_internal/package.json
@@ -5,8 +5,8 @@
   "exports": "./dist/index.mjs",
   "private": true,
   "scripts": {
-    "watch": "bunchee index.ts -w",
-    "build": "bunchee index.ts",
+    "watch": "bunchee --target es2018 index.ts -w",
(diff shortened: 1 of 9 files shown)


koba04 (approved):
👍 LGTM!! That's a great optimization!

huozhi (approved):

PR #2273, opened 2022-12-11
types: make FullConfiguration generic

huozhi:
Make `FullConfiguration` as generic for better type inference. With this change, `fallbackData` from config could be inferred as `Data | undefined` instead of just `any` inside useSWR hook


shuding (commented):
Is it possible to add a test to demonstrate this?

Issue #1933, opened 2022-04-17
Cancelation


shuding:
⚠️ This proposal isn't finalized yet.

An abort signal can be passed to `fetcher`:

```js
useSWR('/api', (url, { signal }) => {
  return fetch(url, { signal }).then(res => res.json())
})
```

And SWR will abort the signal when the request is discarded and the result is no longer needed. After aborted, SWR ignores the `AbortError` thrown.

700software:
This is amazing! Elegant and in spirit with SWR's simplicity.

It would cause aborts to happen in applications where none were happening before. But is it a breaking change? I think not.

l-mbert:
Looks good!
Maybe add a new configurarion option for People who don't want it?

CaptainPants:
I have given implementation of this a go in the above PR if anyone is competent to have a look and give any feedback? Cheers.

PR #2275, opened 2022-12-12
Issue 1933 - Cancellation of requests when no longer needed

CaptainPants:
First attempt at #1933 - adding a cancellation mechanism to fetchers. 

Looking for feedback on the approach and if there is any huge landmines I'm missing. I won't be offended if this is just a big 'no'.

Still working on how to appropriately test this.

TODO:
- Not sure how error handling works as this will cause the await to throw a DOMException from the abort call, and don't want this to send the hook into an error state.
- Testing.
- Tests - fix existing ones and add new ones related to the change.
- Add an example?

### DIFF
--- a/_internal/types.ts
+++ b/_internal/types.ts
@@ -4,7 +4,15 @@ import type { defaultConfig } from './utils/config'
 export type GlobalState = [
   Record<string, RevalidateCallback[]>, // EVENT_REVALIDATORS
   Record<string, [number, number]>, // MUTATION: [ts, end_ts]
-  Record<string, [any, number]>, // FETCH: [data, ts]
+  Record<
+    string,
+    [
+      data: any,
+      timestamp: number,
+      revalidateCount: number,
+      abort: AbortController | undefined
+    ]
+  >, // FETCH: [data, timestamp, revalidateCount, abort]
   Record<string, FetcherResponse<any>>, // PRELOAD
   ScopedMutator, // Mutator
   (key: string, value: any, prev: any) => void, // Setter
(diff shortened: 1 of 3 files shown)


_internal/types.ts
```diff
@@ -189,6 +206,12 @@ export interface PublicConfiguration<
    * @link https://swr.vercel.app/docs/advanced/react-native#customize-focus-and-reconnect-events
    */
   isVisible: () => boolean
+
+  /**
+   * Enable this to pass an instance of AbortSignal to your fetcher that will be aborted when there no useSWR hooks remaining that are listening for the result of the fetch (as
+   * a result for example of navigation, or change of search input). This allows you to cancel a long running a request if the result is no longer relevant, e.g. when doing a search.
+   */
+  abortDiscardedRequests?: boolean
```
700software, line 214: I would like for this feature to be enabled by default. Not the other way around.

Let's emphasize what the new users want rather than reverse compatibility since the risk of problems is low.

Would a new user who has no prior version of swr installed prefer this on by default or not? I think yes but I'm only one user.

davidnormo:
Is there anything missing from this preventing it from being merged?

oanaOM:
This would be very helpful to have! Any ideas when will this get released?
Meanwhile, is there other workaround to make SWR work with AbortController or achieve a cancelling mechanism?

scarqin:
Exciting features! looking forward to launch.

_internal/types.ts
```diff
@@ -14,15 +22,24 @@ export type FetcherResponse<Data = unknown> = Data | Promise<Data>
...
   ...args: any[]
 ) => FetcherResponse<Data>
+/**
+ * Second parameter provided to fetcher functions. At present this only allows
+ * for 'signal' for cancelling a request in progress, but allows for future
+ * enhancements.
+ */
+export interface FetcherOptions {
+  signal?: AbortSignal
+  [key: string]: unknown
```
nikeee, line 32: I'd omit an index signature because it may hide typos for the valid options (only `signal` in this case). It also prevents us from adding properties without introducing a potential breaking change.

PR #2283, opened 2022-12-13
type: enforce array literal as tuple

promer94:
In TypeScript > 4.9, if we use an array literal as key. 

```tsx
useSWR([{ a: '1', b: { c: '3' } }, [1231, '888']], arg => arg)
```
The `arg` will be infered as `((string | number)[] | {
    a: string;
    b: {
        c: string;
    };
})[]`

Screenshot 2022-12-13 at 15 17 42

This pr will enforce array literal to be infered as tuple. The type of `args` will be `[{
    a: string;
    b: {
        c: string;
    };
}, (string | number)[]]`

Screenshot 2022-12-13 at 15 18 18


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-pnpm lint-staged && pnpm types:check
+pnpm lint-staged && pnpm types:check && pnpm test-typing
```
huozhi, line 4: this gonna be long time for each commit, can we let CI do it?

_internal/types.ts
```diff
@@ -259,7 +275,8 @@ export type Arguments =
   | undefined
   | false
 export type Key = Arguments | (() => Arguments)
-
+export type StrictTupleKey = ArgumentsTuple | null | undefined | false
```
huozhi, line 278: is `StrictTupleKey` has to be exported?

_internal/types.ts
```diff
@@ -259,7 +275,8 @@ export type Arguments =
...
+export type StrictKey = StrictTupleKey | (() => StrictTupleKey)
```
huozhi, line 279: Is that possible to replace the `Key` with this?

promer94: Since we still need to support `useSWR('/api')` and `useSWR({ foo: 'bar' })`, we have to use different type here. 

Maybe we could deprecated `useSWR('/api')` and `useSWR({ foo: 'bar' })` in next major release ?

huozhi (approved):
Thanks!

PR #2286, opened 2022-12-14
feat: add extra arg in onSuccess, onError

Lee-sungheon:
Hi!
While using swr 2.0, I felt uncomfortable that extgraArg could not be used in onSuccess and onError.
So I suggest you can use extraArg in onSuccess and onError like this

```ts
async onSuccess(data, key, config, extraArg) {
  ...

  await mutate('otherKey', extraArg.pk);
},
async onError(error, key, config, extraArg) {
  ...

  await mutate('otherKey', extraArg.pk);
},
```


mutation/types.ts
```diff
@@ -38,14 +38,16 @@ export type SWRMutationConfiguration<
     key: string,
     config: Readonly<
       SWRMutationConfiguration<Data, Error, ExtraArg, SWRMutationKey, SWRData>
-    >
+    >,
+    extraArg: any
```
promer94, line 42:
```suggestion
extraArg:  ExtraArg
````

File: mutation/types.ts [hunk 2]
```diff
@@ -38,14 +38,16 @@ export type SWRMutationConfiguration<
...
   ) => void
   onError?: (
     err: Error,
     key: string,
     config: Readonly<
       SWRMutationConfiguration<Data, Error, ExtraArg, SWRMutationKey, SWRData>
-    >
+    >,
+    extraArg: any
```
promer94, line 50:
```suggestion
extraArg:  ExtraArg
````

**Review by promer94 — CHANGES_REQUESTED** [review]

**Comment by promer94** [comment]
I realize there is a alternative way to achieve this.  

```tsx
const { trigger, isMutating } = useSWRMutation('/api/user', sendRequest)
trigger(arg, { onSuccess: () => {
   mutate('other key', arg)
}})
```

You could config `onSuccess` by `trigger`.

Lee-sungheon:
Good alternative way but I want to create a custom hook using useSWRMutation.

When a custom hook is created in a way that uses a trigger, not only does the code become complicated, but it also becomes difficult to further customize the trigger.
```typescript
export const useSendRequest = () => {
  const { trigger, ...others } = useSWRMutation('/api/user', sendRequest)
  const myTrigger = (arg) => trigger(arg, { onSuccess: () => {
    mutate('other key', arg)
  }})

  return {
    trigger: myTrigger,
    ...others
  }
}
```

However, when a custom hook is created as follows, not only the readability of the code is improved, but also the trigger is additionally customizable.
```typescript
export const useSendRequest = () => {
  return useSWRMutation('/api/user', sendRequest, {
    onSuccess: (data, key, config, extraArg) => {
      mutate('other key', extraArg)
    }
  })
}
```

Lee-sungheon:
@huozhi @shuding @promer94 
Can you check the PR?
This is a must-have feature for me.

Issue #1907, opened 2022-04-06
Retry on error with suspense


micha149:
# Bug report

I use `swr` with `graphql-request` and I am struggling with the automatic retry on an api error. Only a single request is made until the error is raised to my ErrorBoundary. If I set `{ suspense: false }` two requests are made and the data shows up as expected.

## Expected Behavior

I would expect that `swr` handles the retry also in suspense mode. From the outside any retry should behave as a very long running request.

```tsx
const useSomeHook = (key: string, enableSuspense: boolean): string | undefined => {
    const calls = useRef<number>(0);

    const fetcher = (): string => {
        calls.current += 1;

        if (calls.current < 2) {
            throw new Error('Something\'s worng');
        }

        return 'yay';
    };

    const { data } = useSwr(key, fetcher, { suspense: enableSuspense, errorRetryInterval: 10 });

    return data;
};

describe('Retry on failing fetcher', () => {
    it('renders correctly without suspense', async () => {
        const { result, waitFor } = renderHook(() => useSomeHook('myKey', false));

        await waitFor(() => result.current !== undefined);

        expect(result.current).toEqual('yay'); // OK
    });

    it('renders correctly with suspense', async () => {
        const { result, waitForNextUpdate } = renderHook(() => useSomeHook('myOtherKey', true));

        await waitForNextUpdate();

        expect(result.current).toEqual('yay'); // Fails with `Something's worng`
    });
});
```

I created a codesandbox to hopefully make the problem a bit more understandable:
https://codesandbox.io/s/swr-suspesnse-retry-bug-sqpjyn?file=/src/App.js

## Additional Context
- `@testing-library/react-hooks@7.0.2`
- `graphql-request@4.2.0`
- `react-dom@17.0.2`
- `react@17.0.2`
- `swr@1.2.2`

Babailan:
this is a useful option if there is an error and you want to retry requests. if you get the data it will not request again.
```
useSWR(key, fetcher, {
  revalidateIfStale: false,
  revalidateOnFocus: false,
  revalidateOnReconnect: false,
  errorRetryInterval:0,
  shouldRetryOnError:true,
})
```

micha149:
@Babailan sorry for the delay. I was on vacation as you replied and missed to answer.

I don't understand what you mean. I am aware of the usefulness of the retry feature. My report is, that it seems not to work when enabling `suspense`.

Babailan:
[...]

PR #2287, opened 2022-12-14
Export SWRConfigContext

andyrichardson:
# About
Added export of SWRConfigContext to allow for
 - consuming context in class components (such as [error boundaries](https://reactjs.org/docs/error-boundaries.html))
 - mocking SWRConfigContext in fixtures/tests

It seems like retrying errors isn't currently possible with suspense (#1907). This change allows clearing/manipulating the cache before resuming from an error boundary.

# Example usage

```tsx
export class ErrorGuard extends Component<any, { error: Error | undefined }> {
  constructor(props: any) {
    super(props);
    this.state = { error: undefined };
  }

  public componentDidCatch(error: any, errorInfo: any) {
    this.setState({ error });
  }

  public render() {
    if (this.state.error === undefined) {
      return <Fragment {...this.props} />;
    }

    return (
      <SWRConfigContext.Consumer>
        {value => 
          <div>
            <div>Error</div>
            <button onClick={() => { value.cache.delete("somekey"); this.setState({ error: undefined }) }}>Retry</button>
          </div>
        }
      </SWRConfigContext.Consumer>
    );
  }
}
```

Additional functionality to automatically refetch errors on mount when using suspense would be a more useful option (can be done in a separate PR)


shuding:
I think the use case of class components makes sense! Wdyt @huozhi?

promer94 (dismissed):

huozhi:
I would recommend to write a helper component for that case with `useSWRConfig` like

```js
function SWRConfigConsumer({ children }) {
  const config = useSWRConfig();
  return children(config);
}

return (
  <SWRConfigConsumer>
     {value => ...}
  </SWRConfigConsumer>
)
```

This is doable with current APIs without exposing extra things but can achieve the same goal.

PR #2289, opened 2022-12-15
Improve type test

promer94:
* make typecheck more strict with `@type-challenges/utils`
* fix `SWRInfiniteHook` types
* simplify `SWRHook`

### DIFF
--- a/_internal/types.ts
+++ b/_internal/types.ts
@@ -275,7 +275,7 @@ export type Arguments =
   | undefined
   | false
 export type Key = Arguments | (() => Arguments)
-type StrictTupleKey = ArgumentsTuple | null | undefined | false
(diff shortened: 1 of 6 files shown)


_internal/types.ts
```diff
@@ -208,34 +208,18 @@ export type ProviderConfiguration = {
  * ```
  */
 export interface SWRHook {
-  <Data = any, Error = any, SWRKey extends StrictKey = StrictKey>(
+  <Data = any, Error = any, SWRKey extends Key = StrictKey>(
```
promer94, line 211: I use `StrictKey` here because i want 
```ts
useSWR<string, any>('/api/', (key: string) => key)
```  
to match a simpler overload signature

Screenshot 2022-12-15 at 18 06 55

instead of

huozhi (approved):

PR #2301, opened 2022-12-16
types: conditional swr response

### LINKED ISSUE #1410: typeof 'data' is always optional even with 'fallbackData' provided.
# Bug report

type of `data`

## Expected Behavior
type of `data` should be `Data`(not `Data | undefined`) if `fallbackData` has provided in type `Data`.

## Repro Steps / Code Example
```typescript
const { data } = useSWR<string>('my-key', {
  fallbackData: 'hello world',
});
```
image

SWR version. 
1.0.0

huozhi:
Resolves #1410

When `fallbackData` is specified for the swr hook, then:
 * it should match the `Data` type returning from `Fetcher`
 * swr hook should always return a fulfilled `data`
 * the `isLoading` always should be `false`

When `suspense` is specified for the swr hook, then:
 * swr hook should always return a fulfilled `data`
 * the `isLoading` always should be `false`

image

### DIFF
--- a/_internal/types.ts
+++ b/_internal/types.ts
@@ -21,7 +21,7 @@ export type Fetcher<
   ? (arg: Arg) => FetcherResponse<Data>
   : SWRKey extends null | undefined | false
   ? never
-  : SWRKey extends string | infer Arg
+  : SWRKey extends infer Arg
   ? (arg: Arg) => FetcherResponse<Data>
   : never
 
--- a/_internal/types.ts
+++ b/_internal/types.ts
@@ -25,11 +25,14 @@ export type Fetcher<
   ? (arg: Arg) => FetcherResponse<Data>
   : never
 
-export type DataCached<Config, Data = any> = Config extends undefined
+export type BlockingData<
+  Data = any,
+  Options = SWROptions<Data>
+> = Options extends undefined
   ? false
-  : Config extends { suspense: true }
+  : Options extends { suspense: true }
   ? true
-  : Config extends { fallbackData: Data }
+  : Options extends { fallbackData: Data }
   ? true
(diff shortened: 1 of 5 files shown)


promer94:
I like the idea. 
But i think we could achieve this without introducing another generic.  I opened #2306 as an possible alternaive

huozhi:
> I like the idea. But i think we could achieve this without introducing another generic. I opened #2306 as an possible alternaive

but it introduces more typing reloading for the hook, so we gonna manage more reloadings. For the new genertic type this PR only introduced one, but #2306 have more new interface with genertics need to manage?

_internal/types.ts
```diff
@@ -21,10 +21,18 @@ export type Fetcher<
   ? (arg: Arg) => FetcherResponse<Data>
   : SWRKey extends null | undefined | false
   ? never
-  : SWRKey extends infer Arg
+  : SWRKey extends string | infer Arg
```
promer94, line 24: why add string here ?

promer94 (approved):

_internal/types.ts
huozhi: good catch, accidentally

BoneTM:
why the `useSWRInfinite` dont impl this stuff?

JordanHoffman:
Sorry I'm still having this problem, I think it's maybe because of generic types or something?

I specified my fetcher function like this:
image

And then in my code it still shows as possibly undefined when doing the hooks:
image

Is it perhaps because doing useSWRImmutable<BuildPageConfig[]> took priority for the returned data type? Is there another way to fix this? i get confused pretty quickly with types.

PR #2306, opened 2022-12-20
types: response type

promer94:


huozhi:
Is this same with #2301 ?

PR #2338, opened 2022-12-26
fix: delay revalidation if a key is already active and has error

### LINKED ISSUE #2336: Loading state and UI lead to Infinite retry
# Bug report

When one of the API is wrong, if the loading status is judged and loading UI is displayed, then swr will retry the wrong API infinitely without interruption. Interestingly, when the loading status is not displayed, it will retry at the expected interval for a certain period of time

## Expected Behavior

no Infinite loop

## Repro Steps / Code Example

[CodeSandbox](https://codesandbox.io/s/sad-wescoff-rh9ye5?file=/src/App.js)

> code line: **37** is the key point 

## Additional Context

SWR version： 2.0.0

Maybe relation to https://github.com/vercel/swr/issues/66

promer94:
close #2336

The reason for #2336 is that 

every time a revalidation is finished, a new swr hook will mount and trigger a new revalidation. 

`isValidating` will also cause similar problem here. 

https://codesandbox.io/s/isvalidating-infinite-render-xgwyyz

### DIFF
--- a/core/use-swr.ts
+++ b/core/use-swr.ts
@@ -183,6 +183,10 @@ export const useSWRHandler = <Data = any, Error = any>(
   )
 
   const isInitialMount = !initialMountedRef.current
+
+  const hasRevalidator =
+    EVENT_REVALIDATORS[key] && EVENT_REVALIDATORS[key].length > 0
+
   const cachedData = cached.data
 
   const data = isUndefined(cachedData) ? fallback : cachedData
@@ -201,6 +205,9 @@ export const useSWRHandler = <Data = any, Error = any>(
   // - Not suspense mode and there is no fallback data and `revalidateIfStale` is enabled.
   // - `revalidateIfStale` is enabled but `data` is not defined.
   const shouldDoInitialRevalidation = (() => {
+    // if a key already has revalidators and also has error, we should not trigger revalidation
+    if (hasRevalidator && !isUndefined(error)) return false
+
     // If `revalidateOnMount` is set, we take the value directly.
     if (isInitialMount && !isUndefined(revalidateOnMount))
       return revalidateOnMount
@@ -511,8 +518,6 @@ export const useSWRHandler = <Data = any, Error = any>(
       return
     }
 
-    const isKeyHasActiveInstance =
-      EVENT_REVALIDATORS[key] && EVENT_REVALIDATORS[key].length > 0
     const unsubEvents = subscribeCallback(key, EVENT_REVALIDATORS, onRevalidate)
 
     // Mark the component as mounted and update corresponding refs.
@@ -522,9 +527,6 @@ export const useSWRHandler = <Data = any, Error = any>(
 
     // Keep the original key in the cache.
     setCache({ _k: fnArg })
-
-    // if a key is already active and has error, we should not trigger revalidation
-    if (isKeyHasActiveInstance && !isUndefined(error)) return
 
     // Trigger a revalidation.
     if (shouldDoInitialRevalidation) {
--- a/core/use-swr.ts
+++ b/core/use-swr.ts
@@ -511,6 +511,8 @@ export const useSWRHandler = <Data = any, Error = any>(
       return
     }
 
+    const isKeyHasActiveInstance =
+      EVENT_REVALIDATORS[key] && EVENT_REVALIDATORS[key].length > 0
     const unsubEvents = subscribeCallback(key, EVENT_REVALIDATORS, onRevalidate)
 
     // Mark the component as mounted and update corresponding refs.
@@ -520,6 +522,9 @@ export const useSWRHandler = <Data = any, Error = any>(
 
     // Keep the original key in the cache.
     setCache({ _k: fnArg })
+
+    // if a key is already active and has error, we should not trigger revalidation
+    if (isKeyHasActiveInstance && !isUndefined(error)) return
 
     // Trigger a revalidation.
     if (shouldDoInitialRevalidation) {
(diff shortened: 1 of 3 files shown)


core/use-swr.ts
```diff
@@ -511,6 +511,8 @@ export const useSWRHandler = <Data = any, Error = any>(
       return
     }
 
+    const isKeyHasActiveInstance =
```
koba04, line 514: IMO, the name of `isKeyHasActiveInstance` might be a bit unclear to me. Is `hasRevalidator` clearer?

core/use-swr.ts
```diff
@@ -521,6 +523,9 @@ export const useSWRHandler = <Data = any, Error = any>(
     // Keep the original key in the cache.
     setCache({ _k: fnArg })
 
+    // if a key is already active and has error, we should not trigger revalidation
+    if (isKeyHasActiveInstance && !isUndefined(error)) return
```
koba04, line 527: This condition gets `unsubEvents` (L516) not to be called when unmounting. It would have to be put in the following condition .

```js
if (shouldDoInitialRevalidation)
```

https://github.com/vercel/swr/pull/2338/files#diff-134c45c7d0aef1e568828f8f2ee0f22398adc692b99cea901978c568ca0836caR530

promer94: Thank you !

core/use-swr.ts
```diff
@@ -183,6 +183,10 @@ export const useSWRHandler = <Data = any, Error = any>(
   )
 
   const isInitialMount = !initialMountedRef.current
+
+  const hasRevalidator =
+    EVENT_REVALIDATORS[key] && EVENT_REVALIDATORS[key].length > 0
```
koba04, line 188: @promer94 Thank you for applying code review!

I reread the current implementation and couldn't understand why we need to check if there is another revalidator only when an error happened.

Isn't it enough to add a check the error is `undefined` like this?

```jsx
    // If there is no stale data and error, we need to revalidate when mount;
    // If `revalidateIfStale` is set to true, we will always revalidate.
    return isUndefined(error) && (isUndefined(data) || config.revalidateIfStale)
```

https://github.com/vercel/swr/blob/3cec4bd18a163bde31d8382915fcf0753d7194ed/core/use-swr.ts#L216-L218

promer94: For most of time, we actully want to start revalidation after hooks mouted  when there is an error in cache.

But for edge case like #2336, we dont want second-mounted hook to start revalidation since the first-mouted hooks will handle it.


We just want a signle instantace to handle the revalidation process.  


https://github.com/vercel/swr/blob/06a644129e66982f2d837047be797df2e506b228/_internal/utils/cache.ts#L17-L24
https://github.com/vercel/swr/blob/3cec4bd18a163bde31d8382915fcf0753d7194ed/_internal/utils/mutate.ts#L106-L110

koba04: @promer94 
Thank you! I understand the intention 👍 

Let me share another solution I came up with.
I feel the issue is caused by an incorrect `isLoading` state on an error, and I'd say `isLoading` should be `true` only when both `data` and `error` are `undefined`.

https://github.com/vercel/swr/blob/3cec4bd18a163bde31d8382915fcf0753d7194ed/core/use-swr.ts#L302-L304

should be

```jsx
      if (isUndefined(getCache().data) && isUndefined(getCache().error)) {
        initialState.isLoading = true
      }
```


What do you think?

promer94: Good Catch !

But `isValidating` will also cause infinite retry here.  So I think we stiil need to check the number of hook to avoid it.

we could apply both changes and a extra test for ` isValidating`.

koba04: > But isValidating will also cause infinite retry here. So I think we stiil need to check the number of hook to avoid it.

I missed the case, thank you!

promer94:
> @promer94 Thank you! I understand the intention 👍
> 
> Let me share another solution I came up with. I feel the issue is caused by an incorrect `isLoading` state on an error, and I'd say `isLoading` should be `true` only when both `data` and `error` are `undefined`.
> 
> https://github.com/vercel/swr/blob/3cec4bd18a163bde31d8382915fcf0753d7194ed/core/use-swr.ts#L302-L304
> 
> should be
> 
> ```js
>       if (isUndefined(getCache().data) && isUndefined(getCache().error)) {
>         initialState.isLoading = true
>       }
> ```
> 
> What do you think?

After reviewing #1928, I think it would be batter to discuss this change in a spreate pr.

core/use-swr.ts
promer94:
```jsx
 return isUndefined(error) && (isUndefined(data) || config.revalidateIfStale) 
```

This implmentation will fix the problem but it also will change the previous behavior. (start revalidation right after hooks mouted when there is an error in cache). It will not start revalidation immediately anymore. It has to wait for a focus event or auto revalidation.

> Should we also check the existing revalidators for non-error cases not only error cases to avoid revalidation by second-mounted hook? 

We actully did that for other kind of revalidators. 

https://github.com/vercel/swr/blob/06a644129e66982f2d837047be797df2e506b228/_internal/utils/cache.ts#L17-L24
https://github.com/vercel/swr/blob/3cec4bd18a163bde31d8382915fcf0753d7194ed/_internal/utils/mutate.ts#L106-L110


And for mounting revalidation, we also did same thing but in different way to prevent second mounted hook to start a new revalidation

https://github.com/vercel/swr/blob/c57edae96a0af03497f657027e2e784d2915965b/core/use-swr.ts#L258

PR #2341, opened 2022-12-27
fix: types properites in exports fields

koba04:
fixes #2308

This fixes the path for the properties of the type in `package.json`.

Each `types` of `package.json` are pointed correctly.

- https://github.com/vercel/swr/blob/main/immutable/package.json#L4
- https://github.com/vercel/swr/blob/main/infinite/package.json#L4
- https://github.com/vercel/swr/blob/main/mutation/package.json#L4

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,19 +23,19 @@
       "require": "./core/dist/index.js"
     },
     "./infinite": {
-      "types": "./infinite/dist/index.d.ts",
+      "types": "./infinite/dist/infinite/index.d.ts",
       "import": "./infinite/dist/index.mjs",
       "module": "./infinite/dist/index.esm.js",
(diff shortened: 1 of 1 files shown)


shuding (approved):

shuding:
Thanks!

huozhi:
We can also remove the extra path in each folder name.

We had `immutable/dist/immutable/index.d.ts` because previously bunchee has some limitations for types generation, and now we can keep it simple `types: ./dist/index.d.ts` in each package

talzion12:
Can the fixed version be published in npm? I don't think it's possible to use swr in typescript with esm without this fix.

huozhi:
@talzion12 the change is released in 2.0.1 🙏

PR #2350, opened 2023-01-06
feat: add all mutate options support to useSWRInfinite

### LINKED ISSUE #1899: useSWRInfinite Mutate function doesn't accept the same arguments as useSWR Mutate

koba04:
fixes #1899

I have to add more tests for the options.

### DIFF
--- a/infinite/index.ts
+++ b/infinite/index.ts
@@ -222,7 +222,7 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
         }
 
         return arguments.length
-          ? swr.mutate(data, shouldRevalidate)
+          ? swr.mutate(data, { ...options, revalidate: shouldRevalidate })
           : swr.mutate()
       },
       // swr.mutate is always the same reference
--- a/test/use-swr-infinite.test.tsx
+++ b/test/use-swr-infinite.test.tsx
@@ -1328,6 +1328,333 @@ describe('useSWRInfinite', () => {
     await screen.findByText('data: foo-1')
   })
 
+  it('should be able to use the functional optimisticData option', async () => {
+    let t = 0
+    const key = createKey()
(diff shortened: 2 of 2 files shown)


shuding (approved):
Thank you!

shuding (approved):
Thank you!

PR #2353, opened 2023-01-08
fix: parameters passed to subscribe is inconsistent with its declaration

hong24:
Just as the title describes, the following is a place that has a such problem:

aab1363a377dbf7d7cb527d78a39400


promer94 (approved):

promer94:
Good catch ! Thank you !

PR #2354, opened 2023-01-08
fix: data passed to refreshInterval function is not latest

hong24:
fix: #2547


koba04:
@hong24 Thank you! Could you add a test for this?

koba04 (commented):
I'd say this is a stale data problem, so I think this should be fixed by adding `data` into the dependency list of `useIsomorphicLayoutEffect`

hong24:
There is an issue when data is a dependency: refreshInterval will be called twice
one time when data is changed,
the other time when revalidating successfully
2fba51545417963b7bfa0a383903092

hong24:
I have other questions：
1. The first call to refreshInterval is with undefined. In my understanding, It's supposed to call refreshinterval after the first fetch is resolved and with the resolved data.
2. When the key is nutl, the fetcher wil be not called, but ref reshInterval will be called over and over againAccording to the related source code, this problem may be expected but It's a little weird and unintuitive.
I'm currently just fixing the latest data, but for the above two issues, I don't have a good solution.

koba04:
> There is an issue when data is a dependency: refreshInterval will be called twice

Is this a problem? I think `refreshInterval` would be a pure function that returns an interval time, so it would be ok to be called twice. It seems to make sense to reset the timer when `data` has been changed.

hong24:
I think this is a problem. Even a pure function called twice doesn't make sense, and the parameters of the `refreshInterval` call are different twice. Once with the latest data and once with the previous data.

suutari:
I got hit by this issue today.

Additionally seems that there's another problem too: If the function passed to `refreshInterval` is staying unchanged between the `useSWR` calls, it seems to get called only once and that call is with the undefined data; but it won't get called  anymore after the data arrives. As a workaround I can define the refreshInterval as a lambda so that it'll change between successive useSWR calls.

Here's an example from my current code, which tries to keep data of a job up-to-date with the refreshInterval:
```typescript
export function useJobInfo(jobId: JobId | null)  {
  return useSWR(
    jobId ? `/jobs/${jobId}` : null,
    fetcher,
    {
      dedupingInterval: 250,

      // This doesn't work, since function would be called only once:
      //refreshInterval: jobId ? getRefreshIntervalOfJob : 0,

      // This is the workaround:
      refreshInterval: jobId ? (jobInfo) => getRefreshIntervalOfJob(jobInfo) : 0,
    }
  );
}

function getRefreshIntervalOfJob(jobInfo: JobInfo | undefined): number {
  ...
}
```

suutari:
> Additionally seems that there's another problem too: If the function passed to `refreshInterval` is staying unchanged between the `useSWR` calls, it seems to get called only once and that call is with the undefined data; but it won't get called anymore after the data arrives.

Actually, now that I checked again, have to correct my statement: It will be called again, if I return non-zero value from the function when it's called for the first time (with the undefined value). However, all successive calls are still done with the undefined value, even if the data is already fetched.

Seems like that the `next` function has the `data` stored in the closure and it won't update when refreshInterval function is called with the timer. The fix in this PR seems to help in my case too.

Any plans to release this fix?

core/use-swr.ts
```diff
@@ -588,10 +588,11 @@ export const useSWRHandler = <Data = any, Error = any>(
     let timer: any
 
     function next() {
+      const cacheData = getCache().data
       // Use the passed interval
       // ...or invoke the function with the updated data to get the interval
       const interval = isFunction(refreshInterval)
-        ? refreshInterval(data)
+        ? refreshInterval(isUndefined(cacheData) ? data : cacheData)
```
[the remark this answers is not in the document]
promer94: I think we should just cached data here to keep it simple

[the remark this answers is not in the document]
hong24: Yes, it is better to use fallback data.

core/use-swr.ts
```diff
@@ -588,10 +588,15 @@ export const useSWRHandler = <Data = any, Error = any>(
...
+      const cacheData = getCache().data
+      const configFallbackData = getConfig().fallbackData
+      const latestFallback = isUndefined(configFallbackData)
+        ? getConfig().fallback[key]
+        : configFallbackData
       // Use the passed interval
       // ...or invoke the function with the updated data to get the interval
       const interval = isFunction(refreshInterval)
-        ? refreshInterval(data)
+        ? refreshInterval(isUndefined(cacheData) ? latestFallback : cacheData)
```
promer94, line 599: The original purpose of this feature is to change the refreshInterval based on **updated fetched data**
ref: #1686

It would be better if we only use cached data here to keep it simple.

promer94 (changes requested):

core/use-swr.ts
promer94, line 599:
```suggestion
        ? refreshInterval(getCache().data)
```

koba04: @promer94 

> The original purpose of this feature is to change the refreshInterval based on updated fetched data
> ref: #1686

You might put the wrong issue number, and #1686 is not relevant to this.

> It would be better if we only use cached data here to keep it simple.

Yeah, I agree with you, and it would be good if we could use cache data only.
But the current implementation uses fallback data as the `data`, so it's safer to keep the behavior to avoid a breaking change. It makes sense for me to pass fallback data when cache data is `undefined`. We could have the change in the next major release.

We might have the change as a bug fix, but I'm not sure how the change affects users.

promer94: Thanks. i just updated the wrong issue number.

Since this feature actually never works as expected after it got released. I feel it is safe to change the implementation in a minor version here. User can still access `fallbackData` from clouser. I guess It won't be a big problem here ?

promer94 (approved):

koba04 (approved):

PR #2386, opened 2023-01-26
fix: when inifinite key changes, should use cached pagesize instead of initialSize

### LINKED ISSUE #2370: useSWRInfinite with Next.js router query parameters vs pages
# Bug report

I've noticed that `useSWRInfinite` is behaving weird in Next.js when using the router query parameters to change the URI key. The issue is only visible after going back or forward in the browser.

## Repro Steps / Code Example

I have made a copy of the [official infinite example](https://github.com/vercel/swr/tree/main/examples/infinite) here:

- https://swr-infinite-issue.vercel.app/
- https://github.com/horgen/swr-infinite-issue

_Note: I'm console logging the `size` to make it easier to see the change in the following cases._

#### Case 1 (working as expected): 

1. Go to https://swr-infinite-issue.vercel.app
2. Click the "load more" button two times.
3. Click the "**pages/page.js**" link at the top. This will change the URI key by going to a new page (...and remounting all the things).
4. Go back to the previous page by click the back button in your browser

**Result:** `useSWRInfinite` will remember the correct `size: 3` of the first page.

#### Case 2 (not working as expected):

1. Go to https://swr-infinite-issue.vercel.app
2. Click the "load more" button two times.
3. Click the "**pages/index.js - 2**" link at the top. This will change the URI key by updating a router query parameter.
4. Go back to the previous page by click the back button in your browser

**Result:** `useSWRInfinite` will remember the correct `size: 3` of the first page for a second before resetting it to 1.

promer94:
Repro: https://codesandbox.io/s/2370-forked-nwkd2v

close #2370

### DIFF
--- a/infinite/index.ts
+++ b/infinite/index.ts
@@ -115,7 +115,8 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
 
       if (infiniteKey) {
         // If the key has been changed, we keep the current page size if persistSize is enabled
-        set({ _l: persistSize ? lastPageSizeRef.current : initialSize })
+        // Otherwise, we reset the page size to cached pageSize
+        set({ _l: persistSize ? lastPageSizeRef.current : resolvePageSize() })
       }
 
       // `initialSize` isn't allowed to change during the lifecycle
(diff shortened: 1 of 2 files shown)


shuding (approved):
Since page size is per-hook, this fix makes sense!

PR #2387, opened 2023-01-27
fix: make compare function only compare data (#2375)

### LINKED ISSUE #2375: Wrong data passed to custom compare function in version 2.0.x
# Bug report

When I configure a custom compare function, the parameters that are passed to my function at runtime are booleans instead of the expected custom type
## Expected Behavior

The expected behaviour is that my custom compare functions receives the expected type returned by the fetcher.
## Repro Steps / Code Example
```
compare(a, b) {
        if (typeof a === 'undefined' || typeof b === 'undefined') {
          return false;
        } else if (typeof a === 'object' && typeof b === 'object') {
          const mappedA = a.results.map((r) => r.id);
          const mappedB = a.results.map((r) => r.id);
          return isEqual(mappedA, mappedB);
        } else {
          return a == b;
        }
      },
```

## Additional Context

SWR version: 2.0.1
Framework: Next.js 12.3.1

Retrospection:
close #2375

### DIFF
--- a/core/use-swr.ts
+++ b/core/use-swr.ts
@@ -111,12 +111,18 @@ export const useSWRHandler = <Data = any, Error = any>(
     let equal = true
     for (const _ in stateDependencies) {
       const t = _ as keyof StateDependencies
-      if (!compare(current[t], prev[t])) {
-        if (t === 'data' && isUndefined(prev[t])) {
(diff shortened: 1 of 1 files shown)


promer94 (approved):
Thank You !

huozhi:
Can we add a test for this change?

PR #2388, opened 2023-01-27
ci: only run release when tags pushed

promer94:
Currently the release action is always running after test action.  
https://github.com/vercel/swr/actions/workflows/release.yml

If i understand correctly,  the release action intend to do release only after the test action is succeed and a new tag is push.

But current configuration cant achieve that.  The `on`  keyword is more like `OR` instead of `AND` . So now the release action will always run after the test action.

So i merge the test and release job. Then we could make sure they will run sequentially.
a special context `github.ref`  is used for distinguishing tags from other events.  

It would be batter if release action only run after the tags was created.


.github/workflows/release.yml
```diff
@@ -11,6 +11,7 @@ on:
 
 jobs:
   release:
+    if: ${{ contains('refs/tags', github.ref) == 'true' }}
```
promer94, line 14: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

.github/workflows/release.yml
```diff
@@ -19,9 +20,7 @@ jobs:
...
 
-      - name: Publish
-        run: |
-          echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
-          npm publish --access public
+      - name: Release
+        run: pnpm publish --access public --no-git-checks
         env:
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN_ELEVATED }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN_ELEVATED }}
```
promer94, line 26: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry

.github/workflows/test.yml
```diff
@@ -22,5 +23,3 @@ jobs:
           pnpm clean
           pnpm build
           pnpm test-packaging
-        env:
```
promer94, line 25: CI is always true

https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

.github/workflows/test.yml
```diff
@@ -22,5 +24,18 @@ jobs:
...
+        uses: actions/checkout@v3
+
+      - name: Install
+        uses: ./.github/workflows/install
+
+      - name: Release
+        run: pnpm publish --access public --no-git-checks
         env:
-          CI: true
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN_ELEVATED }}
```
promer94, line 41: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry

.github/workflows/install/action.yml
```diff
@@ -3,10 +3,12 @@ description: 'Set up and install dependencies'
 
 inputs:
   fetch-depth:
+    description: 'Number of commits to fetch'
     default: '1'
     required: false
   persist-credentials:
-    default: true
+    description: 'Whether to persist credentials'
+    default: "true"
```
huozhi, line 11: Why this needs to be string?

.github/workflows/test.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    tags:
```
huozhi, line 7: Shall we rename this file then? It’s not test anymore.

promer94: yes, but i don't come up with a good name. Do you have any idea ?

.github/workflows/install/action.yml
promer94: screen-shot

.github/workflows/test.yml
shuding: "Release testing", or just "test and release" lol.

huozhi (approved):

PR #2391, opened 2023-01-27
fix: Ensure hydration matches the SSR result during streaming

shuding:
## Problem

Here's a diagram explaining the problem:

image

First, In SSR, the `data` will always be `undefined` for any `useSWR` hook, because `useEffect` never runs during SSR. Same for streaming SSR because it's still SSR, but with some components being delayed a bit.

So, to make sure client hydration matches the SSR result, all `useSWR` hooks should return `undefined` during their hydration render. That's easy when we hydrate all things together, because the cache is always empty initially, and the hydration process will finish before any data fetching happens.

However, with streaming SSR React can partially hydrate a part of the app. In the example above, the first `useSWR` hook was hydrated first, and the second is still pending due to slow network or a suspense boundary. After the hydration of the first `useSWR` hook, it fetches the data and updates the global cache on the client. And then, the second one hydrates. During its hydration render, it already reads the updated cache, instead of `undefined`. This causes a hydration mismatch.

The fix is to pass a server snapshot to `useSyncExternalStore`, so React will use that snapshot during hydration renders. And all we need to do is to make sure that we store the initial cache value somewhere before any write happens on the client, and return that initial cache as the server snapshot.

## Reproduction

It's very tricky to reproduce and test this problem in our Jest setup. So I made this Sandbox to test: https://codesandbox.io/p/sandbox/swr-ssr-forked-tlgw5b?file=%2Fapp%2Fpage.js

Hydration mismatch in [current `swr@latest`](https://tlgw5b-3000.preview.csb.app/):

CleanShot 2023-01-27 at 23 24 35@2x

Works fine [with this fix](https://nf5eeu-3000.preview.csb.app/):

CleanShot 2023-01-27 at 23 25 16@2x

### DIFF
--- a/test/use-swr-streaming-ssr.test.tsx
+++ b/test/use-swr-streaming-ssr.test.tsx
@@ -1,0 +1,81 @@
+import React, { Suspense } from 'react'
+import useSWR from 'swr'
+import {
+  createKey,
+  createResponse,
+  renderWithConfig,
+  hydrateWithConfig,
+  mockConsoleForHydrationErrors,
+  sleep
+} from './utils'
+
+describe('useSWR - streaming', () => {
+  afterEach(() => {
+    jest.clearAllMocks()
+    jest.restoreAllMocks()
+  })
+
+  it('should match ssr result when hydrating', async () => {
+    const ensureAndUnmock = mockConsoleForHydrationErrors()
+
+    const key = createKey()
+
+    // A block fetches the data and updates the cache.
+    function Block() {
+      const { data } = useSWR(key, () => createResponse('SWR', { delay: 10 }))
+      return <div>{data || 'undefined'}</div>
+    }
+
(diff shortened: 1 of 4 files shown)


promer94 (approved):
The diagram and explanation are awesome !

test/use-swr-streaming-ssr.test.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { Suspense } from 'react'
```
promer94, line 1: I would like to add e2e test using playwright for this 

since i already finished the most of jobs in #2252 

I think we should nextjs app-dir here ?

shuding: Sounds good! ❤️

huozhi (approved):

PR #2395, opened 2023-01-28
test: stream ssr e2e

promer94:
## Motivation
This pr intends to setup playwright for swr. So we can properly test scenario like #2391 and add more e2e test easily in the future. 

## Benefit
1. Real browser api could be used

swr uses a lot of navigator specific APIs. It would be batter if we could use real browser enviorment for testing. It's more reliable then jsdom.

2. It is easier to write test

we dont need to worry about details like  [act](https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning) or [StrictMode](https://github.com/reactwg/react-18/discussions/17) in  e2e test.  playwright also has test generator and vscode extension to help us write tests.

3. It has detailed trace info for debugging or performance monitoring

playwright could provider detailed [trace info](https://playwright.dev/docs/trace-viewer) for debugging or performance monitoring. It would help a lot to prevent flaky tests.  It could also collect some [low-level trace info](https://playwright.dev/docs/api/class-browser#browser-start-tracing) for us to evaluate swr's performance.


e2e/site/app/partially-hydrate/block.tsx
```diff
@@ -0,0 +1,26 @@
+'use client'
```
shuding, line 1: It seems that this file is unused?

shuding (approved):
Looks amazing, thank you! 😍

PR #2404, opened 2023-02-02
feat: add the parallel option in useSWRInfinite

koba04:
This PR adds the `parallel` option into `useSWRInfinite` to make revalidations in parallel.

```jsx
useSWRInfinite(getKey, fetcher, { parallel: true })
```

The default value is `false`.

## Motivation

When I revalidate multiple pages with `useSWRInfinite`, `useSWRInfinte` sends the revalidation requests sequentially, which is slow when we have many pages to revalidate. This especially becomes a problem when revalidations happen frequently.

```
// without parallel
page1 ====> page2 ====> page3 ====> done

// parallel
page1
page2 ====> done 
page3
```

## Limitation

The `parallel` option doesn't work with cursor-based APIs because cursor-based APIs send the next request based on the cursor value returned from the current request.

https://swr.vercel.app/docs/pagination#example-2-cursor-or-offset-based-paginated-api

With the `parallel` option, `useSWRInfinite` can't get the `previousPageData` because requests happen in parallel. The current implementation passes `null` as `previousPageData` with the `parallel` option. 
TODO: infer the type of `previousPageData` based on the `parallel` option.

## Alternative

I've also considered the `concurrency = 1` option to be able to handle revalidations more flexibly. But I had to handle `previousPageData` differently based on whether revalidations are sequential or parallel. I think the `parallel` option is not more flexible than the `concurrency` option, but it's clearer than the `concurrency` option.

### DIFF
--- a/infinite/index.ts
+++ b/infinite/index.ts
@@ -57,7 +57,8 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
       revalidateAll = false,
       persistSize = false,
       revalidateFirstPage = true,
-      revalidateOnMount = false
+      revalidateOnMount = false,
+      parallel = false
     } = config
 
     // The serialized key of the first page. This key will be used to store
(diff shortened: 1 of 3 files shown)


shuding (commented):
I like this idea, so if `parallel = true` this works like a list for dynamic number of requests right? It seems that this covers the use case for "useSWRList" too

koba04:
Yeah, that sounds great👍  This still has a concept of `page`, so some cases might not fit this though.

shuding (approved):

PR #2411, opened 2023-02-08
fix: ensure initCache setter function stays within bounds of subscriptions

### LINKED ISSUE #2357: Mutate thrown error "subs[i] not a function."
# Bug report

Hi everyone,

I upgraded SWR from 1.2.2 to 2.0.0. After upgrading, some cases are broken. Mutate function thrown error 'subs[i] not a function.' If I refresh the app completely, there is no error thrown. 

Example case:

1) I have BasketData fetched SWR.
2) I send a request to another service to remove a basket item and then mutate the basketData loading state manually.
3) When the request is done I mutate basketData with the response. In this step, the error is thrown.

If I do the same flow after f5 (reload app), there is no error.

## Expected Behavior

I rollback to SWR 1.2.2 and there is no issue there. I expect the same behavior with 2.0.0.

## Repro Steps / Code Example

## Additional Context

SWR version 2.0.0

Screen Shot 2023-01-11 at 09 31 53

lfbergee:
This PR closes #2357, by creating a copy of the `subscriptions[key]` array in the `setter` function for `initCache`. By doing this we avoid an intermittent bug where `subscriptions[key]` is mutated while the for-loop is running, potentially causing an out of bounce lookup.

By copying the object, mutations of the original object will not affect the `setter` function while executing. Alternatively we could lock write operation while running the read operation. Or use a different iterator method, that make it's own copy instead of a vanilla for-loop.

This is somewhat hard bug to reproduce, as it's highly intermittent.

### DIFF
--- a/_internal/utils/cache.ts
+++ b/_internal/utils/cache.ts
@@ -63,12 +63,10 @@ export const initCache = <Data = any>(
     }
     const setter = (key: string, value: any, prev: any) => {
       provider.set(key, value)
-      if (subscriptions[key]) {
-        // Make a copy of subscriptions to avoid mutations of subscriptions to affect
-        // the array while looping
-        const subs = subscriptions[key].slice()
-        for (let i = subs.length; i--; ) {
-          subs[i](value, prev)
+      const subs = subscriptions[key]
+      if (subs) {
+        for (const fn of subs) {
+          fn(value, prev)
         }
       }
     }
--- a/_internal/utils/cache.ts
+++ b/_internal/utils/cache.ts
@@ -63,8 +63,10 @@ export const initCache = <Data = any>(
     }
     const setter = (key: string, value: any, prev: any) => {
       provider.set(key, value)
-      const subs = subscriptions[key]
-      if (subs) {
+      if (subscriptions[key]) {
+        // Make a copy of subscriptions to avoid mutations of subscriptions to affect
+        // the array while looping
+        const subs = subscriptions[key].slice()
         for (let i = subs.length; i--; ) {
           subs[i](value, prev)
         }


huozhi:
Thanks for the PR! Can you add a test for this?

lfbergee:
> Thanks for the PR! Can you add a test for this?

It's quite challenging making a test that reproduce the original problem, as it looks to me as a race condition when two or more caches are initiated at the same time with the same key for the global store, but different number of subscriptions. I can add a unit test for the `initCache`, but haven't found a way reproduce the error in isolation. I've tested the fix in my application code where I do have the error intermittently, also shipped a patched version to PROD, and this do eliminate the error, no more reporting in Sentry either.

_internal/utils/cache.ts
```diff
@@ -63,8 +63,10 @@ export const initCache = <Data = any>(
     }
     const setter = (key: string, value: any, prev: any) => {
       provider.set(key, value)
-      const subs = subscriptions[key]
-      if (subs) {
+      if (subscriptions[key]) {
+        // Make a copy of subscriptions to avoid mutations of subscriptions to affect
+        // the array while looping
+        const subs = subscriptions[key].slice()
```
promer94, line 69:
```suggestion
const subs = subscriptions[key]
      if (subs) {
        for (const fn of subs) {
          fn(value, prev)
        }
      }
```
Maybe we could use `for of` here, so it don't need a extra copy of subs here ?

promer94 (commented):
This problem may invovle some implementaion details of react which makes it hard to create a test case.

When sub[i] is called, React will be told to update. If the update happened immediately and cause some swr hooks umounted, then sub will be changed and it is not safe to call sub[i] anymore

However when i try to reproduce this problem, i noticed that the update won't happen right way. It seems that all the updates will be batched.   

I think we could merge this now and add test case later.

_internal/utils/cache.ts
lfbergee: Yeah, that should probably achieve the same. I can update on Monday as soon as I have access to a computer! Thanks for feedback and attention!

promer94 (approved):

PR #2431, opened 2023-02-17
refactor: remove useless dataRef, always compare cached data

promer94:
Previously, we store all page data with `inifinite-${firstPageKey}` and individual page with `${pagekey}` into cache, and sync all page data into dataRef.  

During the revalidation or mutation, we will compare dataRef with individual page to update the result.
However, we could just compare cached data directly then we don't need to store an extra copy of page data

i also update the signature of `compareFn` for `useSWRInfinite`

### DIFF
--- a/infinite/index.ts
+++ b/infinite/index.ts
@@ -126,9 +126,9 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
     const shouldRevalidateOnMount = revalidateOnMount && !didMountRef.current
 
     // Actual SWR hook to load all pages in one fetcher.
-    const swr = useSWRNext<Data[], Error>(
+    const swr = useSWRNext(
       infiniteKey,
-      async _infiniteKey => {
+      async key => {
         // get the revalidate context
         const forceRevalidateAll = get()._i
 
@@ -139,7 +139,7 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
         const [getCache] = createCacheHelper<
           Data,
           SWRInfiniteCacheValue<Data[], any>
-        >(cache, _infiniteKey)
+        >(cache, key)
         const cacheData = getCache().data
 
         const revalidators = []
--- a/infinite/index.ts
+++ b/infinite/index.ts
@@ -254,9 +254,9 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
           previousPageData = pageData
         }
         return data
-        // exclude getKey from the dependencies, which isn't allowed to change during the lifecycle
-        // eslint-disable-next-line react-hooks/exhaustive-deps
       },
+      // exclude getKey from the dependencies, which isn't allowed to change during the lifecycle
+      // eslint-disable-next-line react-hooks/exhaustive-deps
       [cache]
     )
     // Extend the SWR API
--- a/infinite/index.ts
+++ b/infinite/index.ts
@@ -288,7 +288,7 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
           // Get the cached page data.
           const pageData = pageKey ? getCache().data : UNDEFINED
 
-          // Return the current data if we can't get it from the cache.
+          // Call `mutate` with infinte cache data if we can't get it from the page cache.
           if (isUndefined(pageData)) {
             return mutate(getInfiniteCache().data)
           }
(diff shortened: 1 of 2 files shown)


infinite/index.ts
```diff
@@ -130,14 +128,19 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
     // Actual SWR hook to load all pages in one fetcher.
     const swr = useSWRNext<Data[], Error>(
       infiniteKey,
-      async () => {
+      async _infiniteKey => {
```
koba04, line 131: Q: Can we use `infiniteKey` instead of `_infiniteKey`?

infinite/index.ts
```diff
@@ -171,11 +174,11 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
...
             isUndefined(pageData) ||
-            (revalidateFirstPage && !i && !isUndefined(dataRef.current)) ||
+            (revalidateFirstPage && !i && !isUndefined(cacheData)) ||
             shouldRevalidateOnMount ||
-            (originalData &&
-              !isUndefined(originalData[i]) &&
-              !config.compare(originalData[i], pageData))
+            (cacheData &&
+              !isUndefined(cacheData[i]) &&
+              !config.compare(cacheData[i], pageData))
```
koba04, line 181: Q: Previously, `originalData` was set only when the revalidation is caused by `mutate` that has `data` and enable the `revalidate` option. Does this change any behavior?

infinite/index.ts
```diff
@@ -256,35 +252,46 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
...
-        const pageData = pageKey ? cache.get(pageKey)?.data : UNDEFINED
-
-        // Return the current data if we can't get it from the cache.
-        if (isUndefined(pageData)) return dataRef.current
-
-        data.push(pageData)
-        previousPageData = pageData
-      }
+    const resolvePagesFromCache = useCallback(
+      (pageSize: number, key: string): Data[] | undefined => {
```
koba04, line 256: Q: If we put `infiniteKey` into the deps array of `useCallback`, does the `key` argument get unnecessary?

infinite/types.ts
```diff
@@ -23,17 +23,22 @@ export type SWRInfiniteKeyLoader<
   Args extends Arguments = Arguments
 > = (index: number, previousPageData: Data | null) => Args
 
+export interface SWRInfiniteCompareFn<Data = any> {
+  (a: Data | undefined, b: Data | undefined): boolean
+  (a: Data[] | undefined, b: Data[] | undefined): boolean
+}
```
koba04, line 29: The type signature is correct, but this might be a breaking change if users put a type signature into their `compare` function.

```ts
useSWRInfinite<Data>(
  getKey,
  fetcher,
  {
    compare: (a: Data, b: Data) {
      // ...
    }
  }
);
```

koba04 (commented):
Great! This helps to understand the implementation 👍 
I've commented some nits questions and a breaking change.

infinite/index.ts
promer94: The type of `infiniteKey` is `string | null`. 

The type and runtime value of `_infiniteKey` here would always be `string`.

We don't need to do extra check later if we use `_infiniteKey` here.

infinite/index.ts
promer94: It would not change the behavior here. 

```set({ _i: [true] })``` will set `forceRevalidateAll` to `true`, then the expression result will be `true`. The value of `originalData(cacheData)` would be ignored here

File: infinite/index.ts (hunk 3 above)
Comment (promer94) [reply infinite/index.ts:256]: I realize we could just inline this function into `setSize`

File: infinite/index.ts (hunk 1 above)
Comment (huozhi) [reply infinite/index.ts:131]: can we rename it? `_xx` feels like a arg that usually being skipped in ts. like `stringifiedKey`

File: infinite/index.ts [hunk 5]
```diff
@@ -293,13 +269,38 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
...
+        for (let i = 0; i < size; ++i) {
+          const [pageKey] = serialize(getKey(i, previousPageData))
+          const [getCache] = createCacheHelper<
+            Data,
+            SWRInfiniteCacheValue<Data, any>
+          >(cache, pageKey)
+          // Get the cached page data.
+          const pageData = pageKey ? getCache().data : UNDEFINED
+
+          // Return the current data if we can't get it from the cache.
```
koba04, line 291:
```suggestion
          // Call `mutate` with infinte cache data if we can't get it from the page cache.
```

koba04 (approved):
👍 LGTM!

Issue #2436, opened 2023-02-19
uswSWRImmutable always load `fallbackData` in suspense mode


Himself65:
See test case https://github.com/vercel/swr/pull/2437

koba04:
@Himself65 
SWR treats `fallbackData` as stale data and `useSWRImmutable` doesn't revalidate if there is stale data (`revalidateIfStale: false`).
In this case, you can use `revalidateOnMount: true` to fetch data on mount, which can be combined with `preload`.

Himself65:
Thanks for the suggestion

PR #2437, opened 2023-02-19
test: add test case for useSWRImmutable

Himself65:
A **expected** test case for https://github.com/vercel/swr/issues/2436

## Expected Behavior

I think `this is 1` should appear on the screen

## Actual Behavior

it's always `this is 0`

### DIFF
--- a/test/use-swr-immutable.test.tsx
+++ b/test/use-swr-immutable.test.tsx
@@ -1,6 +1,6 @@
 import { screen, act, fireEvent } from '@testing-library/react'
-import React, { useState } from 'react'
-import useSWR from 'swr'
+import React, { Suspense, useState } from 'react'
+import useSWR, { preload } from 'swr'
 import useSWRImmutable from 'swr/immutable'
 import {
   sleep,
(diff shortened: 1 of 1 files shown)


Himself65:
```
Summary of all failing tests
FAIL test/use-swr-immutable.test.tsx
  ● useSWR - immutable › should do something

    Unable to find an element with the text: this is 1. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.

    Ignored nodes: comments, <script />, <style />
    <body>
      <div>
        <div>
          this is 0
        </div>
      </div>
    </body>

      128 |     })
      129 |     screen.getByText('this is 0')
    > 130 |     await screen.findByText('this is 1')
          |                  ^
      131 |   })
      132 |
      133 |   it('should not revalidate with the immutable hook', async () => {

      at waitForWrapper (node_modules/.pnpm/@testing-library+dom@8.16.0/node_modules/@testing-library/dom/dist/wait-for.js:187:27)
      at node_modules/.pnpm/@testing-library+dom@8.16.0/node_modules/@testing-library/dom/dist/query-helpers.js:101:33
      at Object.findByText (test/use-swr-immutable.test.tsx:130:18)

```

test/use-swr-immutable.test.tsx
```diff
@@ -100,6 +100,36 @@ describe('useSWR - immutable', () => {
...
+  it('should do something', async () => {
+    async function fetcher(key: string) {
+      expect(key).toBe('constant')
+      return new Promise<string>(resolve => {
+        setTimeout(() => resolve('this is 1'), 1000)
+      })
+    }
+    preload('constant', fetcher)
+    function Component() {
+      const { data } = useSWRImmutable('constant', {
```
Himself65, line 112: If you change `useSWRImmutable` to `useSWR.` The test will pass, but semantically not what I want

PR #2453, opened 2023-02-24
ci: fix publish workflow

promer94:
Previously, we did not setup registry-url, so `setup-node` would not craete a `.npmrc` file.

https://github.com/actions/setup-node/blob/main/src/main.ts#L58-L60
https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages


huozhi (approved):
👍

Issue #2253, opened 2022-11-28
Next.js 13: `SWRConfig` should mark components as `"use client"`


itta611:
# Bug report
SWRConfig is used for top class components (e.g. layout), but the component `SWRConfig` is not defined as the Client Components.
Therefore, developer have to write `"use client"`, which makes all components in the layout as Client Components.

## Description / Observed Behavior
```
Server Error
TypeError: react.createContext is not a function
```

## Expected Behavior
Therefore, the developer must write `"use client"`, which makes all components in the layout client components.

## Additional Context
swr: 1.3.0
next: 13.0.2

SukkaW:
The current workaround would be manually wrapping `<SWRConfig />` in your client components:

```tsx
// my-provider.client.tsx
'use client';
export const MyProvider = ({ children }: React.PropWithChildren<unknown>) => {
  return <SWRConfig>{children}</SWRConfig>
};

// layout.tsx
export default function Layout({ children }: React.PropWithChildren<unknown>) {
  return <MyProvider>{children}</MyProvider>
}
```

And IMHO `<SWRConfig />` should also include `import 'client-only'`.

PR #2464, opened 2023-02-28
fix: add use client to swr config provider

ariesclark:
The issues #2253 and https://github.com/vercel/swr/discussions/2292 both mention this problem,
But none actually fix the issue, this change is minimal and *shouldn't?* break anything.


shuding:
Thanks! CC @huozhi to make this work we'll have to change the way we bundle the core export.

huozhi:
Addressing the `"use client"` directives with bundler updates in  #2696

PR #2475, opened 2023-03-04
fix: avoid creating new snapshot if cache is not updated at client during streaming

### LINKED ISSUE #2473: SWR 2.0.3 unnecessary re-render when useSyncExternalStore snapshots are the same value
# Bug report

There is an extra component render caused by SWR hook call. 
In the example below no SWR stateful values are consumed, so we should not see any re-rendering, but still see that one re-render: https://swr.vercel.app/docs/advanced/performance#dependency-collection

This affects performance and overall core web vitals score.

Downgrading to `2.0.2` fixed the issue and can see only one initial render. 

## Expected Behavior

We shouldn't re-render component if the store snapshot & server snapshot return the same value/object.

## React version

React: `18.2.0`

## SWR version

SWR: `2.0.3`

## Repro Steps / Code Example

https://github.com/domazet93/user-swr-2.0.3-app
https://user-swr-2-0-3-app-with-rerendering.vercel.app/

```
import useSWR from "swr";
import axios from "axios";

const fetch = axios.create({
  baseURL: "https://jsonplaceholder.typicode.com"
});

const useFetchUser = () =>
  useSWR("/users/100", (url) =>
    fetch.get(url).then((res) => {
      const { data: userData } = res;

      if (userData) {
        return userData;
      }

      return undefined; 
    })
  );


export default function UserSWR() {
  useFetchUser();

  return <div>Open React Profiler and check each time SWRTest rendered</div>;
}
```

### SWR `v2.0.2` (no re-rendering)
Screenshot 2023-03-03 at 13 48 18

### SWR `v2.0.3` (with re-rendering)
Screenshot 2023-03-03 at 13 37 06

Screenshot 2023-03-03 at 13 37 41

## Additional Context
Tested with `"react": "18.2.0"`
Introduced in `v2.0.3`: https://github.com/vercel/swr/compare/v2.0.2...v2.0.3

PR: https://github.com/vercel/swr/pull/2391 

This is the code in question:
```
// To make sure that we are returning the same object reference to avoid
// unnecessary re-renders, we keep the previous snapshot and use deep
// comparison to check if we need to return a new one.
let memorizedSnapshot = getSelectedCache(getCache())
const memorizedInitialSnapshot = getSelectedCache(getInitialCache())

return [
  () => {
    const newSnapshot = getSelectedCache(getCache())
    return isEqual(newSnapshot, memorizedSnapshot)
      ? memorizedSnapshot
      : (memorizedSnapshot = newSnapshot)
  },
  () => memorizedInitialSnapshot
]
// eslint-disable-next-line react-hooks/exhaustive-deps
}, [cache, key])
```

This snapshots (objects) get passed to `useSyncExternalStore` and from the docs we should probably serialize it?
https://beta.reactjs.org/reference/react/useSyncExternalStore#parameters

> The server snapshot must be the same between the client and the server, and is usually serialized and passed from the server to the client.

promer94:
In the past, SWR always created new snapshots for both the server and client. 

However, i found that in cases where the cache is not updated on the client during streaming, it's safe to reuse the existing client snapshot to avoid unnecessary rerendering. 

close #2473

### DIFF
--- a/core/use-swr.ts
+++ b/core/use-swr.ts
@@ -160,12 +160,17 @@ export const useSWRHandler = <Data = any, Error = any>(
         ...snapshot
       }
     }
-
+    const cachedData = getCache()
+    const initialData = getInitialCache()
+    const clientSnapshot = getSelectedCache(cachedData)
+    const serverSnapshot =
+      cachedData === initialData
+        ? clientSnapshot
+        : getSelectedCache(initialData)
     // To make sure that we are returning the same object reference to avoid
     // unnecessary re-renders, we keep the previous snapshot and use deep
     // comparison to check if we need to return a new one.
-    let memorizedSnapshot = getSelectedCache(getCache())
-    const memorizedInitialSnapshot = getSelectedCache(getInitialCache())
+    let memorizedSnapshot = clientSnapshot
 
     return [
       () => {
@@ -174,7 +179,7 @@ export const useSWRHandler = <Data = any, Error = any>(
           ? memorizedSnapshot
           : (memorizedSnapshot = newSnapshot)
       },
-      () => memorizedInitialSnapshot
+      () => serverSnapshot
     ]
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [cache, key])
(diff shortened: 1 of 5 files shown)


shuding (approved):
Thanks!

PR #2480, opened 2023-03-05
chore: Add subscription example

huozhi:
This PR adds a basic example of using `useSWRSubscription` with an external data source.

### DIFF
--- a/examples/subscription/pages/index.js
+++ b/examples/subscription/pages/index.js
@@ -1,52 +1,54 @@
-import React, { useCallback, useEffect, useState } from "react";
-import useSWRSubscription from "swr/subscription";
-import EventEmitter from "events";
+import React, { useCallback, useEffect, useState } from "react"
+import useSWRSubscription from "swr/subscription"
+import EventEmitter from "events"
 
-const event = new EventEmitter();
+const event = new EventEmitter()
 
-const swrKey = "sub-key";
+const swrKey = "sub-key"
 
 export default function page() {
-  const [isOn, setIsOn] = useState(true);
-  const [num, setNum] = useState(0);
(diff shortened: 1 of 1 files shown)


examples/subscription/pages/index.js
```diff
@@ -0,0 +1,52 @@
+import React, { useCallback, useEffect, useState } from "react";
+import useSWRSubscription from "swr/subscription";
+import EventEmitter from "events";
```
shuding, line 3: Is this available in the browser?

huozhi: we had polyfill in nextjs

examples/subscription/pages/index.js
```diff
@@ -0,0 +1,52 @@
...
+    [isOn]
+  );
+
+  useEffect(() => {
+    const timer = setInterval(() => {
+      setNum(num + 1);
+    }, 1000);
+
+    return () => clearInterval(timer);
+  }, [num]);
```
shuding, line 41:
```suggestion
const event = new EventEmitter();

// Simulating an external data source.
let stopped = false
async function start () {
  for (let i = 0; i < 100; i++) {
    await new Promise(res => setTimeout(res, 1000))
    if (stopped) return
    if (i % 3 === 1) {
      event.emit("error", new Error("error: " + i));
    } else {
      event.emit("data", "state: " + i);
    }
  }
}

export default function page() {
  const { data, error } = useSWRSubscription('my-sub', (key, { next }) => {
    event.on("data", (value) => next(undefined, value));
    event.on("error", (err) => next(err));
    start();

    return () => {
      stopped = true;
    };
  })
```

What do you think? Mostly I think we should move the data source out as that feels more like the common case.

Issue #2505, opened 2023-03-14
Type of `data` should be a union type with `undefined` when a config of type `SWRConfiguration` is passed to `useSWR`


connorch:
# Bug report

## Description / Observed Behavior

When passing an SWR config of type `SWRConfiguration | undefined` as an argument to `useSWR`, `data` in the swr response is no longer a union type with `undefined` (regardless of whether the swr config defined `suspense: true`).

## Expected Behavior

I expect `data` in the below example to be a union type `User | undefined` because the hook is not running in suspense mode.

## Repro Steps / Code Example

Here is a suggested test showcasing the currently failing case:

```ts
export function testUndefinedData() {
  const fetcher = (k: string) => Promise.resolve({ value: k })
  const { data } = useSWR('/api', fetcher, {} as SWRConfiguration)
  expectType<Equal<typeof data, { value: string } | undefined>>(true)
}
```

## Additional Context

This issue was newly introduced in `2.1.0`.

I have confirmed in a local test environment that it was introduced in this PR: https://github.com/vercel/swr/pull/2452/files

nstepien:
When doing `{} as SWRConfiguration`, you're asserting that the value has a different type, I wouldn't expect it to give you the correct type for `{}` if you assert it to be something else.
Have you tried using `{} satisfies SWRConfiguration` instead?

connorch:
Yeah this is a good point.  Not a big deal here since we do want to trick typescript into thinking that the `{}` is of type `SWRConfiguration`, but `satisfies` is the safer option 👍.

This PR is merged already though so I'm unfortunately that fix would need to come in a separate PR - please feel free to submit one.

connorch:
Fixed in https://github.com/vercel/swr/pull/2506

Issue #2396, opened 2023-01-29
Improved type support for suspense is not working well


mehm8128:
# Bug report

## Description / Observed Behavior

In v2.0.1, https://github.com/vercel/swr/pull/2301 improved type support for `suspense`, so when we make `suspense` true, `data` can't be `undefined`. But when we use generics for `useSWR`, this type support become not working.
The same is true of `fallbackData`.

I think it is because TypeScript hasn't support partial type inferring of generics.

![image](https://user-images.githubusercontent.com/83744975/215336028-2c5be87e-2d2a-493d-a7e5-9f7400444bfa.png)
![image](https://user-images.githubusercontent.com/83744975/215336050-c19d5e44-bbb6-4e1f-9664-48cedb827241.png)

## Expected Behavior

Even if we use generics, `data` must not be `undefined`.

And in this way, it work as we expect.
![image](https://user-images.githubusercontent.com/83744975/215336074-5496b497-cdf0-4170-a7f4-5e4214502081.png)

## Repro Steps / Code Example

I attached images above.

## Additional Context

SWR version: v2.0.3

PR #2452, opened 2023-02-24
Fix: type support for suspense and fallbackData(#2396)

taro-28:
close https://github.com/vercel/swr/issues/2396

PR #2506, opened 2023-03-14
fix: swr infers incorrect `data` type for default `SWRConfig` generic type

connorch:
This PR resolves issue https://github.com/vercel/swr/issues/2505 by reverting PR https://github.com/vercel/swr/pull/2452, and updates the tests to better explain why this issue cannot be fixed: https://github.com/vercel/swr/issues/2396

https://github.com/vercel/swr/issues/2396 suggests that if a generic is passed to the `useSWR` hook, that the library should still infer the type of the SWRConfig passed in.  However, this isn't possible in typescript due to limitations with partial inference.  That is, typescript can either infer all generics or none at all - never some but not others.

Here's a simple example to showcase this partial inference limitation:

```ts

type Foo = <A = any, B = any, C = any>(a: A, b: B, c: C) => [A, B, C]
const foo: Foo = (a, b, c) => [a, b, c]
const bar1 = foo(1, 2, 3)
//     ^? const bar1: [number, number, number]
const bar2 = foo<number>(1, 2, 3)
//     ^? const bar2: [number, any, any]  - Passing just one explicit generic to `foo` removes type inference for the rest of the generics.
```

https://github.com/vercel/swr/pull/2452 didn't actually fix the issue https://github.com/vercel/swr/issues/2396 - it just forced it so that the default of the `SWRConfig` generic forced the type check in `BlockingData` to always evaluated to `true`.  This is now tested against in the new test case `testConfigAsSWRConfiguration`.  I also updated all of the tests added in that PR to point to some documentation explaining this typescript limitation.

Some other documentation on this issue:
https://github.com/microsoft/TypeScript/issues/26242
https://stackoverflow.com/questions/60377365/typescript-infer-type-of-generic-after-optional-first-generic

### DIFF
--- a/_internal/types.ts
+++ b/_internal/types.ts
@@ -232,23 +232,27 @@ export interface SWRHook {
     SWRKey extends Key = StrictKey,
     SWROptions extends
       | SWRConfiguration<Data, Error, Fetcher<Data, SWRKey>>
-      | undefined = SWRConfiguration<Data, Error, Fetcher<Data, SWRKey>>
+      | undefined =
+      | SWRConfiguration<Data, Error, Fetcher<Data, SWRKey>>
+      | undefined
   >(
     key: SWRKey,
-    config: SWROptions | undefined
-  ): SWRResponse<Data, Error, Required<SWROptions>>
+    config: SWROptions
+  ): SWRResponse<Data, Error, SWROptions>
   <
     Data = any,
     Error = any,
     SWRKey extends Key = StrictKey,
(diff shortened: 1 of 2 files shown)


connorch:
This PR fixes https://github.com/vercel/swr/issues/2505

promer94 (approved):

promer94:
Thanks for the comments and links !

connorch:
Thanks for the speedy merge! 🙌

PR #2517, opened 2023-03-19
test: fix syntax error in Equal type alias implementation

SACHINnANYAKKARA:
This TypeScript code defining a type alias Equal<A, B>. The type alias compares two types A and B and returns a boolean value indicating whether they are equal or not.

The implementation uses a conditional type with two nested conditional type expressions. However, there is a syntax error in the implementation. The find issue text is not a valid TypeScript syntax, and it causes a syntax error.


huozhi:
Thanks for the PR, can you add a typing test for that?

SACHINnANYAKKARA:
@huozhi Sure I'll add typing test

huozhi (approved):
thanks

PR #2525, opened 2023-03-23
fix: Improve useSubscription typing

oosawy:
Fixes useSubscription typing that is not accept generics and do not inter key type.

For the generics problem, I believe that is caused by the TypeScript behind the difference between `type` and `interface`, anyway using `interface` instead of `type` solves the ploblem.

Also, I checked other hooks (`useSWR`, `useSWRInfinite`, `useSWRMutation`, `useSWRImmutable`) and they used `interface` for typing and works fine.

reproduction:
https://stackblitz.com/edit/react-ts-ur5fbw?file=App.tsx

### DIFF
--- a/subscription/index.ts
+++ b/subscription/index.ts
@@ -22,11 +22,13 @@ export type SWRSubscriptionResponse<Data = any, Error = any> = {
   error?: Error
 }
 
-export type SWRSubscriptionHook<Data = any, Error = any> = (
-  key: Key,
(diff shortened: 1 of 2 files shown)


promer94:
Thanks for your contribution! 

but i need to address this issue and #2522 with a more complex implementation. Therefore, I have opened another pull request in #2535 

So i will close this pr now, let me know if you have any further questions.

PR #2536, opened 2023-03-31
chore: remove engines

### LINKED ISSUE #2531: This package is breaking builds trying to use pnpm 8.0
Because  the `:/package.json` specifies

```json
 "engines": {
    "pnpm": "7"
 }
```

I cannot build my project with 8.0 and `swr` is the only package that does this, so I wonder if it is necessary

Here's the error I get when I run `pnpm i`
```
Lockfile is up to date, resolution step is skipped
 ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)

Your pnpm version is incompatible with "/swr/2.0.3(react@18.2.0)".

Expected version: 7
Got: 8.0.0

This is happening because the package's manifest has an engines.pnpm field specified.
To fix this issue, install the required pnpm version globally.
```

promer94:
close #2531

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,9 +142,6 @@
     "arrowParens": "avoid",
     "trailingComma": "none"
   },
-  "engines": {
-    "pnpm": "7"
-  },
   "dependencies": {
     "use-sync-external-store": "^1.2.0"
   }


promer94:
https://github.com/vercel/swr/blob/8f8501ce8bfad4e8a079ad4dbe00c0b287987900/package.json#L13

we do have `packageManager": "pnpm@7.x.x` in our package.json. So i think it is safe to delete the `engines` field ?

PR #2537, opened 2023-04-02
Rename subscription types

huozhi:
Rename `SWRSubNext` added in #2535 to `SWRSubscriptionOptions`, as next is part of the options. We designed it as an object for easily extending with more arguments in the future if needed


promer94 (approved):
Thanks !

PR #2550, opened 2023-04-07
Fix #2548: pass origin key to subcription callback

### LINKED ISSUE #2548: useSWRSubscription key doesnt deconstruct array
# Bug report

When using an array key the key that is passed to the subscribe function is a string and not an array

## Expected Behavior
I expect it to be deconstructed similar to is shown under "Usage" here https://swr.vercel.app/docs/subscription

## Repro Steps / Code Example

```
  const key = "a key";
  const { data } = useSWRSubscription(["key", key], (passedKey, { next }) => {
    passedKey === '@"key","a key",' // returns true
    return () => {};
  });
```

https://codesandbox.io/s/relaxed-silence-h556kz?file=/src/App.js

## Additional Context
2.1.2

Zheaoli:
Fix #2548


promer94 (approved):
LGTM！ Thanks ！

Issue #2216, opened 2022-10-28
Since v1.0.0 `cache.set` does not serialize keys.


ansavchenco:
# Bug report

## Description / Observed Behavior

Directly setting cache with `cache.set` used to serialize keys automatically but starting from `v1.0.0` it stops doing so. 

## Expected Behavior

I'm not sure if that's expected behaviour but at least this change is not mentioned neither on [the releases page](https://github.com/vercel/swr/releases/tag/1.0.0) nor in [this blog post](https://swr.vercel.app/blog/swr-v1).

## Repro Steps / Code Example

Here is the `v0.5.7` code and the it's cache ([codesandbox link](https://codesandbox.io/s/swr-old-cache-i52uz5?file=/src/App.tsx)).
```typescript
import { useEffect } from "react";
import useSWR, { cache } from "swr";

export function App() {
  useEffect(() => {
    cache.set(["/api/user/:id", "123"], { id: "123", name: "John" });
  }, []);

  useSWR(["/api/user/:id", "234"], () =>
    Promise.resolve({ id: "234", name: "Jack" })
  );

  console.log(cache);

  return <h1>Hello</h1>;
}
```
```
[
    {
        "key": "validating@arg@\"/api/user/:id\"@\"234\"",
        "value": false
    },
    {
        "key": "arg@\"/api/user/:id\"@\"123\"",
        "value": {"id": "123","name": "John"}
    },
    {
        "key": "err@arg@\"/api/user/:id\"@\"234\""
    },
    {
        "key": "arg@\"/api/user/:id\"@\"234\"",
        "value": {"id": "234","name": "Jack"}
    }
]
```

And here is the `v1.0.0` code and the it's cache ([codesandbox link](https://codesandbox.io/s/swr-new-cache-775vy7?file=/src/App.tsx)). The same for `v1.3.0`.
```typescript
import { useEffect } from "react";
import useSWR, { useSWRConfig } from "swr";

export function App() {
  const { cache } = useSWRConfig();

  useEffect(() => {
    cache.set(["/api/user/:id", "123"], { id: "123", name: "John" });
  }, [cache]);

  useSWR(["/api/user/:id", "234"], () =>
    Promise.resolve({ id: "234", name: "Jack" })
  );

  console.log(cache);

  return <h1>Hello</h1>;
}
```
```
[
    {
        "key": "$req$arg$\"/api/user/:id\"$\"234\"",
        "value": false
    },
    {
        "key": ["/api/user/:id","123"],
[...]

PR #2554, opened 2023-04-07
Update Cache Interface types

dmmulroy:
# Update Cache Interface types

## Update 04/08/23
Per discussion with @promer94, the solution outlined in [An Alternative Approach](#alternative-approach) has been implemented rather than updating the `keys` interface.

## Motivation & Background

As it's currently typed, the `Cache` interface that is required for users to create and provide their own cache implementations mandates that the `keys` method return an `IterableIterator<string>`. I believe this to be incorrectly typed, or at the very least, inconveniently typed and a burden for consumers. I think that the correct type should be `IterableIterator<Key>` given that the `get`, `set`, and `delete` methods all require a `Key` as input. 

Consider the [default implementation](https://github.com/vercel/swr/blob/main/_internal/utils/config.ts#L44) provided by `swr`:

```typescript
// Default cache provider
const [cache, mutate] = initCache(new Map()) as [Cache<any>, ScopedMutator<any>]
```
This is effectively the same as:

```typescript
const cacheMap: Cache<any> = new Map<Key, State<any, any>>()
const [cache, mutate] = initCache(cacheMap) as [Cache<any>, ScopedMutator<any>]
```
Which results in a type error:

```
Type 'Map<Key, State<any, any>>' is not assignable to type 'Cache<any>'.
     The types returned by 'keys()' are incompatible between these types.
       Type 'IterableIterator<Key>' is not assignable to type 'IterableIterator<string>'.
         Type 'Key' is not assignable to type 'string'.
           Type 'undefined' is not assignable to type 'string'.
```

This puts a burden on users/consumers of `swr` when implementing their own cache, especially if the underlying data structure is a `Map`. 

My use case is creating a custom LRU Cache to back `swr`, see the following gist for its implementation: https://gist.github.com/dmmulroy/33c31bef81554b099726b3a680dad974

Because `keys` is typed as `IterableIterator<string>` I have to cast.

```typescript
/**
   * Returns an iterable iterator over the keys in the cache.
   *
   * @returns {IterableIterator<string>} - An iterator of the cache keys.
   */
  public keys(): IterableIterator<string> {
    return this.cache.keys() as IterableIterator<string>
  }
```

## An Alternative Approach <a id="alternative-approach"></a>
An alternative approach would be to update the `Cache` interface to the following:

```typescript
export interface Cache<Data = any> {
  keys(): IterableIterator<string>
  get(key: string): State<Data> | undefined
  set(key: string, value: State<Data>): void
  delete(key: string): void
}
```

This actually may be more accurate and reflect what occurs at run time. As far as I can tell, any time the cache is accessed by `swr` internally the `key` is always serialized into a `string` via the `serialize`[^1] function prior to access. 

You can see that any time the cache is accessed[^2], it's accessed via callbacks returned from `createCacheHelper`[^3] in which the `SWRGlobalState` is referenced/retrieved and used get to the `setter` callback from `initCache`[^4] which is stored in global state. In all instances (you can search by reference to confirm), the `Getter` and `Setter` callbacks are passed `string` keys which have been serialized.

In addition, I believe this may actually be worth considering because `unstable_serialize`[^5] is publicly exported. I assume this is to aid consumers/users if they choose to directly access/mutate the cache (despite explicitly being called out in the docs as anti pattern[^6]) as seen in #2216. 

[^1]: https://github.com/vercel/swr/blob/main/_internal/utils/serialize.ts#L6
[^2]: [Getter Callback](https://github.com/vercel/swr/blob/main/_internal/utils/helper.ts#L39), [Setter Callback](https://github.com/vercel/swr/blob/main/_internal/utils/helper.ts#L41), [Previous value in Setter](https://github.com/vercel/swr/blob/main/_internal/utils/helper.ts#L43), [Setter Callback from SWRGlobalState](https://github.com/vercel/swr/blob/main/_internal/utils/cache.ts#L64)
[^3]: https://github.com/vercel/swr/blob/main/_internal/utils/helper.ts#L32
[^4]: https://github.com/vercel/swr/blob/main/_internal/utils/cache.ts#L26
[^5]: https://github.com/vercel/swr/blob/main/core/use-swr.ts#L679
[^6]: https://swr.vercel.app/docs/advanced/cache#modify-the-cache-data

### DIFF
--- a/_internal/types.ts
+++ b/_internal/types.ts
@@ -428,7 +428,7 @@ export type RevalidateCallback = <K extends RevalidateEvent>(
 ) => RevalidateCallbackReturnType[K]
 
 export interface Cache<Data = any> {
-  keys(): IterableIterator<string>
+  keys(): IterableIterator<Key>
   get(key: Key): State<Data> | undefined
   set(key: Key, value: State<Data>): void
   delete(key: Key): void
--- a/_internal/types.ts
+++ b/_internal/types.ts
@@ -428,10 +428,10 @@ export type RevalidateCallback = <K extends RevalidateEvent>(
 ) => RevalidateCallbackReturnType[K]
 
 export interface Cache<Data = any> {
-  keys(): IterableIterator<Key>
-  get(key: Key): State<Data> | undefined
-  set(key: Key, value: State<Data>): void
-  delete(key: Key): void
+  keys(): IterableIterator<string>
+  get(key: string): State<Data> | undefined
+  set(key: string, value: State<Data>): void
+  delete(key: string): void
 }
 
 export interface StateDependencies {
--- a/_internal/utils/helper.ts
+++ b/_internal/utils/helper.ts
@@ -36,7 +36,7 @@ export const createCacheHelper = <Data = any, T = State<Data, any>>(
   const state = SWRGlobalState.get(cache) as GlobalState
   return [
     // Getter
-    () => (cache.get(key) || EMPTY_CACHE) as T,
+    () => ((!isUndefined(key) && cache.get(key)) || EMPTY_CACHE) as T,
     // Setter
     (info: T) => {
       if (!isUndefined(key)) {
@@ -61,7 +61,7 @@ export const createCacheHelper = <Data = any, T = State<Data, any>>(
       }
 
       // If we haven't done any client-side updates, we return the current value.
-      return (cache.get(key) || EMPTY_CACHE) as T
+      return ((!isUndefined(key) && cache.get(key)) || EMPTY_CACHE) as T
     }
   ] as const
 }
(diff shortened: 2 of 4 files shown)


dmmulroy:
I also just noticed that [the docs are incorrect](https://swr.vercel.app/docs/advanced/cache#cache-provider) as well and list the type of `Cache` as the following:

```typescript
interface Cache<Data> {
  get(key: string): Data | undefined
  set(key: string, value: Data): void
  delete(key: string): void
  keys(): IterableIterator<string>
}
```

promer94:
Thanks for the PR !

I think we should take the alternative approach here. 
It is more simple and the fact is that we only use string key here.

dmmulroy:
Awesome - I'll be happy to make those changes and get them pushed up tomorrow or sunday.

dmmulroy:
Changes have been made, take a look when you have the chance @promer94 🚀

promer94 (approved):
Thanks

PR #2546, opened 2023-04-06
feat: add name export for `useSWR`

Innei:
Add name export for `useSWR`, to improve auto import this hook by tsserver.

Issue #2379, opened 2023-01-25
Add named export of useSWR to the library or fix default export naming types


IgnusG:
# Bug report

## Description / Observed Behavior

It's really difficult to use `useSWR` when typescript's auto import is used since the library does not actually specify useSWR as the name for the default export. VSCode's TypeScript integration then for example shows just this:

![image](https://user-images.githubusercontent.com/6438760/214676581-56aed0cb-0507-40a9-b680-e030cfc003a8.png)

## Expected Behavior

Changing the index.d.ts of swr to this:

![image](https://user-images.githubusercontent.com/6438760/214676841-7249aa43-d82b-43f7-bb2b-54fcd7e69a82.png)

Makes the import suggestions much more useful:

![image](https://user-images.githubusercontent.com/6438760/214676910-4987da74-eca6-49b6-b073-8629503e6ead.png)

Alternatively it would be great to provide `useSWR` as a named export along with the default export for anyone who prefers this over default exports (god knows there's a whole internet dev culture preaching against using them).

## Repro Steps / Code Example

1. Using typescript type `useSWR` and trigger the autocomplete suggestions of eg. VSCode
2. Watch the results not offer the actual `useSWR` hook as an import suggestion

PR #2563, opened 2023-04-14
types: Allow auto-import by improving generated types

oosawy:
NOTE: This is NOT to add or change API or exports and all changes is about code styling that should not effect actual behavior

The changes let tsc emit types with the hook function name not `_default` so we can auto-import by that function name (even default exported).

#### How types change
Now:

```.d.ts
declare const _default: SWRHook;

export { SWRConfig, _default as default, unstable_serialize };
```

With the changes:

```.d.ts
declare const useSWR: SWRHook;

export { SWRConfig, useSWR as default, unstable_serialize };
```

Also, this is done in [`useSWRSubscription`](https://github.com/vercel/swr/blob/9371f1d0632fe7d9c7ce65608cf1c1119cca01fd/subscription/index.ts#L124-L129) and [its types](https://unpkg.com/browse/swr@2.1.3/subscription/dist/index.d.ts) is exporting `useSWRSubscription as default`. So we can already auto-import `useSWRSubscription` today.

#### Related issues and prs
- #1889
- #2546
- #2379

### DIFF
--- a/core/use-swr.ts
+++ b/core/use-swr.ts
@@ -693,4 +693,6 @@ export const unstable_serialize = (key: Key) => serialize(key)[0]
  * }
  * ```
  */
-export default withArgs<SWRHook>(useSWRHandler)
+const useSWR = withArgs<SWRHook>(useSWRHandler)
+
+export default useSWR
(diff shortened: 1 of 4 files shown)


huozhi (approved):
Thanks! Great improvement!

PR #2564, opened 2023-04-17
fix: pass serialized args to preload fetcher

oosawy:
Fixed `preload()` to accept the kay function similar to the `useSWR` hook, also improves the types.

The 'preload' function currently passes the fetcher function the key that was used to call it. For example, calling `preload(() => 'key', args => console.log(args))` wil show `() => 'key'`.

These changes may impact someone who have encountered this bug and worked around it like `args()` in the fetcher function. However, with the improved types, it could be detected as an error in TypeScript.


promer94 (approved):
Thanks !

PR #2575, opened 2023-04-21
fix: add more protection against undefined document

tvthatsme:
Related to [this discussion](https://github.com/vercel/swr/discussions/2493), somewhere in the lifecycle of jest environments getting set up and torn down, the document object isn't always available. This change doesn't _seem_ necessary, but solves the issue that I am experiencing randomly in CI/CD pipelines but consistently when running a large testing suite with --runInBand.

I've tried setting the document object globally in jest
```
global.beforeEach(() => {
  Object.defineProperty(document, 'visibilityState', {
    configurable: false,
    writable: false,
    value: 'visible'
  })
})
```
and similar things but that doens't fix the issue.

Happy to provide more details as I can or additional code changes to see this change land in `swr`. Thanks!


giovannivocale-molar:
I am running into this issues a lot in my tests. Could this PR be reviewed?  🙏

chrislicodes:
> I am running into this issues a lot in my tests. Could this PR be reviewed? 🙏

Same for me! A review would be amazing 🥹

huozhi:
Thanks for the PR

From the conversation also mentioned: 
> the problem is that the test runtime has started tearing down its jsdom environment, so accesses to document throw errors

This method is available in browser, it's more a jest setup problem. Making that optional shouldn't be necessary

promer94:
I think we could have a workaround for this problem 

For testing, you could provide your own `inActive` implementation 
https://swr.vercel.app/docs/advanced/react-native#customize-focus-and-reconnect-events

```tsx
<SWRConfig
  value={{
    isVisible() {
      const visibilityState = isDocumentDefined && document?.visibilityState
  return isUndefined(visibilityState) || visibilityState !== 'hidden'
    },
  }}
>
  <App />
</SWRConfig>
```

Also add optional chain for this statement will add extra bytes and runtime overhead. 
Screenshot 2023-06-19 at 20 27 17
Screenshot 2023-06-19 at 20 27 23

PR #2576, opened 2023-04-22
fix: Fix dependency tracking and useSES bug

### LINKED ISSUE #2446: Some fetch requests can get stuck when multiple requests are made by a single component
# Bug report

When a component makes multiple fetch requests, and once one request gets fulfilled, there is a chance that the remaining requests get stuck (`data` remains as `undefined` despite response being received as seen from network tab).

## Expected Behavior
I would expect the remaining requests to continue getting updated. 

## Repro Steps / Code Example
[Repo to reproduce bug](https://github.com/justhonnoldwegoinghome/swr-stale-error-repro.git)

### ❌ Doesn't work
- `index.tsx` reproduces the error (gets stuck on `Loading...` because `programmingJoke.data` and/or `pun.data` stays as `undefined`) but not all the time. Might require a few browser refreshes for bug to surface. 
```typescript
// index.tsx

import useSWR from "swr";

import { getJoke } from "@/api/getJoke";
import { getProgrammingJoke } from "@/api/getProgrammingJoke";
import { getPun } from "@/api/getPun";

  const joke = useSWR("joke", getJoke);
  const programmingJoke = useSWR("programming-joke", getProgrammingJoke);
  const pun = useSWR("pun", getPun);

  if (!joke.data || !programmingJoke.data || !pun.data)
    return <div>Loading...</div>;

  return (
    <div>
      <p>{joke.data.joke}</p>
      <br />
      <p>{programmingJoke.data.joke}</p>
      <br />
      <p>{pun.data.joke}</p>
    </div>
  );
}
```
### ✅ Works
- `works.tsx` shows how to resolve error. But doesn't explain observed behaviour. 
```typescript
// works.tsx

import useSWR from "swr";

import { getJoke } from "@/api/getJoke";
import { getProgrammingJoke } from "@/api/getProgrammingJoke";
import { getPun } from "@/api/getPun";

  const { data: joke } = useSWR("joke", getJoke);
  const { data: programmingJoke } = useSWR(
    "programming-joke",
    getProgrammingJoke
  );
  const { data: pun } = useSWR("pun", getPun);

  if (!joke || !programmingJoke || !pun) return <div>Loading...</div>;

  return (
    <div>
      <p>{joke.joke}</p>
      <br />
      <p>{programmingJoke.joke}</p>
      <br />
      <p>{pun.joke}</p>
    </div>
  );
}
```

## Additional Context

SWR version: 2.0.3

shuding:
Fixes #2446. CC @promer94.

The overall idea of `useSES` is that state changes are triggered from the external. It works in a pipeline of "external state change → compare it with local state's snapshot → trigger re-render and update local state". The main catch here is that SWR's dependency tracking system (`stateDependencies.data = true`) isn't considered as a state in that loop. But then, even if we put `stateDependencies` part of the state and compare/update it through `useSES`, we can't do a synchronous read of the snapshot which is required as we use a getter:

```js
get data() {
  // Trigger a re-render synchronously with the updated snapshot
  setCache({ stateDependencies_data: true })

  return returnData
},
```

Mentally this should work (similar to [this example](https://legacy.reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops)) but it doesn't seem to with `useSES`.

So in this PR, an extra ref `latestComparedState` is added to track the latest snapshot that `useSES` received inside, and return that state if possible. This shouldn't cause any problem in concurrent rendering because the process should happen in the same component instance and tick as `useSES`'s returned state, but as a complete state.

This way we are still managing the state via `useSES` but only re-renders it when necessary.

Here's a diagram explaining the idea:

image

### codesandbox ci
https://codesandbox.io/s/2446-grkt00?file=/src/App.js
https://codesandbox.io/s/swr-re-render-dn2whh?file=/src/App.tsx

### DIFF
--- a/core/use-swr.ts
+++ b/core/use-swr.ts
@@ -21,7 +21,6 @@ import {
   internalMutate,
   revalidateEvents,
   mergeObjects
-  // stableHash
 } from 'swr/_internal'
 import type {
   State,
@@ -103,16 +102,11 @@ export const useSWRHandler = <Data = any, Error = any>(
     >(cache, key)
 
   const stateDependencies = useRef<StateDependencies>({}).current
-  const latestComparedState = useRef<State<Data, any>>({})
-
   const fallback = isUndefined(fallbackData)
     ? config.fallback[key]
     : fallbackData
 
-  const compareSelectedSnapshots = (
-    prev: State<Data, any>,
-    current: State<Data, any>
-  ) => {
+  const isEqual = (prev: State<Data, any>, current: State<Data, any>) => {
     for (const _ in stateDependencies) {
       const t = _ as keyof StateDependencies
       if (t === 'data') {
@@ -120,7 +114,7 @@ export const useSWRHandler = <Data = any, Error = any>(
           if (!isUndefined(prev[t])) {
             return false
           }
-          if (!compare(cachedDataWithLaggyAndFallback, current[t])) {
+          if (!compare(returnedData, current[t])) {
             return false
           }
         }
@@ -130,13 +124,6 @@ export const useSWRHandler = <Data = any, Error = any>(
         }
       }
     }
-
-    // Since we only compare the selected fields, even if the states are
-    // considered as equal they might still differ. When the tracked
-    // dependencies change, we need to access the state synchronously as it's
-    // too late to trigger an external state change.
-    latestComparedState.current = current
-
     return true
   }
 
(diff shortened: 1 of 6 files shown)


promer94:
> I'm not sure why there's a failing test case


The test is failing due to changes in the implementation of `isEqual`, resulting in more calls to `compare` than in the previous version. We just need to update the snapshots

promer94 (commented):
The most difficult part of this bug is that when the value of stateDependencies.data changes, returnData has actually been calculated through `getSnapshot`. 

Alternatively,  I think that we can modify the implementation of `getSnapshot`. 

By updating the value of `memorizedSnapshot` but not modifying its reference, we can ensure that the value returned by `useSES` is always synchronized with the cache and won't cause any unnecessary rerenders.

examples/basic/pages/index.js
```diff
@@ -1,21 +1,49 @@
-import Link from 'next/link'
-import fetch from '../libs/fetch'
+// import Link from 'next/link'
```
huozhi, line 1: remove the comments?

core/use-swr.ts
```diff
@@ -174,9 +171,17 @@ export const useSWRHandler = <Data = any, Error = any>(
     return [
       () => {
         const newSnapshot = getSelectedCache(getCache())
-        return isEqual(newSnapshot, memorizedSnapshot)
-          ? memorizedSnapshot
-          : (memorizedSnapshot = newSnapshot)
+        const compareResult = isEqual(newSnapshot, memorizedSnapshot)
+        if (compareResult) {
+          memorizedSnapshot.data = newSnapshot.data
```
huozhi, line 176: why we update properties when they're equal? maybe also add a comment here to explain?

shuding: Because the compare process only considers selected fields (the ones that you've read in the getters before). However we want to always copy over and override other fields here too, which is the key of this fix.

Agreed that it's worth a comment, maybe linking to this PR.

Issue #2580, opened 2023-04-24
`useSWRMutation` throw an error `useSWR is not a function` in Server-Side-Rendering


qiqiboy:
# Bug report

## Description / Observed Behavior

`useSWRMutation` will throw an error "Uncaught TypeError: useSWR is not a function" when used in SSR.

## Expected Behavior

No error should be thrown.

## Repro Steps / Code Example

```bash
$ node
Welcome to Node.js v16.16.0.
Type ".help" for more information.
> const useSWRMutation = require('swr/mutation').default; useSWRMutation('test', () => Promise.resolve([]));
Uncaught TypeError: useSWR is not a function
    at Object.default (/Users/qiqiboy/develop/swr/node_modules/swr/_internal/dist/index.js:722:16)
> 
```

## Additional Context

Version: 2.1.4

Hmoulvad:
Agree I don't think 2.4.1 is ready yet. Looking at: https://codesandbox.io/s/swr-infinite-xv8q7u?file=/src/App.js it fails as well.

florianliebig:
I get the same error in JEST testing context of a next.js app. Downgrading to 2.1.3 fixes it.

![2023-04-25 at 11 20](https://user-images.githubusercontent.com/10775702/234232762-6d38f354-1dee-4250-b8ad-d210a18bb9b5.png)

florianliebig:
For the jest case it works for me with 2.1.5 - thanks!

PR #2583, opened 2023-04-25
test: add a new test setting to run tests with build files

koba04:
#2580 only happens with build files, so it's hard to detect the bug before releasing a new version. We have some e2e tests, unfortunately, they are not failed because of the build setting.
To avoid that, I've added a new setting to run existing tests with build files.
This doesn't pass with v2.1.4.


huozhi (approved):
Nice! 👍

PR #2591, opened 2023-04-29
feat: pass arg on options.onSuccess of `useSWRMutation` and `useSWR`

benevbright:
It's useful when you build your custom hook with `useSWRMutation`. 

```javascript
const useUpdatePost = () => {
  const { trigger, isMutating } = useSWRMutation(
    key,
    (url, { arg }) => fetch(url, { method: "POST", body: arg }),
    {
      onSuccess: (data, key, config) => {
        // I want to do something with response data and arg here
        // but I can't access fetcher's arg
      },
    }
  );
  return { trigger, isMutating };
};
```

Currently, there is no way to access `arg` in `onSuccess`. Thus let's pass `arg` on `onSuccess`.

### DIFF
--- a/mutation/types.ts
+++ b/mutation/types.ts
@@ -38,7 +38,8 @@ export type SWRMutationConfiguration<
     key: string,
     config: Readonly<
       SWRMutationConfiguration<Data, Error, ExtraArg, SWRMutationKey, SWRData>
-    >
+    >,
+    arg: any
   ) => void
   onError?: (
     err: Error,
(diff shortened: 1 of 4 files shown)


mutation/types.ts
```diff
@@ -38,7 +38,8 @@ export type SWRMutationConfiguration<
     key: string,
     config: Readonly<
       SWRMutationConfiguration<Data, Error, ExtraArg, SWRMutationKey, SWRData>
-    >
+    >,
+    arg: any
```
tlatmdxo123, line 42: @benevbright how about using the ExtraArg type instead of the any type?

benevbright: sorry for late response. I found that my PR is duplicated: https://github.com/vercel/swr/pull/2286
will close this one.

Issue #193, opened 2019-12-10
Clear out Error upon re-fetch?


christianflorez:
Hi all I have some code that looks like the following:
```js
function useApi(
  service
  query,
  variables
) {
  const { data, error } = useSWR([service, query, variables], fetch, {
    shouldRetryOnError: false,
  })

  const isLoading = !data && !error

  return { isLoading, error, data }
}
```

My fetch is a simple function that uses axios under the hood to fetch data. Currently, whenever an error happens in the fetch function, `useSWR` returns that error and then gets exposed at the end of the `useApi` function. However, if a new request to that same endpoint gets created, that error persists until the fetch call terminates successfully, which clears out the error, or the fetch call fails, in which case the error is updated.

Is there a way to "clear out" the error every time the fetch function calls the same endpoint? That way, an error is potentially pending while the request is in flight, rather than having the error persist until a fetch finalizes and determines a new error state.

jclem:
So far I've been using `SWRConfig` to clear out the entire cache on errors to prevent this:

```jsx
// import {cache} from 'swr'

<SWRConfig value={{
  onError: () => cache.clear()
}}></SWRConfig>
```

This doesn't seem like a great solution, but I _think_ it's working for now. I tried deleting just the cache key for the error, but the error still appears to be cached, anyway, in that case.

fcrezza:
Very late to the party and yes swr will automatically clear an error in specific endpoint if the result of that endpoint was successful but in case you want to fetch data again after the error occurs in the first fetch, you can clear an error in a specific endpoint by just calling mutate and pass null value like this
```js 
mutate('/endpoint', null, true)
```
the code above will automatically clear the error related to the endpoint and swr will revalidate the endpoint, this is not the case if we pass an undefined value, the error still persist
```js 
// import useSWR from 'swr'
const {data, error, mutate} = useSWR('/api_endpoint', fetcher)
mutate(undefined, true)
console.log(error) // error still persist
``

**Comment by shuding** [comment]
You can do this to manually clear the error state as well, **without starting a new revalidation**:

```js
const { data, error, mutate } = useSWR('/api_endpoint', fetcher)
mutate(v => v, false)
```

avispeng:
> You can do this to manually clear the error state as well, **without starting a new revalidation**:
> 
> ```js
> const { data, error, mutate } = useSWR('/api_endpoint', fetcher)
> mutate(v => v, false)
> ```

This seems to stop working now. Without revalidating, error will not be cleared.

[...]

PR #2592, opened 2023-04-29
fix: reset the error when mutate succeeded

### LINKED ISSUE #2586: error not cleared upon mutate() without revalidation
# Bug report

Hi there. We have a custom hook `useUser()` that calls `useSWR('/api/user')` inside. There is a component A that uses `useUser()` and diables `shouldRetryOnError`. Inside component A, `mutate(customUser, {revalidate: false})` is used to mutate the user data, and it will then lead to unmount of component A, and mount of component B which also uses `useUser()`. 

The current behaviour is that, useUser() will not fetch the data on mount of component B because the error is not cleared, and retry-on-error is disabled (in component A only). Explicitly using `revalidateOnMount` in useUser() in component B doesn't seem to help. We have to call mutate() explicitly on mount of component B (inside `useEffect()` hook) to fetch the data.

## Expected Behavior

The expected behaviour is that, when error happens in useUser() in component A, it will not retry. When A is unmounted, and B is mounted, useUser() inside B should fetch the data.

## Repro Steps / Code Example

I have made a sandbox that shows the error will not be cleared upon mutation without revalidation. [CodeSandbox](https://codesandbox.io/s/modest-beaver-by3kvn?file=/src/App.js)
In the sandbox, if swr@1.x.x is used, the behaviour is expected. the issue seems to be only in swr@2.x.x

## Additional Context

SWR version: 2.1.5

While I am searching the old issues in the repo, I found [this](https://github.com/vercel/swr/issues/193#issuecomment-931928418) which says the error should be cleared in mutate() without revalidation.

I am wondering if this is a intentional change in v2.

Thanks

koba04:
fixes #2586

Currently, there is no way to reset the `error` without revalidation. However, in v1, we can do this by `mutate(v => v, false)` that @shuding mentioned at https://github.com/vercel/swr/issues/193#issuecomment-931928418.

That change was introduced at #2182 is to avoid propagating `mutate` errors to fetching errors(global cache). This this change makes sense, but the behavior that `mutate` clears `error` when it succeeds also makes sense because we no longer need the `error` after `mutate` succeeded.

@shuding What do you think?

If this doesn't make sense, I think it would be nice if we provide another way to reset `error` without revalidation.


promer94 (approved):

pkunat:
Hello @shuding & @huozhi,
Are you planning to add this fix to 2.2.0/2.2.1? I couldn't find it in the changelog and the code is also not on any of the beta branches. I tried to use the main branch, but got stuck trying to find a way to access the global mutate function, which got moved to modules that are no longer exported (intentional?).

huozhi:
@pkunat sorry for missing the changelog, just published the release log of [v2.2.0-beta.1](https://github.com/vercel/swr/releases/tag/v2.2.0-beta.1). We're testing for v2.2.0 now, will release soon

Issue #2477, opened 2023-03-05
useSWRMutation has generics in different order compared to SWRMutationResponse


vegidio:
# Bug report

## Description / Observed Behavior

I noticed that the generics used in `useSWRMutation`

```typescript
<Data = any, Error = any, SWRMutationKey extends Key = Key, ExtraArg = never>
```

are in a different order compared to `SWRMutationResponse`

```typescript
<Data = any, Error = any, ExtraArg = never, SWRMutationKey extends Key = Key>
```

`SWRMutationKey` and `ExtraArg` are in different places on each case.

## Expected Behavior

I think it would make more sense to have the generics on the same order everywhere. This is a little confusing and I lost more time than I would like to admit trying to figure out why my code was not working, until I realized that the order was changed.

## Repro Steps / Code Example

The code below doesn't work, even though the generics on `SWRMutationResponse` appear on the same order as on `useSWRMutation`:

```typescript
function useTest(): SWRMutationResponse<User, any, Credentials, Key> {
    return useSWRMutation<User, any, Credentials, Key>('something', () => {});
}
```

But if I swap places between `ExtraArg` and `SWRMutationKey` then it works, but it's confusing imo:

```typescript
function useTest(): SWRMutationResponse<User, any, Credentials, Key> {
    return useSWRMutation<User, any, Key, Credentials>('something', () => {});
}
```

## Additional Context

Version 2.0.4

PR #2604, opened 2023-05-04
types: improve `useSWRMutation` type.

### LINKED ISSUE #2516: useSWRMutation's trigger response type adds `undefined`
# Bug report

`trigger` response type adds `undefined`

https://github.com/vercel/swr/blob/1389391fa0300b5baa51a9195fb0971b58071318/mutation/types.ts#L76

## Expected Behavior

Since you would use it with `try/catch` (also in the example in the document), it would be better to be `Promise<data>` rather than `Promise<data | undefined>`

SWR version. 
2.1.0

promer94:
close: #2516 
close: #2477

## Previous
The return type of `trigger` will always be `Data | undefined`
```tsx
const { trigger } = useSWRMutation('/api/user', fetcher)

expectType<Promise<string | undefined>>(trigger())
```

## Now
The return type of `trigger` will be `Data` by default, unless `throwOnError` is set to `false`

```tsx
const { trigger } = useSWRMutation('/api/user', fetcher)

expectType<Promise<string>>(trigger())
expectType<Promise<string | undefined >>(trigger(null, { throwOnError: false }))
```

```tsx
const { trigger } = useSWRMutation('/api/user', fetcher, { throwOnError: false })

expectType<Promise<string|undefined>>(trigger())
expectType<Promise<string>>(trigger(null, { throwOnError: true }))
```


shuding (approved):
Thanks!

benevbright:
cool. thanks! @promer94

PR #2649, opened 2023-06-03
fix: keepPreviousData should also work in suspense

### LINKED ISSUE #2650: Any plans on supporting suspense combined with keepPreviousData?
### Discussed in https://github.com/vercel/swr/discussions/2587

<sup>Originally posted by **WagnerMoreira** April 28, 2023</sup>
> I'm aware that suspense currently does not work with keepPreviousData or a middleware like the `laggy` from the docs example because `swr.data` is never undefined while using suspense so the features that rely on this state do not work.

I'm doing a migration on a project that had swr 0.4.x to the latest 2.1.5, and I need to use the keepPreviousData option a lot, I ended up using middleware to achieve the same because of the granular control, so this project uses suspense everywhere, and there is a <Suspense> at the root level, route levels, etc... so I'm being forced to disable suspense in some places because I need to use keepPreviousData, but that causes weird behaviors since suspense is disabled, React Query supports this so it is possible.

Would be more than wonderful if SWR could support suspense and keepPreviousData/laggy middleware at the same time.
if anyone is curious to see the differences, I created a repo to illustrate this https://github.com/WagnerMoreira/react-query-keep-prev-data</div>

promer94:
https://codesandbox.io/s/suspense-keeppreviousdata-4iotsh?file=/src/App.js

close #2650

### DIFF
--- a/core/use-swr.ts
+++ b/core/use-swr.ts
@@ -705,7 +705,7 @@ export const useSWRHandler = <Data = any, Error = any>(
         value?: boolean
         reason?: unknown
       } = revalidate(WITH_DEDUPE)
-      if (keepPreviousData && !isUndefined(returnedData)) {
+      if (!isUndefined(returnedData)) {
         promise.status = 'fulfilled'
         promise.value = true
       }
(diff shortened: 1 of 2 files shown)


shuding (approved):
Nice fix, thanks!

Issue #2653, opened 2023-06-04
Server action mutation doesn't flush updates until the action finishes


keyz:
# Bug report

## Description / Observed Behavior

(see repro below)

I'm experimenting with using (abusing? 🤔) a server action as a mutation fetcher, and I found that a server action mutation doesn't flush updates until the action finishes.

My hypothesis is it probably has something to do with both `swr` and React marking the update as a transition, therefore starving the update, but I haven't got to the bottom of it yet; wanted to open an issue early to see if y'all have any feedback/thoughts.

## Expected Behavior

The mutation should flush an `isMutating` update immediately

## Repro Steps / Code Example

https://codesandbox.io/p/sandbox/determined-dubinsky-wgf568?file=%2Fsrc%2Fapp%2Fpage.tsx%3A17%2C19

- The first "trigger" button doesn't flush updates until the action finishes
- The second "trigger" updates `isMutating` immediately as expected

<img width="489" alt="image" src="https://github.com/vercel/swr/assets/2268452/d711caec-ea00-410d-b8af-9bc490d515a2">

## Additional Context

Tested on latest canary versions: 
- swr: `2.2.0-beta.0`
- next: `13.4.5-canary.4`

promer94:
Thanks for your Repro!

keyz:
Thanks for the quick turnaround @promer94 -- would love to see a new beta release :)

huozhi:
Reopen this issue as reverted in #2681 
x-ref: https://github.com/vercel/swr/pull/2654#issuecomment-1599307815

PR #2654, opened 2023-06-05
fix: remove `startTransition` so mutation hook could update immediately

promer94:
Currently the mutate hook state is updated in `startTransition`. So when it is used with server action, it won't flush the state update into dom immediately.  

close: #2653


shuding (approved):

sebmarkbage:
This doesn't seem like the right fix and is a breaking change.

Seems like only the isMutating flag should be outside the startTransition or better yet, use the isPending flag from useTransition.

Issue #2239, opened 2022-11-11
Error not catched by `onError` using `suspense`


thisisclint21:
# Bug report

## Description / Observed Behavior

I'm trying to use `SWR` with `suspense`, and the error (401 in my case) is not catched by `onError` when using suspense components.

Also I got a visual error like this:

<img width="503" alt="Screenshot 2022-11-11 at 21 05 36" src="https://user-images.githubusercontent.com/5221349/201423280-2a4f850c-4c78-4bbd-9e5d-35e5634760b8.png">

## Expected Behavior

Retrieve the error in `onError` and not visually.

## Repro Steps / Code Example

```js
<ErrorBoundary fallback={<p>TEST</p>}>
	<Suspense fallback={<UserSummarySkeleton />}>
		<UserSummary />
	</Suspense>
</ErrorBoundary>
```

```js
const { data: userSummary, error } = useSWR<UserSummary>(userId && `/v3/users/${userId}`, {
    suspense: true,
    fallback: {}
});
```

## Additional Context

```
"next": "^13.0.3",
"swr": "^1.3.0",
"react-error-boundary": "^3.1.4",
```

koba04:
@thisisclint21 could you reproduce this on CodeSandbox?

mehm8128:
Hello, I encountered same issue.
I reproduced this issue on CodeSandbox, so please check it.
When you remove `suspense: true` from global settings, this become work correctly.
https://codesandbox.io/s/swr-state-forked-34iwp7?file=/src/Component.js

brunubarbosa:
Coming across the same issue I found this in the Q&A section:
https://github.com/vercel/swr/discussions/747
It seems to be the expected behavior when using suspend

Issue #1093, opened 2021-04-03
If revalidateOnFocus and suspense are used together, when an error occurs onfoucs listeners are never cleaned up


gaoxiaoliangz:
# Bug report

## Description / Observed Behavior

If revalidateOnFocus and suspense are used together, when an error occurs onfoucs listeners are never cleaned up

## Expected Behavior

When errors are caught in error boundary, onfocus listeners are cleaned up, a page refocus never triggers a revalidation.

## Additional Context

SWR version: 0.5.5

promer94:
Hi, could you please provide a codesandbox repro ? 
 
FYI: https://codesandbox.io/s/suspense-focus-5c0zf?file=/src/App.js
This seems works

shuding:
I think this is tricky – when a component throws and gets caught by the error boundary, the hooks will be unmounted so you can't re-render anymore. The same applies to a Suspense boundary as well.

This needs more discussion, also related to #168.

Issue #602, opened 2020-08-17
Data is not re-validated on error with suspense: true


cinan:
# Bug report

## Description / Observed Behavior

Data is not re-validated on a component mount if fetcher throws an error and `suspense` config is on.

## Expected Behavior

Re-validating behaviour should be the same with `suspense: true` or `suspense: false`.

## Repro Steps / Code Example

https://codesandbox.io/s/vigilant-worker-dr5so?file=/src/App.js

Click on "Toogle fetch component". This will mount the component using swr and throw an error (as expected). Then, double click on the button to remount the component. In Devtools network log I can see network calls are missing when component is re-mounted.
 
With `suspense: false` the scenario causes data re-validation on every remount (in Network log I can see a new network call when component is re-mounted.

## Additional Context

SWR version: 0.3.0

promer94:
If there is an error in the cache, 
swr will throw the error immediately in suspense mode 
and it will not trigger useEffect hooks in the swr.  

So swr wont re-validated with errored cache in suspense mode.

You can find a workaroud in #160

PR #2658, opened 2023-06-09
feat: improve preload and suspense integration

promer94:
This pr improves the integration between `preload` and suspense mode. 

`preload` primarily used to solve the waterfall problem of suspense, but now it also could be used to implement retry for suspense.

close: #160 
close: #602 
close: #1907 
close: #1093 
close: #2239

### Implementation
1. allowing `useSWR` to directly access the cached `PRELOAD` promise when suspense. 
2. call `mutate` with the `PRELOAD` promise to update the data cache.

### [Example](https://codesandbox.io/p/sandbox/suspense-retry-ss4m7k)
```tsx
import { Suspense } from 'react'
import { ErrorBoundary } from 'react-error-boundary'
import useSWR, { preload } from 'swr'

let count = 0
export const fetcher = () => {
  count++
  if (count === 1) return Promise.reject('wrong')
  return Promise.resolve({ name: 'success' })
}

const key = 'manual-retry'

const Demo = () => {
  const { data } = useSWR(key, fetcher, {
    suspense: true
  })
  return <div>data: {data?.name}</div>
}

function Fallback({ resetErrorBoundary }) {
  return (
    <div role="alert">
      <p>Something went wrong</p>
      <button
        onClick={resetErrorBoundary}
      >
        retry
      </button>
    </div>
  )
}

function App() {
  return (
    <div className="App">
      <ErrorBoundary FallbackComponent={Fallback} onReset={() => preload(key, fetcher)}>
        <Suspense fallback={<div>loading</div>}>
          <Demo />
        </Suspense>
      </ErrorBoundary>
    </div>
  )
}
```


tsconfig.json
```diff
@@ -17,6 +17,8 @@
     "target": "ES2018",
     "baseUrl": ".",
     "noEmitOnError": true,
+    /** FIXME: only use this to skip build error */
+    "downlevelIteration": true,
```
huozhi, line 21: we didn't enable this by purpose https://github.com/vercel/swr/pull/2181

_internal/utils/mutate.ts
```diff
@@ -73,8 +74,7 @@ export async function internalMutate<Data>(
     const keyFilter = _key
     const matchedKeys: Key[] = []
     const it = cache.keys()
-    for (let keyIt = it.next(); !keyIt.done; keyIt = it.next()) {
-      const key = keyIt.value
+    for (const key of it) {
```
huozhi, line 77: Needs to revert due to #2181

promer94: we have already upgraded the build target to es2018 and we also already use for of in https://github.com/vercel/swr/blob/7badbb9d428d2ac48873b3a3f11d54e7cec0c7cf/_internal/utils/cache.ts#L68

I also have checked the build result and it seemed like the for of part was kept.

I guess it won't be a problem here ?

huozhi: should be safe now with es2018. then i think you can just remove the FIXME comment in tsconfig

huozhi (approved):

Issue #2665, opened 2023-06-12
Unexpected Requirement for Explicit Undefined in useSWRMutation ExtraArg


saengmotmi:
# Bug report

## Description / Observed Behavior

In the current implementation of SWR, if an `ExtraArg` exists in the `useSWRMutation` hook, it's mandatory to pass `undefined` explicitly, even if the `ExtraArg` is optional. This is due to the conditional typing that requires a value for `ExtraArg` if it exists.

In my opinion, I think the problem originated from this PR: https://github.com/vercel/swr/pull/2421

May I create a PR for this issue?

## Expected Behavior

The expected behavior is that if `ExtraArg` is optional, `undefined` should be assumed when no value is passed, instead of requiring `undefined` to be passed explicitly.

## Repro Steps / Code Example

Here is a code snippet demonstrating the issue:

```tsx
import useSWRMutation from 'swr/mutation';

// ...

const {
  data: registeredDevicesData,
  isLoading: isRegisteredDevicesDataLoading,
} = useSWR(getRegisteredDevicesRoute(), getRegisteredDevices);

// ... 

const attemptRegisterDevice = async () => {
  try {
    await registerDeviceTrigger(undefined); // TO-BE: await registerDeviceTrigger()
    // ...
  } catch (e) {
    // ...
  }
};
```

If undefined is not passed as a parameter to the trigger, the following error message will be received:

```
Expected 1-2 arguments, but got 0.ts(2554)
index.d.ts(27, 214): An argument for 'extraArgument' was not provided.
```

My Proposed solution: Modify the conditional typing in the useSWRMutation hook such that ExtraArg is optional, allowing undefined to be assumed when no value is passed. This will eliminate the need to pass undefined explicitly.

```ts
// SWRMutationResponse type
trigger: <SWRData = Data>(
  extraArgument?: ExtraArg, // TO-BE: Optional
  options?: SWRMutationConfiguration<
    Data,
    Error,
    ExtraArg,
    SWRMutationKey,
    SWRData
  >
) => Promise<Data | undefined>
```

## Additional Context
SWR version: 2.1.3

promer94:
Thanks for reporting!

> May I create a PR for this issue?

Sure ! That would be awesome ❤️

Also you could always ping me if you have any question

Riley19280:
I am also encountering this issue in v2.2.1-beta.0. Would be nice not to have to update 200+ api calls with an explicit `undefined` when updating.

PR #2666, opened 2023-06-13
Fix Conditional Typing in useSWRMutation to Allow Optional ExtraArg Without Explicitly Passing Undefined

saengmotmi:
This PR contains the fixes for the issue: https://github.com/vercel/swr/issues/2665.

This is my first contribution to SWR. Thank you in advance :)


mutation/types.ts
```diff
@@ -56,7 +56,7 @@ interface TriggerWithArgs<
   ExtraArg = never
 > {
   <SWRData = Data>(
-    extraArgument: ExtraArg,
+    extraArgument?: ExtraArg,
```
promer94, line 59: I don't think this is the right fix. And it breaks the type validation for required `args`

What we expect here are 

For required `arg`

```tsx
const { trigger, reset, data, error } = useSWRMutation(
    '/api/user',
    (_, opts: { arg: number }) => String(opts.arg)
)
// should throw error, but your change breaks this.
trigger()
// pass type validation
trigger(1)
```

For optional `arg`

```tsx
const { trigger, reset, data, error } = useSWRMutation(
    '/api/user',
    (_, opts: { arg?: number }) => String(opts.arg)
)
// pass type validation
trigger()
// pass type validation
trigger(1)
```

test/type/trigger.ts
```diff
@@ -94,9 +94,6 @@ export function useTestSWRMutation() {
   )
 
   const test = () => {
-    // @ts-expect-error `arg` should be 'foo'
```
promer94, line 97: This part is actually relevant, you shouldn't remove them

promer94 (changes requested):
I think the correct way to fix this problem is to have a new `TriggerWithOptionsArgs` type. 

So you can improve the type inference implementation here.
https://github.com/vercel/swr/blob/7badbb9d428d2ac48873b3a3f11d54e7cec0c7cf/mutation/types.ts#L142-L144

`ExtraArg = never`  ->  `TriggerWithoutArgs`
`ExtraArg = number | undefined` -> `TriggerWithOptionsArgs`
`ExtraArg = number` -> `TriggerWithArgs`


Also. You need to add a new test for optional `arg`

saengmotmi:
@promer94 Oh, I see. I had made incorrect changes. I understand your suggestion. I've made changes according to your suggestion.

I introduced a new `TriggerWithOptionsArgs` type and made the appropriate adjustments for type inference. I have also added a new test for the optional argument.

Please review my latest changes and let me know if there's anything else I need to address.

promer94 (approved):
Thanks!

PR #2668, opened 2023-06-15
Pass displayed data as second parameter of functional optimistic data

francescogior:
Pass `displayedData` as second parameter to functional `optimisticData`.

To better support slow networks, one may want to use `displayedData` rather than `currentData` when using functional `optimisticData` so that several mutations can apply the optimistic updates before any is resolved and keep the UI consistent with the user actions.

### BEFORE 
https://github.com/vercel/swr/assets/3280300/c50abcd4-445d-41dd-9c53-9554e7330c55

### AFTER

https://github.com/vercel/swr/assets/3280300/cb9d37d1-78ab-4dbf-a8e3-03a2fb6b5842

### DIFF
--- a/_internal/src/types.ts
+++ b/_internal/src/types.ts
@@ -317,7 +317,9 @@ export type MutatorOptions<Data = any> = {
   populateCache?:
     | boolean
     | ((result: any, currentData: Data | undefined) => Data)
-  optimisticData?: Data | ((currentData?: Data) => Data)
+  optimisticData?:
+    | Data
+    | ((currentData: Data | undefined, displayedData: Data | undefined) => Data)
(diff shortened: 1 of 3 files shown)


huozhi:
Thanks for the PR! Can you add a test for this case?

francescogior:
> Thanks for the PR! Can you add a test for this case?

Test case added

huozhi (approved):

huozhi:
Thanks!

Rowanto:
Man, today I was randomly looking into the optimisticData signature and found this second params,.. @francescogior you made our day here in office,..

francescogior:
Great to hear that☺️

PR #2691, opened 2023-06-24
fix: only make data and error update as a non-blocking transition

promer94:
close: #2653


test/use-swr-remote-mutation.test.tsx
```diff
@@ -784,7 +784,7 @@ describe('useSWR - remote mutation', () => {
     await act(() => sleep(20))
 
     // Shouldn't have intermediate states
-    expect(logger).toEqual([undefined, 2])
```
promer94, line 787: It has extra render because `isPending` changes

mutation/src/index.ts
```diff
@@ -76,15 +75,19 @@ const mutation = (<Data, Error>() =>
 
           // If it's reset after the mutation, we don't broadcast any state change.
           if (ditchMutationsUntilRef.current <= mutationStartedAt) {
-            setState({ data, isMutating: false, error: undefined })
+            startTransition(() =>
```
promer94, line 78: we use `React.startTransition` here since `isPending` from `useTransition` will always cause rerender even if `isMutation` is not used.

shuding (approved):
Thanks!

Issue #2208, opened 2022-10-22
Default export type not being picked up with `module: "nodenext"`


wight554:
# Bug report

## Description / Observed Behavior

When trying to use imported useSWR hook I'm getting the following error:
```
import useSWR
This expression is not callable.
  Type 'typeof import("/Users/Volodymyr_Zhdanov/playground/blog-template/node_modules/swr/dist/index")' has no call signatures.
 ```

## Expected Behavior

Import should pick-up typings

## Repro Steps / Code Example

1. Add these settings to tsconfig:
```
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
```
2. Add `"type": "module"` to package.json
3. Import and use useSWR hook
4. See the error

## Additional Context
SWR version: ^1.3.0
Typescript version: ^4.8.4

3 possible fixes:

1. Add named export for useSWR
2. Use CJS styled export in typings: `export default useSWR;` => `export = useSWR;`
3. Use declaration merging to make default export namespace and function simultaneously, example here: https://github.com/fastify/fastify-cookie/pull/184/

**Comment by promer94** [comment]
Ref: https://github.com/microsoft/TypeScript/issues/50466

**Comment by wight554** [comment]
> Ref: [microsoft/TypeScript#50466](https://github.com/microsoft/TypeScript/issues/50466)

This can be referenced as well: https://github.com/microsoft/TypeScript/issues/48845

**Comment by promer94** [comment]
After TypeScript 5.0 got released.

This could be a workaround
```json
{
  "compilerOptions": {
    "moduleResolution": "bundler",
   }
}
```
https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#moduleresolution-bundler

dubiousdavid:
This still doesn't work for me.

PR #2695, opened 2023-06-26
build: generate d.mts for for .mjs, so typescript could resolve types correctly

promer94:
upgrade bunchee to generate a separate `d.mts` for for `.mjs`.

ensure the declaration file extension always matches the JS file extension.

fix [FalseCJS](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseCJS.md)  problem.

add [@arethetypeswrong/cli](https://www.npmjs.com/package/@arethetypeswrong/cli) in ci to prevent future issues with TypeScript types

close: #2208


mutation/package.json
```diff
@@ -3,9 +3,7 @@
   "module": "./dist/index.esm.js",
   "types": "./dist/index.d.ts",
   "exports": {
-    "types": "./dist/index.d.ts",
     "import": "./dist/index.mjs",
-    "module": "./dist/index.esm.js",
```
huozhi, line 8: why we need to delete module field?

huozhi (commented):
Can we separate the unrelated changes such as actions job changes and other modules upgration into different PR? This would help issue tracking in the future and reduce the chances of breaking

mutation/package.json
promer94: https://webpack.js.org/guides/package-exports/#target-environment-independent-packages

I think this field is designed for `node` only.

It prevents packages from being bundled twice when referencing packages with `require()` and `import` at same time. 

So i think we probably don't need this feature here ?

package.json
```diff
@@ -82,6 +86,7 @@
     "build": "pnpm build-package _internal && pnpm build-package core && pnpm build-package infinite && pnpm build-package immutable && pnpm build-package mutation && pnpm build-package subscription",
     "build:e2e": "pnpm next build e2e/site",
     "build-package": "bunchee --cwd",
+    "attw": "attw --pack",
```
huozhi, line 89: what does `attw` do here? seems not in the dependency

mutation/package.json
huozhi: Feel it's still cirtical to change here as it could be bundled on server side in nodejs runtime. Also it's not related to the types generation. Would be good to create a separate PR and plan in a minor release instead of carrying it into this PR

huozhi (commented):
Can we update the PR description to mention the approach and the changes?

package.json
promer94: https://github.com/arethetypeswrong/arethetypeswrong.github.io


It's like publint, but for types

mutation/package.json
promer94: 👌

huozhi (approved):

Issue #2221, opened 2022-11-01
Add `"client-only"`


shuding:
SWR is highly depending on client React features like state, context, effect, etc. so it can't be used in server components. The best way to avoid this is to add a `"client-only"` import to the core.

zenflow:
Then you need a different API for fetching data in server components. ☹️

shuding:
SWR can't run on the server side (e.g. during SSR), that's the case from the beginning. And server components won't be hydrated on the client, so everything inside them will not run on the browser.

One way to think about it is that you can't run `useSWR` inside `getServerSideProps`.

PR #2696, opened 2023-06-26
Add use client directive for client components exports

huozhi:
Upgrade bundler (where it supports preserving the "use client" directive now) and add `"use client"` for the all API except `react-server` exports

Addressing few RSC related discussion, since we've already provided the `react-server` exports condition for server components now we can add `"use client"` directive to the API which only should be used in client components

Related to #2221
Rsolves #2292
Closes #2464

### DIFF
--- a/core/src/index.ts
+++ b/core/src/index.ts
@@ -1,3 +1,5 @@
+'use client'
+
 // useSWR
 import useSWR from './use-swr'
 export default useSWR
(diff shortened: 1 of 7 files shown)


shuding (approved):
Although we don't need the directive for APIs other than `SWRConfig`, it's fine to have this added for now.

promer94 (approved):
LGTM

huozhi:
@shuding if we want to control the client boundary as small as possible, it needs to split that file as an extra chunk or not bundling at all. Splitting into extra chunk sounds possible to do, I can investigate it later for the bundler updates

Issue #2690, opened 2023-06-24
The type and runtime behavior of `mutate/trigger` fn  are mismatched


promer94:
# Problem
The runtime behavior of `mutate/trigger` is complex and the current type does not correctly match the actual behavior.

```tsx
mutate('/key', fetcher, { populateCache: true, revalidate: true }) // -> revalidation result
```

```tsx
mutate('/key', fetcher, { populateCache: true, revalidate: false }) // -> fetcher result
```

```tsx
mutate('/key', fetcher, { populateCache: () => Data, revalidate: true }) // -> revalidation result
```

```tsx
mutate('/key', fetcher, { populateCache: () => Data, revalidate: false }) // -> Data
```

```tsx
mutate('/key', fetcher, { populateCache: false, revalidate: true }) // -> fetcher result
```

```tsx
mutate('/key', fetcher, { populateCache: false, revalidate: false }) // -> fetcher result
```

## related issues
#2647 #2661

westmark:
The type `SWRMutationConfiguration` looks like this in 2.2.0

```
type SWRMutationConfiguration<Data, Error, SWRMutationKey extends Key = Key, ExtraArg = any, SWRData = any> = {
...
populateCache?: boolean | ((result: Data, currentData: SWRData | undefined) => SWRData);
    optimisticData?: SWRData | ((currentData?: SWRData) => SWRData);
```

However, S

Ponyets:
I also got this issue.
@promer94 What do you think should be the correct behavior of mutate/trigger?
I suggest it always return the fetcher's result, undefined if fetcher isn't provided, because sometimes we have to mutate local data in a chain.

For example, in an IM app, when user sends a new message to a conversation, we need to insert this message to local message list and update the corresponding conversation's latest message.
I'm not sure if the following is the best practice, anyway it's our current way:
```Typescript
export function useSendMessage(conversationId:string, message: string) {
    const insertMessage: () => Promise<Message> = () => {
           // promise that insert message through a post request
    }
    const {trigger: triggerMessages} = useSWRMutation<Message[]>(MESSAGGS_KEY_OF(conversationId), post, {
        populateCache(result, currentData) => {
              // the first issue here, type mismatched, 
              // mutate/trigger expects the fetcher's response type the same as the existing data,
              // which is not always true
              const insertedMessage = result as unknown as Message
              return [message, ...currentData]
        }
    })
[...]

Issue #2661, opened 2023-06-09
Wrong type for trigger function in useSWRMutation hook


jauniusmentimeter:
# Bug report

## Description / Observed Behavior

Wrong type for trigger function in `useSWRMutation` hook

## Expected Behavior

See code example below. Expected return type for `trigger` in this case should be `Promise<Item[] | undefined>` and not `Promise<Item | undefined>`

## Repro Steps / Code Example

```ts
  type Item = { id: number };

  const fetchKey = "myKey";

  const { data } = useSwr(fetchKey, fetcher); // data is of type Item[]

  const updateSpecificItem = () =>
    new Promise<Item>((resolve) => resolve({ id: 1 })); // this has type Promise<Item>

  const { trigger } = useSWRMutation(fetchKey, updateSpecificItem, {
    populateCache: (updatedItem, items) => {
      const filteredItems = items.filter((item) => item.id !== 1);
      return [...items, updatedItem];
    },
  });

  return (
    <button
      onClick={async () => {
        // result has type Item | undefined, but it should be Item[] | undefined,
        // since that's what we get from populateCache
        const result = await trigger();
      }}
    >
      update item
    </button>
  );
```

## Additional Context

SWR version : 2.1.5

promer94:
Closing in favor of #2690

Issue #2647, opened 2023-05-31
MutatorCallback type too strict when passing a function to populateCache


svenlombaert:
# Bug report

## Description / Observed Behavior

So, I have kind of a complicated example where the typings of the `MutatorCallback` are actually too strict. It's typed with return type `Data` which is good if you use the `populateCache` option with `true`.

There's also the option to use `populateCache` with a function, which is correctly typed as this: `(result: any, currentData: Data | undefined) => Data`. `result` could be `any` but not in `MutatorCallback` where it's restricted to `Data`

## Expected Behavior

It should accept `any` as a return type in the `MutatorCallback` type, preferably *only* when a callback is passed to `populateCache`

## Repro Steps / Code Example

This is a bit hard to set up but I've got something like this:

```typescript
export interface ConnectionListApiResponse {
  EXCHANGE: ExchangeConnection[];
  WALLET: Wallet[];
  BROKER: BrokerConnection[];
}

const { data: connectionsData, mutate: mutateConnectionList } =
  useRequest<ConnectionListApiResponse>(
    {
      url: '/connection/list',
    },
  );
```

I then have separate polling endpoints which return a partial `ConnectionListApiResponse`, one for BROKER, one for WALLET and one for EXCHANGE. My mutate looks something like this:

```typescript
const fetchWallets = async () => {
    const walletConnectionResponse =
      await api.get<{ success: boolean; connections: Wallet[] }>(
          '/wallet-connection'
        );

    return {
      WALLET: walletConnectionResponse.data.connections,
    };
  };

  await mutateConnectionList(fetchWallets(), {
    populateCache: (
      result: Pick<ConnectionListApiResponse, 'WALLET'>,
      currentData,
    ) => {
      if (!currentData) {
        return result;
      }

      return {
        ...currentData,
        data: {
          ...currentData.data,
          ...result,
        },
      };
    },
    revalidate: false,
  });
```

## Additional Context

SWR version: 2.1.3

promer94:
For your specific use case. I would recommend you not to use `populateCache`.

Instead, you could do

```tsx
mutateConnectionList(
  (current = {}) => 
    fetchWallets()
    .then(v => ({...current, WALLET: v}))
    .catch(() => current)
)
```

to partially update the cache.

promer94:
Closing in favor of #2690

PR #2708, opened 2023-07-11
fix(mutate): fix types of mutate/trigger; make mutate/trigger always return the result of fetcher

Ponyets:
Fix for [issue 2690](https://github.com/vercel/swr/issues/2690) [issue 2647](https://github.com/vercel/swr/issues/2647) and [issue 2661](https://github.com/vercel/swr/issues/2661)

1. Make mutate/trigger always return the result of fetcher
2. Make KeyedMutator/ScopedMutator accept generic type parameters to make the data type flexible and work with populateCache

### DIFF
--- a/_internal/src/types.ts
+++ b/_internal/src/types.ts
@@ -310,11 +310,15 @@ export type MutatorCallback<Data = any> = (
   currentData?: Data
 ) => Promise<undefined | Data> | undefined | Data
 
-export type MutatorOptions<Data = any> = {
+/**
+ * @typeParam Data - The type of the data related to the key
+ * @typeParam MutationData - The type of the data returned by the mutator
+ */
+export type MutatorOptions<Data = any, MutationData = Data> = {
   revalidate?: boolean
   populateCache?:
     | boolean
-    | ((result: any, currentData: Data | undefined) => Data)
+    | ((result: MutationData, currentData: Data | undefined) => Data)
   optimisticData?:
     | Data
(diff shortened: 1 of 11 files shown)


Ponyets:
Hello. 
It seems this PR need an approval to run the test workflow, since I had to modify a test file.
Could a maintainer kindly give it a try? Thanks.

Ponyets:
I ran `pnpm run-all-checks` locally this time. It passed.
Fixed `FIXME` left by @shuding in `test/type/mutate.ts`.

_internal/src/types.ts
```diff
@@ -310,14 +310,17 @@ export type MutatorCallback<Data = any> = (
   currentData?: Data
 ) => Promise<undefined | Data> | undefined | Data
 
-export type MutatorOptions<Data = any> = {
+export type MutatorOptions<SWRData = any, Data = SWRData> = {
```
shuding, line 313:
```suggestion
export type MutatorOptions<Data = any, MutationData = Data> = {
```

Maybe this naming is better? I think we’d like to have a JSDoc here explaining what these generics mean actually. Still kinda confusing.

shuding (commented):
Thanks for the PR, it looks great in general!

_internal/src/types.ts
Ponyets: Updated as your suggestion, JSDoc added as well.

PR #2723, opened 2023-07-25
fix(infinite): Fix the ability to use preload along with useSWRInfinite

agadzik:
This PR fixes the use-case where I want to preload the first page of data from useSWRInfinite using the `preload` function from `swr`. 

The issue was that `useSWRInfinite` prefixes the cache key's internally.  Adding additional logic to the preload `middleware` function to handle the condition where the key starts with INFINITE_PREFIX as the consumer calling `preload` does not know about the prefix.

### DIFF
--- a/_internal/src/utils/preload.ts
+++ b/_internal/src/utils/preload.ts
@@ -50,8 +50,7 @@ export const middleware: Middleware =
 
         let normalizedKey = key
         if (key.startsWith(INFINITE_PREFIX)) {
-          normalizedKey = key.replace(INFINITE_PREFIX, '')
-          console.log(key, normalizedKey)
+          normalizedKey = key.slice(INFINITE_PREFIX.length)
         }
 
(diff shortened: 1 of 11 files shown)


_internal/src/utils/preload.ts
```diff
@@ -46,9 +47,16 @@ export const middleware: Middleware =
       ((...args: any[]) => {
         const [key] = serialize(key_)
         const [, , , PRELOAD] = SWRGlobalState.get(cache) as GlobalState
-        const req = PRELOAD[key]
+
+        let normalizedKey = key
+        if (key.startsWith(INFINITE_PREFIX)) {
+          normalizedKey = key.replace(INFINITE_PREFIX, '')
+          console.log(key, normalizedKey)
```
shuding, line 54:
```suggestion
          normalizedKey = key.slice(INFINITE_PREFIX.length)
```

PR #2731, opened 2023-07-31
Fix the issue that useSWR revalidation isn't triggered if the useSWR call happens after mutation

Ponyets:
E.g.:
```
   req------------->res
mutate------->end
```
In this case, the res would be ignored while req would not be triggered a revalidation.


Ponyets:
My current workaround is to mutate again in the onSuccess callback of the first mutation.

Ponyets:
Test case added.

PR #2734, opened 2023-08-02
docs: use isLoading instead of !data

pawfa:
Hello,

I was going to use swr and when I was looking at the github readme I've noticed that the loading state was handled by `!data`. However on the swr docs main page https://swr.vercel.app/ overview example has the `isLoading` value.

I thought that it would be nice to have this readme example consistent with website.


shuding (approved):
Thank you!

PR #2753, opened 2023-08-21
fix: default to fetch type in keyed mutator

### LINKED ISSUE #2744: Mutate typescript
# Bug report

## Expected Behavior

## Repro Steps / Code Example

image

image

image

image

## Additional Context

SWR version. 
2.2.1

The 2.2.0 version works fine for me

linkvt:
Closes #2744 

Hi,

as a few other also noticed the typings used in the bound mutator do not default anymore to the data type used in `useSWR<ThisOne>` and needs to be specified explicitly everytime right now.

Looking at the other types surrounding the `KeyedMutator` type it seems reasonable to default to the data type specified at `useSWR<...>`, e.g. as done here:

https://github.com/vercel/swr/blob/e510955d42b472c8d45034c51b8acc12980aa2e7/_internal/src/types.ts#L372-L381

Let me know if and where a test could be added if you want to test the typings.

Thanks, Vincent

### DIFF
--- a/_internal/src/types.ts
+++ b/_internal/src/types.ts
@@ -395,12 +395,9 @@ export interface ScopedMutator {
  * @typeParam MutationData - The type of the data returned by the mutator
  */
 export type KeyedMutator<Data> = <MutationData = Data>(
-  data?:
-    | MutationData
-    | Promise<MutationData | undefined>
-    | MutatorCallback<MutationData>,
+  data?: Data | Promise<Data | undefined> | MutatorCallback<Data>,
   opts?: boolean | MutatorOptions<Data, MutationData>
-) => Promise<MutationData | undefined>
+) => Promise<Data | MutationData | undefined>
 
 export type SWRConfiguration<
   Data = any,
--- a/_internal/src/types.ts
+++ b/_internal/src/types.ts
@@ -394,7 +394,7 @@ export interface ScopedMutator {
  * @typeParam Data - The type of the data related to the key
  * @typeParam MutationData - The type of the data returned by the mutator
  */
-export type KeyedMutator<Data> = <MutationData>(
+export type KeyedMutator<Data> = <MutationData = Data>(
   data?:
     | MutationData
     | Promise<MutationData | undefined>
--- a/test/type/mutate.ts
+++ b/test/type/mutate.ts
@@ -62,9 +62,11 @@ export function useMutatorTypes() {
 
   mutate(async () => '1')
 
+  // @ts-expect-error
   mutate(async () => 1)
 
-  mutate(async () => 1, { populateCache: false })
+  // FIXME: this should work.
+  // mutate(async () => 1, { populateCache: false })
 }
 
 export function useConfigMutate() {
--- a/infinite/src/index.ts
+++ b/infinite/src/index.ts
@@ -232,8 +232,12 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
 
     const mutate = useCallback(
       // eslint-disable-next-line func-names
-      function <T>(
-        data?: undefined | T | Promise<T | undefined> | MutatorCallback<T>,
+      function <T = Data[]>(
+        data?:
+          | undefined
+          | Data[]
+          | Promise<Data[] | undefined>
+          | MutatorCallback<Data[]>,
         opts?: undefined | boolean | MutatorOptions<Data[], T>
       ) {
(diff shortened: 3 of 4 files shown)


linkvt:
Hello @shuding or @huozhi , would you mind looking at this tiny PR fixing the types of a keyed mutator? I just rebased the code, let me know if something is missing here.
Thanks!

huozhi:
Hi, could you also add a types test for this case so we can avoid the regression later? Thanks

promer94 (approved):
Thank you !

test/type/mutate.ts
```diff
@@ -62,9 +62,11 @@ export function useMutatorTypes() {
 
   mutate(async () => '1')
 
+  // @ts-expect-error
   mutate(async () => 1)
 
-  mutate(async () => 1, { populateCache: false })
+  // FIXME: this should work.
+  // mutate(async () => 1, { populateCache: false })
```
huozhi, line 69: @promer94 thanks for adding the test, maybe we should create an issue to follow up on this? it feels like a regression for `mutate()` API types

promer94: The original pr #2708 did not fix this correctly so there is no regression here.

I also think we should not support this because the type implementation would be really complicated and not useful in practice. You should always make sure you data has same structure as cache when using the bounded mutate fn.

I just keep it here as a reference.

huozhi: right the type is number, I updated them in https://github.com/vercel/swr/pull/2781

AmirHosseinKarimi:
Hey guys. Thanks for improving the SWR.
When can we have these changes on the new version?
Is there any routine release plan?

huozhi:
@AmirHosseinKarimi A few type improvements including this are landed in [2.2.3](https://www.npmjs.com/package/swr/v/2.2.3)

_internal/src/types.ts
```diff
@@ -394,13 +394,10 @@ export interface ScopedMutator {
  * @typeParam Data - The type of the data related to the key
  * @typeParam MutationData - The type of the data returned by the mutator
  */
-export type KeyedMutator<Data> = <MutationData>(
-  data?:
-    | MutationData
-    | Promise<MutationData | undefined>
-    | MutatorCallback<MutationData>,
+export type KeyedMutator<Data> = <MutationData = Data>(
+  data?: Data | Promise<Data | undefined> | MutatorCallback<Data>,
```
ckohen, line 398: @promer94 Why was this change reverted? It makes the generic completely useless and returning different `MutationData` to be passed to `populateCache` impossible with proper types.

This probably needs to be overload based to function properly, otherwise `MutationData` will just be inferred improperly from the data arg, although it is mostly possible to do without overloads with some rather extensive types:

```ts
type KeyedMutator<Data> = <MutationData = Data>(data?: MutationData | Promise<MutationData | undefined> | MutatorCallback<MutationData>, opts?: boolean | Omit<MutatorOptions<Data, MutationData>, 'populateCache'> & (MutationData extends Data ? Pick<MutatorOptions<Data, Data>, 'populateCache'> : { populateCache: Exclude<MutatorOptions<Data, MutationData>['populateCache'], boolean> })) => Promise<Data | MutationData | undefined>;
```

I have tested the above and it works as expected, except in the case where `opts` is not passed or is a `boolean`. This is why I suggested the overload approach, since if options is not passed as an object, typescript doesn't enforce it being passed.

simenbrekken-visma:
Not sure if I understood the above changes completely, but after upgrading from 2.0.3 to 2.2.3 we had to update all our `populateCache` methods:

**Before (2.0.3)**

```ts
const { data = [], mutate } = useSWR<ListItem[]>('/api/list', fetcher);

const addListItem = (name: string) => {
    mutate(
        async () => {
            const item: ListItem = await createListItem(name);

            return item;
        },
        {
            populateCache: (result, currentData = []) => [
                ...currentData,
                result,
            ],
        },
    );
};
```

**After (2.2.3)**

```ts
const { data = [], mutate } = useSWR<ListItem[]>('/api/list', fetcher);

const addListItem = (name: string) => {
    mutate(
        async () => {
            const item: ListItem = await createListItem(name);

            return [item];
        },
        {
            populateCache: (result, currentData = []) => [
                ...currentData,
                ...result,
            ],
        },
    );
};
```

Not sure if I've missed a generic that can be passed to work around this, but having to modify our mutators to return the same type as given to `useSWR` feels a bit clunky.

PR #2757, opened 2023-08-24
feat: Use custom functions to hash arguments

quavious:
## Description

- Hash arguments with custom functions for some constructors.
  - Prevent converting instances to ascending numbers like `9~`.

- Can inject functions in SWR config
```tsx
<SWRConfig
  value={{
    customHashes: [
      // functions to serialize instances
    ],
  }}
>
  <Component {...pageProps} />
</SWRConfig>
```
- Tests for custom hash functions

## Discussions Previously Registered

https://github.com/vercel/swr/discussions/2685


promer94:
Thanks for your pr !

I think You could always use to middleware to customize how to hash your arguments. We currently do not have intention to add an extra option for this

PR #2759, opened 2023-08-24
fix: remove permissive type

### LINKED ISSUE #2758: SWRHook noting enforcing fetch arguments and Key types
# Bug report

While trying to enforce types of the arguments of our fetcher the type override for `SWRHooks` fell through to `BareFetcher` which allowed any arguments even if they did not match the key.

## Expected Behavior

Should throw a typescript error because the types don't match.

## Repro Steps / Code Example

```
const fetcher = ({ url }:{ url: string }) => url
// No typescript error
useSWR({ path: '/api' }, fetcher)
```

Should throw a typescript error because `{ path: string }` does not match `{ url: string }`.

## Additional Context

* SWR version 2.2.1
* Attempted to remove arguments from `BaseFetcher` which seemed reasonable that you shouldn't allow the BaseFetcher to contain arguments otherwise it's not a BaseFetcher, this caused tons of cascading issues.
* I believe I have a solution replacing the BaseFetcher override with a more basic type. It currently doesn't seem to cause cascading type issues and provides the expected behavior.
* This occurred while using the Fetcher type.

wcatron:
Resolve #2758 by remove permissive type. Because BaseFetcher allows `(...args: any)` when your key and fetcher key mismatch you fall through to the override:

```
<Data = any, Error = any>(
     key: Key,
     fetcher: BareFetcher<Data> | null
   ): SWRResponse<Data, Error>
```

This should instead be flagged by typescript. Removing that override broke some tests added in #1640 which appeared to be designed to allow for backwards compatible hooks that don't conform to `StrictKey` to be compatible. Specifically these two type tests:

```
useSWR<{ id: number }>('/api/', () => ({ id: 123 }))
useSWR<string, any>('/api/', (key: string) => key)
```

The new override covers these while retaining type inference by just not defaulting to `StrictKey` which appears to be the main issue.

### DIFF
--- a/_internal/src/types.ts
+++ b/_internal/src/types.ts
@@ -231,6 +231,10 @@ export interface SWRHook {
     key: SWRKey,
     fetcher: Fetcher<Data, SWRKey> | null
   ): SWRResponse<Data, Error>
+  <Data = any, Error = any, SWRKey extends Key = Key>(
+    key: SWRKey,
+    fetcher: Fetcher<Data, SWRKey> | null
+  ): SWRResponse<Data, Error>
   <
     Data = any,
     Error = any,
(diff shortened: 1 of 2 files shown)


promer94 (approved):
Thank you!

PR #2761, opened 2023-08-26
fix: simplify `ArgumentsTuple`

Andarist:
this was already redundant but it will actually cause problems with TS 5.3 (in the future), see: https://github.com/microsoft/TypeScript/pull/55229#issuecomment-1694383442


promer94 (approved):

promer94:
Thanks！

shuding:
Thank you @Andarist!

PR #2767, opened 2023-09-04
Remove `index.js` suffix of `use-sync-external-store/shim` to support React Native

### LINKED ISSUE #2402: Successful fetch does not cause rerender on React Native
# Bug report

On React Native project I've decided to try out latest(2.0.3) swr version (I previously was using 1.3.0 version). I've encountered issue that upon successful retrieval of data, the `useSwr` `data` property does not change - is undefined, `isLoading` also stays `true`, but for some reason `onSuccess` is called. Data and isLoading changes only whenever I change screen - this causes rerender and `data` property gets filled with data, isLoading is set to false.

## Expected Behavior

I expect the `useSwr` hook to change its property `data` to received value whenever call is successful, just like in `1.3.0` version.

## Repro Steps / Code Example

```
  const fetchCampaignById = async (id: string) => {
    return await api.fetchCampaignDetails(id, language as LanguageCode)
  }

  const { data, isLoading } = useSwr(
    ['campaigns', donations],
    () => fetchCampaignById(donations[0].campaignId),
    {
      onSuccess: () => {
        console.log('success callback')
      },
    }
  )
```

where `api.fetchCampaign` is a `Promise` which returns directly array of data once promise is resolved. Success callback is logged, api calls also do not fail when inspecting network debugger, yet data property does not change

## Additional Context

SWR version - `2.0.3`

malash:
This PR fixed #2402

According to [this document](https://reactnative.dev/docs/platform-specific-code#native-specific-extensions-ie-sharing-code-with-nodejs-and-web), React Native-specific code can use the '.native.js' suffix, while standard React web code uses the '.js' suffix.  By removing `index.js` suffix of `use-sync-external-store/shim`, bundling tools like Metro/Webpack could choose the right file for each platform.

I haven't tested this PR on React Native because I'm not familiar with React Native development and I have no idea how to downgrade React to 17. However, I can reproduce the issue and test the PR on [GojiJS](https://github.com/airbnb/goji-js),  which is a React-like framework based on `react-reconciler` and shares many technical specifications with React Native, including the '.native.js' suffix. It would be appreciated if anyone could help verify the changes.

### DIFF
--- a/core/src/use-swr.ts
+++ b/core/src/use-swr.ts
@@ -5,7 +5,7 @@ import ReactExports, {
   useDebugValue,
   useMemo
 } from 'react'
-import { useSyncExternalStore } from 'use-sync-external-store/shim/index.js'
+import { useSyncExternalStore } from 'use-sync-external-store/shim'
 
 import {
   defaultConfig,
--- a/infinite/src/index.ts
+++ b/infinite/src/index.ts
@@ -33,7 +33,7 @@ import type {
   SWRInfiniteCacheValue,
   SWRInfiniteCompareFn
 } from './types'
-import { useSyncExternalStore } from 'use-sync-external-store/shim/index.js'
+import { useSyncExternalStore } from 'use-sync-external-store/shim'
 import { getFirstPageKey } from './serialize'
 
 // const INFINITE_PREFIX = '$inf$'


promer94 (approved):

promer94:
Thank you!

maziar-dandc:
This has broken our prod builds, reproduced on node v14, v16.
Going to pin SWR to an older version for the time being.
If this is caused by a problem from our side, please feel free to point it out, I'm not a FE expert by any means.

```
[14:57:16] Error in plugin "webpack-stream"
Message:
    Module not found: Error: Can't resolve 'use-sync-external-store/shim' in '/home/~~~~/node_modules/swr/core/dist'
Did you mean 'index.js'?
BREAKING CHANGE: The request 'use-sync-external-store/shim' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
Did you miss the leading dot in 'resolve.extensions'? Did you mean '[".*",".js",".jsx"]' instead of '["*",".js",".jsx"]'?
Details:
    domainEmitter: [object Object]
    domainThrown: false

[14:57:16] 'default' errored after 21 s

```

PR #2770, opened 2023-09-05
ci: simplify ci config and bump some deps version

promer94:


package.json
```diff
@@ -92,7 +92,7 @@
     "build": "pnpm build-package _internal && pnpm build-package core && pnpm build-package infinite && pnpm build-package immutable && pnpm build-package mutation && pnpm build-package subscription",
     "build:e2e": "pnpm next build e2e/site",
     "build-package": "bunchee --cwd",
-    "attw": "attw --pack",
+    "attw": "attw --pack --ignore-rules cjs-only-exports-default",
```
huozhi, line 95: why we need to ignore this rule?

can we also run attw in the test:build process on ci so it's get covered on CI?

promer94: It has already been included in CI.  

We need to ignore this rule because SWR use default exports which could cause problems. https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/CJSOnlyExportsDefault.md#-cjs-default-export

huozhi (approved):

huozhi:
@promer94 I got it passed, would be appreciated if you can explain a bit what you initially tried to change like why they can be removed. thx

.github/workflows/trigger-release.yml
```diff
@@ -20,15 +20,12 @@ on:
 name: Trigger Release
 
 env:
-  PNPM_VERSION: 8.4.0
```
promer94, line 23: pnpm version could be directly readed from `package.json`

.github/workflows/trigger-release.yml
```diff
@@ -20,15 +20,12 @@ on:
...
   SEMVER_TYPE: ${{ github.event.inputs.semverType }}
   RELEASE_TYPE: ${{ github.event.inputs.releaseType }}
 
 jobs:
   start:
     runs-on: ubuntu-latest
-
-    environment: release-${{ github.event.inputs.releaseType }}-${{ github.event.inputs.semverType }}
```
promer94, line 31: I think we dont have multiple enviornments now, so this could be removed

.github/workflows/trigger-release.yml
```diff
@@ -53,11 +52,6 @@ jobs:
           node ./scripts/bump-next-version.js
 
       - name: Git push
-        env:
-            RELEASE_BOT_GITHUB_TOKEN: ${{ secrets.RELEASE_BOT_GITHUB_TOKEN }}
```
promer94, line 57: I think git never use this env

huozhi:
Thanks, sounds good!

PR #2776, opened 2023-09-06
Add option forceRevalidateOnReconnect to hard-revalidate when regaining a network connection

notuxic:
### TLDR

Adds an option `forceRevalidateOnReconnect`, which when set to `true`, triggers a revalidation after reconnecting even when `dedupingInterval` hasn't passed yet. To compensate spotty network connections, an option `reconnectThrottleInterval` allows to throttle further revalidations for this event.

### Background

I'm currently working on an app using SWR with a high `dedupingInterval` (30 seconds). Depending on network availability, the fetcher fetches data either from network or a local cache.

Now if users navigate to a screen while offline, and then turn on the network (for example because they see the "no-network" placeholder when there's no cached data), the data won't be revalidated since the 30 seconds of the `dedupingInterval` haven't passed yet.

### forceRevalidateOnReconnect

With the new option `forceRevalidateOnReconnect` set to `true` (defaulting to `false`), SWR will call `revalidate()` instead of `softRevalidate()` to bypass `dedupingInterval`, similar to mutate-events.

### reconnectThrottleInterval

On spotty network connections, the above new option may trigger a lot of unwanted/unneeded revalidations. For this reason a second option `reconnectThrottleInterval` allows to throttle revalidations on reconnect-events, similar to the existing `focusThrottleInterval` for focus-events. 

`reconnectThrottleInterval` defaults to `5000` for consistency with `focusThrottleInterval`.

#### Default value for reconnectThrottleInterval

Currently `reconnectThrottleInterval` uses the same default value as `focusThrottleInterval` for consistency.

A default value of `0` would preserve backwards-compatible behaviour on reconnect-events, but may not be as good of a default for the feature.

Maybe the default value should be even higher, since 5 seconds is quite short in the case of spotty network connections.


notuxic:
ping @shuding @huozhi, requesting review, since it's been a month since opening the PR.

PR #2800, opened 2023-09-24
feat: introduce SWRNoCacheOnMount (for suspense)

benevbright:
related issues:
- [Imposing no cache for a SWR call](https://github.com/vercel/swr/discussions/456)
- https://github.com/vercel/swr/discussions/2318

### Problem (with `suspense: true`)

Sometimes you want to start component with the latest data especially when you work with **an edit-form, which typically gives a chance to pre-fill default data only on mount.** 

So you want to force fetching on mount and not return `data` so that `<Suspense>` triggered while fetching. 

You can do something like putting a random SWR key with `useRef` **but with `suspense: true` this workaround doesn't work** because `ref` gets reset when promise is resolved which causes infinite re-render. 

### suggestion

```
<SWRNoCacheOnMount>
  <Suspense fallback={<Loader />}>
    <MyEditProfileForm>
  </Suspense>
</SWRNoCacheOnMount>
```
```
const MyEditProfileForm = () => {
  const { data } = useSWR(KEY, {
    noCacheOnMount: true,  // <-- new config
    suspense: true
  }
  // render form
}
```

EDIT: 
Meanwhile, I've written workaround: https://github.com/vercel/swr/discussions/2318#discussioncomment-7686542

### DIFF
--- a/_internal/src/index.ts
+++ b/_internal/src/index.ts
@@ -1,8 +1,9 @@
 import SWRConfig from './utils/config-context'
 import * as revalidateEvents from './events'
 import { INFINITE_PREFIX } from './constants'
+import SWRNoCacheOnMount from './utils/nocache-context'
 
-export { SWRConfig, revalidateEvents, INFINITE_PREFIX }
+export { SWRConfig, revalidateEvents, INFINITE_PREFIX, SWRNoCacheOnMount }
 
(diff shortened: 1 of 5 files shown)


benevbright:
I've gone with workaround: https://github.com/vercel/swr/discussions/2318#discussioncomment-7686542

PR #2805, opened 2023-09-27
Improve IDE import IntelliSense

A7med3bdulBaset:
This PR exports the hooks as named exports as well as default exports to make VS code import easier. 

the current behavior is when writing e.g. `const {data} = useSWR` VS code doesn't suggest the default hook, but others like `useSWRConfig` (named)

I see that this issue has already been discussed and accepted https://github.com/vercel/swr/pull/56 but after some updates, the issue occurred again.

### DIFF
--- a/core/src/index.ts
+++ b/core/src/index.ts
@@ -3,6 +3,7 @@ import 'client-only'
 // useSWR
 import useSWR from './use-swr'
 export default useSWR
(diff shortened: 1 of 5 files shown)


koba04:
Thank you! But we don't have a plan to have named exports for hooks.
refs. #1889

Issue #1590, opened 2021-10-25
infinite: do not revalidate every page on cursor-based pagination


hazae41:
# Bug report

## Description / Observed Behavior

When I use .mutate(), all pages are revalidated even tough nothing changed in the first page, and the keys didn't change

This behaviour happens regardless of `revalidateIfStale`, and regardless of the new `revalidateFirstPage`

## Expected Behavior

Do not revalidate if the first page is the same and keys are the same

## Repro Steps / Code Example

```typescript
const key = useCallback((page: number, previous: Data | null) => {
  // First page
  if (page === 0 || !previous)
    return h(`topics`) // h is just a query builder
  // No next page
  if (!previous.after)
    return null
  // Next page
  const [after] = previous.after
  return h(`topics`, { after })
}, [])

const res = useSWRInfinite<Data>(key, fetcher, {})

const refresh = useCallback(() => {
  res.mutate()
}, [res])

return <button onClick={refresh}>
  refresh
</button>
```

## Additional Context

happening on both swr@1.0.1 and swr@1.1.0-beta.6

hazae41:
Still happens on beta 8, all pages are revalidated when doing `mutate(undefined)`

```typescript
const refresh = useCallback(() => {
  res.mutate(d => d, true)
}, [res.mutate])
```

shuding:
Very weird, let me try to reproduce this.

shuding:
If you call `res.mutate()`, everything will be revalidated. That's currently the expected behavior:

https://github.com/vercel/swr/blob/e1677522d86017ec84099f90be03ec9607baf34b/infinite/index.ts#L190-L191

We need to improve the local mutation API for pages a bit.

iDarkLightning:
Hi!

Have there been any updates regarding this issue? I'd like to be able to mutate & revalidate a specified page, and not all the loaded pages when calling `mutate()`. Is this possible in any way with the current implementation?

misa-minic:
Hi!
Any updates regarding this issue?

hazae41:
If you find this annoying, you can use my library

https://github.com/hazae41/xswr 

You can revalidate any single page without revalidating everything

misa-minic:
I have to use SWR only, but I found some work around.

khanhld1910:
hi @misa-minic could you share your solution? I thought about storing all items into a `useState` and then modifying a specific item after it's updated but I feel there should be a better solution

PR #2862, opened 2023-12-22
feat: pass a function to the revalidate option in mutate

koba04:
This PR proposes to allow us to pass a function to the `revalidate` option.

## Motivation

My main motivation is to provide a way to revalidate specific pages with `useSWRInfinite`. See more details in the section below.
refs. https://github.com/vercel/swr/issues/1590#issuecomment-1000447513, https://github.com/vercel/swr/discussions/616#discussioncomment-81971

This also allows us a conditional revalidation based on the result of the mutation.

I can implement this feature as a separate option, but I think having `revalidate` and `revalidatePage` (?) is confusing. This proposal only adds a way to pass a function to the `revalidate` option, but this gives us more capabilities.

## Interface

```diff
- revalidate?: boolean
+ revalidate?: boolean | (data: Data, key: Arguments) => boolean
```

`data` is the latest data, which means this is the mutated data if `mutate` has the `data` argument.
`key` is the `key` value bounded to the mutation. If `fetcher` accepts a function as its `key` argument, this is the return value of the function.

## with `useSWRInfinite`

If you use `mutate` returned from `useSWRInfinite`, the `revalidate` function is called with each page and `useSWRInfinite` only revalidates the pages that return `true`, which means you can revalidate specific pages. This drastically improves performance in the case we know what pages we should revalidate.

```jsx
const { mutate, size } = useSWRInfinite((index) => [`/api/?page=${index + 1}`, index + 1], fetcher);

mutate({
  // only revalidate the last page
  revalidate: (data, [_, page]) => page === size
});
```

### DIFF
--- a/src/infinite/index.ts
+++ b/src/infinite/index.ts
@@ -21,7 +21,6 @@ import type {
   SWRHook,
   MutatorCallback,
   Middleware,
-  MutatorOptions,
   GlobalState
 } from '../_internal'
 import type {
@@ -31,7 +30,8 @@ import type {
   SWRInfiniteKeyLoader,
   SWRInfiniteFetcher,
   SWRInfiniteCacheValue,
-  SWRInfiniteCompareFn
+  SWRInfiniteCompareFn,
+  SWRInfiniteMutatorOptions
 } from './types'
 import { useSyncExternalStore } from 'use-sync-external-store/shim/index.js'
 import { getFirstPageKey } from './serialize'
@@ -55,7 +55,7 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
     fn: BareFetcher<Data> | null,
     config: Omit<typeof SWRConfig.defaultValue, 'fetcher'> &
       Omit<SWRInfiniteConfiguration<Data, Error>, 'fetcher'>
-  ): SWRInfiniteResponse<Data, Error> => {
+  ) => {
     const didMountRef = useRef<boolean>(false)
     const {
       cache,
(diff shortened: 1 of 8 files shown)


src/infinite/index.ts
```diff
@@ -187,7 +189,12 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
             (cacheData &&
               !isUndefined(cacheData[i]) &&
               !config.compare(cacheData[i], pageData))
-          if (fn && shouldFetchPage) {
+          if (
+            fn &&
+            (typeof shouldRevalidatePage === 'function'
```
huozhi, line 194: should this `shouldRevalidatePage()` result combined with `shouldFetchPage`? looks like if `shouldFetchPage` is false but `shouldRevalidatePage()` returns true we'll still revalidate

koba04: @huozhi 
Thank you for your feedback!
That's a good point👍 I think that's a technical decision.

The current implementation uses the result of the `revalidate` function and ignores other conditions if it's defined. This is based on the current `revalidate` option, which means the current `useSWRInfinite` with `revalidate: false` never revalidate.

Do you think we should revalidate in some scenarios even if the `revalidate` function returns `false`?

```js
const shouldFetchPage =
  shouldRevalidatePage(pageData, pageArg) || // if false, the following conditions are tested
  revalidateAll ||
  forceRevalidateAll ||
  isUndefined(pageData) ||
  (revalidateFirstPage && !i && !isUndefined(cacheData)) ||
  shouldRevalidateOnMount ||
  (cacheData &&
    !isUndefined(cacheData[i]) &&
    !config.compare(cacheData[i], pageData)
```

If we combine `shouldRevalidatePage` with `shouldFetchPage`, this makes it difficult to disable revalidation for specific pages because revalidation might occur even if `shouldRevalidatePage` returns false.

shuding (commented):
The code and the feature itself sound good to me 👍

huozhi (approved):

PR #2865, opened 2023-12-29
build: use new bundle convention

huozhi:
Upgrade to mordern build convention of bunchee, in this way we can build the package easier. This is a "two birds one stone" PR, explaining in the following.

### Module Resolution

Previously they're separated into different folders with sub folder package.json, but for `exports` field they're not necessary. `exports` field will only pick up from root `package.json`. 

In the new forms of the sub folders, comparing among different module resolution of TypeScript:

| module resolution | solution |
|:-- |:-- |
| node 10 | use `main`/`module` fields of all package.json, including sub packages |
| node 16 | use `import` condition of `exports` in root package.json |
| dual package | use `import` or `require` imports condition in root package.json |

### Reorganize Source Files

* All sources are in the `src/` folder
* All bundles are output into `dist/` foler
* Use relative imports without any typescript path alias configuration
* Configure the different types for each export condition (`import` / `require` / `module`)
* Keep the `<folder>/package.json` pointing to `dist/` files for node 10 module resolution compatibility

## TL;DR

The source files are formed into the below structure

```
- src/
  | - index/
  | - _internal/
  | - <other folders...>/
```

[arethetypeswrong CLI](https://arethetypeswrong.github.io) testing result:

```
swr v2.2.4

Build tools:
- @arethetypeswrong/cli@^0.7.0
- typescript@5.1.3

 No problems found 🌟


┌────────────────────┬───────────┬───────────────────┬───────────────────┬───────────┐
│                    │ node10    │ node16 (from CJS) │ node16 (from ESM) │ bundler   │
├────────────────────┼───────────┼───────────────────┼───────────────────┼───────────┤
│ "swr/package.json" │ 🟢 (JSON) │ 🟢 (JSON)         │ 🟢 (JSON)         │ 🟢 (JSON) │
├────────────────────┼───────────┼───────────────────┼───────────────────┼───────────┤
│ "swr"              │ 🟢        │ 🟢 (CJS)          │ 🟢 (ESM)          │ 🟢        │
├────────────────────┼───────────┼───────────────────┼───────────────────┼───────────┤
│ "swr/infinite"     │ 🟢        │ 🟢 (CJS)          │ 🟢 (ESM)          │ 🟢        │
├────────────────────┼───────────┼───────────────────┼───────────────────┼───────────┤
│ "swr/immutable"    │ 🟢        │ 🟢 (CJS)          │ 🟢 (ESM)          │ 🟢        │
├────────────────────┼───────────┼───────────────────┼───────────────────┼───────────┤
│ "swr/subscription" │ 🟢        │ 🟢 (CJS)          │ 🟢 (ESM)          │ 🟢        │
├────────────────────┼───────────┼───────────────────┼───────────────────┼───────────┤
│ "swr/mutation"     │ 🟢        │ 🟢 (CJS)          │ 🟢 (ESM)          │ 🟢        │
├────────────────────┼───────────┼───────────────────┼───────────────────┼───────────┤
│ "swr/_internal"    │ 🟢        │ 🟢 (CJS)          │ 🟢 (ESM)          │ 🟢        │
└────────────────────┴───────────┴───────────────────┴───────────────────┴───────────┘
```


core/package.json
```diff
@@ -3,7 +3,6 @@
   "module": "./dist/index.esm.js",
   "types": "./dist/index.d.ts",
   "exports": {
-    "react-server": "./dist/react-server.mjs",
```
shuding, line 6: Will this cause issues when using inside Next.js? How will Webpack handle it on the server layer?

huozhi: This file can be removed, the `core/package.json` is root package.json now. webpack will follow the standard to read the `exports` in `package.json`. Removed in [5b9c0fc](https://github.com/vercel/swr/pull/2865/commits/5b9c0fc247f668db41eeecff5297cc4a076132fa)

package.json
```diff
@@ -11,74 +11,88 @@
...
-      ],
-      "subscription": [
-        "./subscription/dist/index.d.ts"
-      ]
-    }
-  },
+  "main": "./dist/core/index.js",
+  "module": "./dist/core/index.esm.js",
+  "types": "./dist/core/index.d.ts",
   "exports": {
```
koba04, line 17: We now have all exports in here, so could we remove all `package.json` files in subdirectories like `_internal/`, `immutable/`, `infinite/`, `mutation/` and `subscription/`?

jest.config.js
```diff
@@ -5,12 +5,12 @@ module.exports = {
   modulePathIgnorePatterns: ['<rootDir>/examples/'],
   setupFilesAfterEnv: ['<rootDir>/test/jest-setup.ts'],
   moduleNameMapper: {
-    '^swr$': '<rootDir>/core/src/index.ts',
-    '^swr/infinite$': '<rootDir>/infinite/src/index.ts',
-    '^swr/immutable$': '<rootDir>/immutable/src/index.ts',
-    '^swr/subscription$': '<rootDir>/subscription/src/index.ts',
-    '^swr/mutation$': '<rootDir>/mutation/src/index.ts',
-    '^swr/_internal$': '<rootDir>/_internal/src/index.ts'
+    '^swr$': '<rootDir>/src/index.ts',
```
koba04, line 8: This file doesn't exist.

```suggestion
    '^swr$': '<rootDir>/src/core/index.ts',
```

src/index/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from '../core'
```
koba04, line 1: Is the `/index/` directory necessary? This is not pointed at anywhere.

package.json
huozhi: We kept them for `node10` resolution. can deprecate maybe in the next major version.

src/index/index.ts
koba04: Thank you!
I don't fully understand how `bunchee` works, so I thought this was unnecessary because `package.json` points `core/*` files in the `exports` field. But this is necessary to generate `dist/core/*` files.
When  I removed `/index/`, the build files weren't created.

src/index/index.ts
```diff
@@ -0,0 +1,2 @@
...
+export { default } from '../core'
```
koba04, line 2: @huozhi Why is it necessary? This adds the named export for `useSWR`, doesn't it?

huozhi: Not sure you mean the file or the re-export of default.

For the index file, `./index/` matches `.` export, this is the name convention for bunchee, also acts like nodejs resolve behavior.

For the export, `default` need to export separatedly. we didn't export `useSWR` as named export in `core` or `index`

koba04: Thank you! I misunderstood it

jest.config.js
```diff
@@ -5,12 +5,12 @@ module.exports = {
   modulePathIgnorePatterns: ['<rootDir>/examples/'],
   setupFilesAfterEnv: ['<rootDir>/test/jest-setup.ts'],
   moduleNameMapper: {
-    '^swr$': '<rootDir>/core/src/index.ts',
-    '^swr/infinite$': '<rootDir>/infinite/src/index.ts',
-    '^swr/immutable$': '<rootDir>/immutable/src/index.ts',
-    '^swr/subscription$': '<rootDir>/subscription/src/index.ts',
-    '^swr/mutation$': '<rootDir>/mutation/src/index.ts',
-    '^swr/_internal$': '<rootDir>/_internal/src/index.ts'
+    '^swr$': '<rootDir>/index/index.ts',
```
koba04, line 8: `pnpm test` is failing. I've made a PR to run the command on CI #2872

```suggestion
    '^swr$': '<rootDir>/src/index/index.ts',
```

huozhi: Good catch!

koba04 (approved):
👍 LGTM!!! 👏👏👏

PR #2882, opened 2024-01-25
fix: SWRConfiguration type

### LINKED ISSUE #2881: Type mismatch on SWR Configuration
The properties in the `react-native` section seem no longer compatible with the `SWRConfiguration` type.

In the docs you have the example of 

```tsx
<SWRConfig
  value={{
    provider: () => new Map(),
    isOnline() {
      /* Customize the network state detector */
      return true
    },
    isVisible() {
      /* Customize the visibility state detector */
      return true
    },
    initFocus(callback) {
      /* Register the listener with your state provider */
    },
    initReconnect(callback) {
      /* Register the listener with your state provider */
    }
  }}
>
```

This however misaligns with the exported type of `SWRConfiguration` as the following properties are missing

- `provider`
- `initFocus`
- `initReconnect`

Upon further inspection of the exported `index.d.ts`, the missing properties come from type `Config` which is however not exported.

```ts
type Config = SWRConfiguration & Partial<ProviderConfiguration> & {
    provider?: (cache: Readonly<Cache>) => Cache;
};
```

where `ProviderConfiguration` is defined as

```ts
type ProviderConfiguration = {
    initFocus: (callback: () => void) => (() => void) | void;
    initReconnect: (callback: () => void) => (() => void) | void;
};
```
Unless there's an internal reason for not exposing these methods _(which seems contradictory with the `react-native` documentation)_ the exported type for `SWRConfig` should be `Config` instead

Furthermore when importing the component `SWRConfig` it just defaults to `:any`, seems a resolution issue with directly importing react over just importing it's types in the SWRConfig definition. `import ReactExports from 'react'`

hulla-dev:
## PR description 📖 
Fixes the type mismatch in `SWRConfiguration` with the non-exported internal `Config` type with the FC `SWRConfig` export. This was causing the following issues:

- Custom configs could not be typed with `:SWRConfiguration`
- The internal `Config` type was used but not exposted
- Aligns the documentation with `TypeScript` section and `react-native` section, as combining these two approaches produce a type error right now.

## Issue reference 🔗 

- Closes #2881

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
     "build": "bunchee",
     "build:e2e": "pnpm next build e2e/site",
     "attw": "attw --pack",
-    "types:check": "pnpm -r run types:check",
+    "types:check": "tsc --noEmit",
     "prepublishOnly": "pnpm clean && pnpm build",
     "publish-beta": "pnpm publish --tag beta",
     "format": "prettier --write ./**/*.{ts,tsx}",
--- a/src/_internal/types.ts
+++ b/src/_internal/types.ts
@@ -403,7 +403,10 @@ export type SWRConfiguration<
   Data = any,
   Error = any,
   Fn extends BareFetcher<any> = BareFetcher<any>
-> = Partial<PublicConfiguration<Data, Error, Fn>>
+> = Partial<PublicConfiguration<Data, Error, Fn>> &
+  Partial<ProviderConfiguration> & {
+    provider?: (cache: Readonly<Cache>) => Cache
+  }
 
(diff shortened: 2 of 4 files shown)


huozhi:
Thanks for the PR, can we add a type test for this case?

hulla-dev:
> Thanks for the PR, can we add a type test for this case?

Will do later today / tomorrow. Got some work stuff to take care of first.

huozhi (approved):
I added a test for it, looks good now, thanks!

PR #2891, opened 2024-02-16
[Experimental] Support promises as fallback data

shuding:
This PR introduces the ability to accept promises as fallback data for SWR read hooks. The behavior is that the resolved value of the promise will be used as fallback instead of the promise itself.

The resolution of the promise is implemented via `use` which means that it will suspend the upper boundary.


promer94 (commented):
I think we should avoid upgrading react to a specific version here?

we currently have a spearte testing workflow for react@canary at https://github.com/vercel/swr/actions/workflows/test-canary.yml

so i guess we could create a sperate jest config for this and only test it in react canary ?

src/index/index.react-server.ts
```diff
@@ -1 +1,2 @@
 export { unstable_serialize } from '../core/serialize'
+export { SWRConfig } from './config'
```
promer94, line 2: Should we expose it to server components ?

Since <SWRConfig /> can take function as props.

huozhi: yeah we decide to expose it now, cannot pass down function is like a general issue now that hard to get rid of

shuding: Users can create a new client boundary to work around this, so might be a special case / documentation issue.

src/core/use-swr.ts
```diff
@@ -7,6 +7,8 @@ import ReactExports, {
 } from 'react'
 import { useSyncExternalStore } from 'use-sync-external-store/shim/index.js'
 
+console.log('ReactExports.version', ReactExports.version)
+
```
shuding: Leftover console.log.

test/use-swr-promise.test.tsx
```diff
@@ -4,6 +4,11 @@ import { createKey, createResponse, renderWithConfig, sleep } from './utils'
 import { Suspense } from 'react'
 
 describe('useSWR - promise', () => {
+  if (!process.env.REACT_CANARY) {
+    it('skip test non version of react', () => {})
```
shuding: I guess we want to test all versions, or maybe >= 18.2. Not just canaries.

huozhi (approved):

vpontis:
@shuding I think this broke something with our types.

We modify the type of `data`  when creating our hook based on `useSWR` so that we know `data` is defined if you specify `fallbackData`. This is a pretty nice type improvement. 

```tsx
  // We can be confident that data will be nonnull even if the request fails,
  // if we defined fallbackData in the config.
  data: SWRConfig extends { fallbackData: Response }
    ? Response
    : Response | undefined;
```

This is part of the bigger hook:

```tsx
import { AxiosError } from "axios";
import { useCallback } from "react";
import useSWR, { KeyedMutator, SWRConfiguration } from "swr";

export type Refresh = () => Promise<unknown>;

type GetApi = any;

export const useZmClient = <
  Path extends keyof GetApi,
  Request extends GetApi[Path]["request"],
  Response extends GetApi[Path]["response"],
  SWRConfig extends SWRConfiguration<Response>,
>({
  path,
  pause,
  args,
  swrConfig,
  additionalKey,
}: {
  path: Path;
  pause?: boolean;
  args?: Request;
  swrConfig?: SWRConfig;
  additionalKey?: string;
}): {
  // We can be confident that data will be nonnull even if the request fails,
  // if we defined fallbackData in the config.
  data: SWRConfig extends { fallbackData: Response }
    ? Response
    : Response | undefined;
  error: AxiosError<{ code?: string }> | undefined;
  mutate: KeyedMutator<Response>;
  refresh: Refresh;
} => {
  const swrKey = [path, JSON.stringify(args)];
  if (additionalKey) {
    swrKey.push(additionalKey);
  }

  const { data, error, mutate } = useSWR(
    pause ? null : swrKey,
    () => null as any,
    swrConfig,
  );
  const refresh = useCallback(async () => {
    await mutate();
  }, [mutate]);

  return { data: data!, error, mutate, refresh };
};
```

This will now give a type error:

```
components/hooks/useZmClient.tsx:53:5 - error TS2769: No overload matches this call.
  Overload 1 of 12, '(key: string[] | null, fetcher: ((arg: string[]) => FetcherResponse<Response>) | null, config: SWRConfigurationWithOptionalFallback<SWRConfig> | undefined): SWRResponse<...>', gave the following error.
    Argument of type 'SWRConfig | undefined' is not assignable to parameter of type 'SWRConfigurationWithOptionalFallback<SWRConfig> | undefined'.
      Type 'SWRConfig' is not assignable to type 'SWRConfigurationWithOptionalFallback<SWRConfig> | undefined'.
        Type 'SWRConfiguration<Response>' is not assignable to type 'SWRConfigurationWithOptionalFallback<SWRConfig> | undefined'.
          Type 'SWRConfiguration<Response>' is not assignable to type 'SWRConfigurationWithOptionalFallback<SWRConfig>'.
  Overload 2 of 12, '(key: Key, fetcher: BareFetcher<Response> | null, config: SWRConfigurationWithOptionalFallback<SWRConfig> | undefined): SWRResponse<...>', gave the following error.
    Argument of type 'SWRConfig | undefined' is not assignable to parameter of type 'SWRConfigurationWithOptionalFallback<SWRConfig> | undefined'.
      Type 'SWRConfig' is not assignable to type 'SWRConfigurationWithOptionalFallback<SWRConfig> | undefined'.
        Type 'SWRConfiguration<Response>' is not assignable to type 'SWRConfigurationWithOptionalFallback<SWRConfig> | undefined'.
          Type 'SWRConfiguration<Response>' is not assignable to type 'SWRConfigurationWithOptionalFallback<SWRConfig>'.

53     swrConfig,
       ~~~~~~~~~
```

I think it would be nice to either:

1. update SWR types to be smart about the `data` type based on if `fallbackData` is an object
2. update the configuration types a bit to support what I'm trying to do above

Or maybe I'm doing something else wrong. If you'd like me to open an issue or try to fix something, please let me know!

PR #2897, opened 2024-02-17
build: simplify react-server export and update bundler

huozhi:
* Simplify the export and mark swr context with "use client"
* Upgrade bundler

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,7 +140,7 @@
     "@types/use-sync-external-store": "^0.0.3",
     "@typescript-eslint/eslint-plugin": "5.59.8",
     "@typescript-eslint/parser": "5.59.8",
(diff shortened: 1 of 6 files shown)


shuding (changes requested):
This approach now bundles the code for `SWRConfig`, whereas the current main branch re-exports `SWRConfig` from `swr/_internal`. So this change kinda "clones" the module instead of sharing the same instance.

We need to ensure that 1️⃣ and 2️⃣ are the same instance:

- server layer, import SWRConfig
  - enters "use client"
    - exports SWRConfig // ← 1️⃣

- client layer, import SWRConfig
  - exports SWRConfig // ← 2️⃣

shuding (approved):

PR #2913, opened 2024-03-15
fix: check if config.fallback is undefined

### LINKED ISSUE #2912: Type error when fallback option with undefined.
# Bug report

An error will occur if the fallback option for useSWR is undefined.

```ts
const { data, error } = useSWR("/api/hello", {
  fallback: undefined
});
```

fallback option can be specified as undefined in the type definition.　　
The above code causes the following error.
```
TypeError: Cannot read properties of undefined (reading '/api/hello')
```

## Expected Behavior
I think fallback option should check if it is undefined.
(For example, [here](https://github.com/vercel/swr/blob/b103154bd0003861933d12d1de286d31acd2104d/src/core/use-swr.ts#L147))

## Repro Steps / Code Example

[CodeSandbox](https://codesandbox.io/p/devbox/swr-fallback-option-error-3rxpk8?file=%2Fapp%2Fpage.tsx%3A5%2C37)

## Additional Context

SWR version.  2.2.5

taku-hatano:
# PR description

Check if config.fallback is undefined.
If config.fallback is undefined, don't access to config.fallback[key].

# Issue reference
- Closes #2912

### DIFF
--- a/src/core/use-swr.ts
+++ b/src/core/use-swr.ts
@@ -144,7 +144,11 @@ export const useSWRHandler = <Data = any, Error = any>(
 
   // Resolve the fallback data from either the inline option, or the global provider.
   // If it's a promise, we simply let React suspend and resolve it for us.
-  let fallback = isUndefined(fallbackData) ? config.fallback[key] : fallbackData
+  let fallback = isUndefined(fallbackData)
+    ? isUndefined(config.fallback)
+      ? undefined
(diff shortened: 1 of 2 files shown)


promer94:
Could you add a test for this pr ?

taku-hatano:
@promer94 
Thank you for your review.
I added test!

promer94 (approved):

PR #2915, opened 2024-03-18
fix: Replace the deprecated 'window' with 'globalThis' for Deno

saul-atomrigs:
The `window` object became deprecated in Deno (https://github.com/denoland/deno/pull/22057). Instead, they recommend us to use `globalThis` or `self`. For the change, I also referred to here (https://deno.land/x/udibo_react_app@0.3.3/env.ts?source=#L7)


huozhi (approved):

huozhi:
Thanks! good to hear `window` is finally gone soon lol

DominicBao:
globalThis is not defined：globalThis 在一些低版本的浏览器上并不兼容

PR #2918, opened 2024-03-22
test: remove console.error times check

promer94:
Those checks could be safely removed since it's just implementation detail.


.github/workflows/test-canary.yml
```diff
@@ -19,8 +19,6 @@ jobs:
         run: corepack pnpm upgrade react@canary react-dom@canary use-sync-external-store@canary
 
       - name: Lint and test
-        env:
-          TEST_REACT_LEGACY: 1
```
promer94, line 23: I think we should not enable lagacy root in canary test ?

Issue #2877, opened 2024-01-21
useSWR deduped by mistake after cache clear


BertrandBordage:
# Bug report

## Description / Observed Behavior

This happens when you clear the cache, as described at the end of: https://swr.vercel.app/docs/mutation#mutate-multiple-items
If a `useSWR` call was done shortly before that cache clear, then this call will be skipped if it happens shortly after the cache clear, and the returned data will be `undefined`.
This issue happens only when the delay between the first call and the second call is smaller than `dedupingInterval` (2 s by default).

Setting `dedupingInterval` to 0 works around the issue, but disabling dedupes is not always desirable.

## Expected Behavior

All `useSWR` requests called for the first time since a cache clear must run and not be deduped.

The problem comes from the fact that the documentation calls this "clear all cache data":
```typescript
mutate(() => true, undefined, { revalidate: false })
```
when in practice it’s not exactly a cache clear, as it does not remove any cache key, but sets them to `undefined` instead.

The proper solution would probably be to have an official way of clearing the cache.

## Repro Steps / Code Example

Full example: https://codesandbox.io/p/sandbox/useswr-deduped-after-cache-clear-qkrwmj

https://github.com/vercel/swr/assets/1119169/a2e190e6-4070-4d78-b040-faa7e3591fc1

## Additional Context

SWR version 2.2.4

wavded:
Ran into same issue. In my case, user logs out, clear cache (per documentation), and logs in before the dedup interval, gets no data.

shhonarmandi:
The SWR cache utilizes JavaScript's built-in Map object for data storage. To clear the entire cache, you can directly call the clear method on the cache object retrieved using useSWRConfig.

```
const { cache } = useSWRConfig();
cache.clear();
```

Important Note:
Currently, there are no official TypeScript definitions for the clear method on the cache object or the JavaScript Map object itself. However, a pull request addressing this issue can be found here:
[https://github.com/vercel/swr/pull/2950/files](https://github.com/vercel/swr/pull/2950/files)

PR #2950, opened 2024-04-25
feat: add clear method to cache object type

shhonarmandi:
This commit updates the type definition of the Cache object to align with the native JavaScript Map data structure. This offers users access to clear method available on Map.

cache.clear(): Removes all key-value pairs from the cache (useful scenarios include user logout).

Relevant issue:
[https://github.com/vercel/swr/issues/2877](https://github.com/vercel/swr/issues/2877)


koba04:
Thank you for your PR!
As you mentioned, although the default Cache object is implemented with `Map` internally, it doesn't mean the Cache must be `Map.` We intentionally restrict the Cache interface to only APIs SWR uses internally so that users can implement their own Cache store without relying on `Map`.

Let me consider the clearing cache feature separately.

shhonarmandi:
@koba04 

Thank you for your feedback.

I updated the pull request and removed unnecessary methods and properties that you mentioned.

PR #2954, opened 2024-04-30
Add SWRInfiniteMutatorOptions type to export

ludwigbacklund:
When using `"moduleResolution": "Bundler"` in tsconfig.json for a TypeScript project, and implicitly referencing `SWRInfiniteMutatorOptions` you get this error:

```bash
Exported variable '<redacted>' has or is using name 'SWRInfiniteMutatorOptions' from external module "/path/node_modules/swr/dist/infinite/index" but cannot be named.ts(4023)
```

These errors can be resolved by explicitly exporting these implicitly references types from `swr`.

### DIFF
--- a/src/infinite/index.ts
+++ b/src/infinite/index.ts
@@ -358,5 +358,6 @@ export {
   SWRInfiniteHook,
   SWRInfiniteKeyLoader,
   SWRInfiniteFetcher,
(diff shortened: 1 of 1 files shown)


koba04 (approved):
👍 LGTM, thank you!

PR #3027, opened 2024-10-18
Improve-Type-Safety-and-State-Access-in-useStateWithDeps-Hook

O-BERNARDOFOEGBU:
This pull request aims to improve the type safety and reliability of the useStateWithDeps hook in React. Specifically, the changes address potential issues with accessing state properties using dynamic keys, ensuring that TypeScript correctly identifies and enforces the types of state properties.

Changes Implemented:
Type-Safe Iteration:

Replaced the direct for...in loop with a more type-safe iteration using Object.prototype.hasOwnProperty.call() to avoid errors related to iterating over unexpected properties.
Non-null Assertion:

Added a non-null assertion (payload[k]!) to ensure that payload[k] is not undefined when updating the state, preventing runtime issues.
Improved Type-Safe Access:

Used const k = key as keyof S to inform TypeScript that key is a valid key of S, improving the type inference and ensuring that currentState[k] and payload[k] are correctly typed.
Functional Consistency:

The changes maintain the core functionality of the hook while improving its safety and performance, ensuring that it tracks dependencies and triggers re-renders as intended.
This PR does not alter the primary behavior of the hook but makes the code more robust and type-safe, improving maintainability.

### DIFF
--- a/src/mutation/index.ts
+++ b/src/mutation/index.ts
@@ -38,17 +38,11 @@ const mutation = (<Data, Error>() =>
       error: Error | undefined
       isMutating: boolean
     }>({
-      data: undefined,
-      error: undefined,
+      data: UNDEFINED,
+      error: UNDEFINED,
       isMutating: false
-    }) //Modified the initial state setup so that data can be Data | undefined instead of just undefined
-    // allows data to hold both undefined initially and the resolved Data type after the fetcher runs.
+    })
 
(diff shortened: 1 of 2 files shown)


huozhi (approved):
Thanks!

O-BERNARDOFOEGBU:
> Thanks!

you are very welcome.

PR #3036, opened 2024-11-05
fix #3030 and run relateive test in edge-runtime

### LINKED ISSUE #3030: unstable_serialize doesn't work with NextJS edge runtime
# Bug report

When calling `unstable_serialize` from a server component that uses the NextJS **edge runtime** with an array or object key, the key is not serialized properly.

As an example, the code below logs `1~` to the console:

```javascript
export const runtime = 'edge'

export default function Home() {
  const key = ['test']
  console.log(unstable_serialize(key))

  return "Hello"
}
```

## Expected Behavior
I would expect the behavior in **edge runtime** to match the **NodeJS runtime**, with `@"test",` logged to the console (example code below)

```javascript
export default function Home() {
  const key = ['test']
  console.log(unstable_serialize(key))

  return "Hello"
}
```

### Debugging
I've done some digging, and the problem seems to originate from the `stableHash` function in the file: https://github.com/vercel/swr/blob/v2.2.5/src/_internal/utils/hash.ts.

- Line 40 (`if (constructor == Array) {`) and Line 48 (`if (constructor == OBJECT) {`) don't return true for arrays/objects when using the edge runtime.

Further testing reveals that although `key.constructor` doesn't equal `Array` in the **edge runtime**, `key instanceof Array` returns true (same applies to objects).

A possible solution could be to change:

- Line 40 to `if (arg instanceof Array) {`
- Line 48 to `if (arg instanceof OBJECT) {`

This works in both the **edge runtime** and **NodeJS runtime**, but I'm unsure of the historical reasons for not using `instanceof`. Additionally, I'm not sure why `arg.constructor` doesn't equal `Array` or `OBJECT` in the **edge runtime**.

### Package Versions
nextjs-swr-tests@0.1.0
├── @types/node@20.16.15
├── @types/react-dom@18.3.1
├── @types/react@18.3.12
├── eslint-config-next@14.2.16
├── eslint@8.57.1
├── next@14.2.16
├── react-dom@18.3.1
├── react@18.3.1
├── swr@2.2.5
└── typescript@5.6.3

## Repro Steps / Code Example
I've built a test repository that demonstrates the issue. To reproduce:

1. Clone the repo from https://github.com/william1616/nextjs-swr-edge-runtime-tests.
2. Install the packages (`npm i`), and then run the development server (`npm run dev`).
3. Open a browser and navigate to http://localhost:3000.
4. Check the console output:
    - `1~` (the output of `unstable_serialize`).
    - `true` (`key instanceof Array` is true).
    - `false` (`key.constructor == Array` is false).
5. In https://github.com/william1616/nextjs-swr-edge-runtime-tests/blob/main/src/app/page.tsx, comment out line 5 to change the runtime from edge to NodeJS.
6. Refresh http://localhost:3000 in your browser.
7. Check the console output:
    - `@"test",` (the output of `unstable_serialize`).
    - `true` (`key instanceof Array` is true).
    - `true` (`key.constructor == Array` is true).

promer94:
This pr intends to fix the inconsistent behavoir of swr's `hash` in edge-runtime dev mode.

The reason for this problem is because we're using `node.js` to simulate the `edge-runtime`  for local development 

The global `Array` and `Object` of edge-runtime are injected from `node.js` and the JS engine does not use global Array/Object constructor while building an array/object literal.

Then in dev mode
```
[].constructor !== Array.constructor
{}.construrtor !== Object.contructor
```

So we need find alternative to determine whether a key is an array or a `plain object`

### DIFF
--- a/package.json
+++ b/package.json
@@ -167,6 +167,7 @@
     "trailingComma": "none"
   },
   "dependencies": {
+    "@edge-runtime/jest-environment": "^3.0.4",
     "dequal": "^2.0.3",
     "use-sync-external-store": "^1.2.0"
   }
--- a/src/_internal/utils/hash.ts
+++ b/src/_internal/utils/hash.ts
@@ -19,13 +19,14 @@ let counter = 0
 // parsable.
 export const stableHash = (arg: any): string => {
   const type = typeof arg
-  const constructor = arg && arg.constructor
-  const isDate = constructor == Date
-
+  const objectTypeName = OBJECT.prototype.toString.call(arg).slice(8, -1)
+  const isDate = objectTypeName === 'Date'
+  const isRegex = objectTypeName === 'RegExp'
+  const isPlainObject = objectTypeName === 'Object'
   let result: any
   let index: any
 
-  if (OBJECT(arg) === arg && !isDate && constructor != RegExp) {
+  if (OBJECT(arg) === arg && !isDate && !isRegex) {
     // Object/function, not null/date/regexp. Use WeakMap to store the id first.
     // If it's already hashed, directly return the result.
     result = table.get(arg)
@@ -37,7 +38,7 @@ export const stableHash = (arg: any): string => {
     result = ++counter + '~'
     table.set(arg, result)
 
-    if (constructor == Array) {
+    if (Array.isArray(arg)) {
       // Array.
       result = '@'
       for (index = 0; index < arg.length; index++) {
@@ -45,7 +46,7 @@ export const stableHash = (arg: any): string => {
       }
       table.set(arg, result)
     }
-    if (constructor == OBJECT) {
+    if (isPlainObject) {
       // Object, sort keys.
       result = '#'
       const keys = OBJECT.keys(arg).sort()
--- a/src/_internal/utils/hash.ts
+++ b/src/_internal/utils/hash.ts
@@ -6,6 +6,9 @@ import { OBJECT, isUndefined } from './shared'
 // complexity is almost O(1).
 const table = new WeakMap<object, number | string>()
 
+const isObjectType = (value: any, type: string) =>
+  OBJECT.prototype.toString.call(value) === `[object ${type}]`
+
 // counter of the key
 let counter = 0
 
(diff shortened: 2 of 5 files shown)


test/unit/serialize.test.ts
```diff
@@ -1,3 +1,6 @@
+/**
+ * @jest-environment @edge-runtime/jest-environment
```
koba04, line 2: Should we test this on the jsdom environment as well as the edge environment?
(I'm not sure the difference between those environments, this might not be necessary)

promer94: From my perspective, `edge-runtime` is like a strict subset of `nodejs`, so if it works for `edge-runtime`, it should also work for `nodejs`.

I feel like it might not be necessary for now.

src/_internal/utils/hash.ts
```diff
@@ -19,13 +19,14 @@ let counter = 0
 // parsable.
 export const stableHash = (arg: any): string => {
   const type = typeof arg
-  const constructor = arg && arg.constructor
-  const isDate = constructor == Date
-
+  const objectTypeName = Object.prototype.toString.call(arg).slice(8, -1)
```
koba04, line 22:
```suggestion
  const objectTypeName = OBJECT.prototype.toString.call(arg).slice(8, -1)
```

src/_internal/utils/hash.ts
```diff
@@ -19,13 +19,14 @@ let counter = 0
 // parsable.
 export const stableHash = (arg: any): string => {
   const type = typeof arg
-  const constructor = arg && arg.constructor
-  const isDate = constructor == Date
-
+  const objectTypeName = OBJECT.prototype.toString.call(arg).slice(8, -1)
```
huozhi, line 22: it's not covering the rest part will match `[Object xxx]` or not

src/_internal/utils/hash.ts
```diff
@@ -37,15 +38,15 @@ export const stableHash = (arg: any): string => {
     result = ++counter + '~'
     table.set(arg, result)
 
-    if (constructor == Array) {
+    if (Array.isArray(arg)) {
```
huozhi, line 41: nice

huozhi (commented):
Can you add a PR description and some comments about the change in `hash`? It's important for others to understand the approach and grab the context when they read the code later unless directly by reading the code

package.json
```diff
@@ -167,7 +167,8 @@
     "trailingComma": "none"
   },
   "dependencies": {
+    "@edge-runtime/jest-environment": "^3.0.4",
```
huozhi, line 170: this should be dev dependency

package.json
```diff
@@ -167,7 +167,8 @@
...
     "dequal": "^2.0.3",
     "use-sync-external-store": "^1.2.0"
   }
-}
+}
```
huozhi, line 174: can we keep the trailign line break for each file?

huozhi (approved):

PR #3054, opened 2024-12-17
fix: missing `throwOnError` in SWRMutationHook options

Ram4GB:
Missing `throwOnError` in `SWRMutationHook`'s options.  Update code to let users pass option `throwOnError`.

Example:

```
useSWRMutation('/api/mutate-server-action', () => action(), { throwOnError: true })
```


huozhi (approved):

huozhi:
Types check is still failing, we can land it once it's pssed!

PR #4022, opened 2025-01-03
chore: update license year

maxktz:
Happy new year, love your project <3


phryneas:
Please stop these PRs. Best case it does nothing, worst case it invokes legal problems. The year should only change on licensing changes.
A user that opened 1200 of these PRs a few days ago was already banned from GitHub.

PR #4087, opened 2025-02-05
(fix) keepPreviousData: return fallback instead of undefined value

dvoytenko:
Fix for https://github.com/vercel/swr/pull/4084. Only `undefined` values should be avoided. `null` is ok.

### DIFF
--- a/src/index/use-swr.ts
+++ b/src/index/use-swr.ts
@@ -278,7 +278,9 @@ export const useSWRHandler = <Data = any, Error = any>(
 
   const returnedData = keepPreviousData
     ? isUndefined(cachedData)
-      ? laggyDataRef.current ?? data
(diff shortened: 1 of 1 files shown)


src/index/use-swr.ts
```diff
@@ -278,7 +278,9 @@ export const useSWRHandler = <Data = any, Error = any>(
 
   const returnedData = keepPreviousData
     ? isUndefined(cachedData)
-      ? laggyDataRef.current ?? data
+      ? isUndefined(laggyDataRef.current)
```
huozhi, line 281:
```suggestion
      // checking undefined to avoid null being fallback as well
      ? isUndefined(laggyDataRef.current)
```

huozhi (approved):

PR #4118, opened 2025-04-17
fix: check "if (!error)" skip error if value is cast to false

victorgrachev:
Hello!
"if (!error)" make cast value to boolean type and if value is cast to false then the error is skipped.

Sandbox: https://codesandbox.io/p/sandbox/6s9dw7

Example:

````javascript
import useSWRMutation from "swr/mutation";

const fetcher = () => {
  return new Promise((_, reject) => reject("")); // or null/undefined/false/0
};

const Component = () => {
  const { trigger } = useSWRMutation("/example", fetcher);

  const handleClick = () => {
    trigger()
      .then(() => {
        console.log("skip error");
      })
      .catch(() => {
        console.log("error");
      });
  };

  return <button onClick={handleClick}>Execute</button>;
};
```

### DIFF
--- a/src/_internal/utils/mutate.ts
+++ b/src/_internal/utils/mutate.ts
@@ -123,6 +123,7 @@ export async function internalMutate<Data>(
 
     let data: any = _data
     let error: unknown
+    let isError = false
 
     // Update global timestamps.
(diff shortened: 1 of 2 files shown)


victorgrachev:
Test ready

PR #4133, opened 2025-06-09
feat: add parallelLimit option to useSWRInfinite

FAL-coffee:
This PR adds a new parallelLimit option to useSWRInfinite. When used with the parallel option, it limits the number of concurrent fetches to the specified value, instead of fetching all pages in parallel at once. This can help control the load on the server and improve performance for applications that need to fetch many pages.

### DIFF
--- a/src/infinite/index.ts
+++ b/src/infinite/index.ts
@@ -56,7 +56,8 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
       persistSize = false,
       revalidateFirstPage = true,
       revalidateOnMount = false,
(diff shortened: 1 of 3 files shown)


src/infinite/index.ts
```diff
@@ -215,9 +216,22 @@ export const infinite = (<Data, Error>(useSWRNext: SWRHook) =>
           }
         }
 
-        // flush all revalidateions in parallel
+        // flush all revalidations in parallel with optional limit
```
FAL-coffee, line 219: typo 
revalidateions → revalidations

PR #4137, opened 2025-06-25
[wip] Add an e2e test page for testing the swr performance when many hooks are used.

gdborton:
I'm trying to showcase a performance issue in SWR. The code as is adds a test page that has two render paths, one renders 10k divs directly, the other renders 10k divs using the same swr hook.

The render path that uses swr is extremely slow to paint.

TODO:
 - [x] create the test page
 - [ ] Add a test that ensures lots of hooks isn't orders of magnitude slower than no hooks
 - [ ] Fix the failing test.

### DIFF
--- a/e2e/site/app/perf/page.tsx
+++ b/e2e/site/app/perf/page.tsx
@@ -1,0 +1,63 @@
+'use client'
+import { useState } from 'react'
+import useSWR from 'swr'
+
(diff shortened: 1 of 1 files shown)


gdborton:
This was fixed here:
https://github.com/vercel/swr/pull/4138

Issue #1670, opened 2021-11-25
global mutate can't revalidate useSWRInfinite data


williamwa:
# Bug report

## Description / Observed Behavior

What kind of issues did you encounter with SWR?

For example a Todo app, on the list page:

I using `useSWRInfinite` to load todos, the first page's key is `/api/todos`,

Then I delete a Todo (send request to backend to deleted it), and I need revalidate the data:

If I use `mutate` returned from the `useSWRInfinite` call, it will work.
If I use `mutate` from `const { mutate } = useSWRConfig()`, it will not work.

## Expected Behavior

How did you expect SWR to behave here?

Both way need way. because I may call the delete not on the same component, so I can't pass the mutate function to other component

## Additional Context

SWR version.  1.0.1

anothertempore:
Have you passed your key to global mutate? https://swr.vercel.app/docs/mutation#revalidate

williamwa:
> Have you passed your key to global mutate? https://swr.vercel.app/docs/mutation#revalidate

Thanks for reply, Yes, when using global mutate, I use `mutate('/api/todos')`.

williamwa:
I checked code, there is a `const INFINITE_PREFIX = '$inf$'` added to the key when using infinite, so I try use global mutate, and call, `mutate('$inf$/api/todos')`, and it work.

huozhi:
You could use `unstable_serialize` to get the keys of swr infinite

```js
import { unstable_serialize } from 'swr/infinite'

mutate(
   unstable_serialize((index, prevPageData) => `my-key-of-page-${index}`),
)
```

shuding:
We should add this to the docs, but there is currently a bug in Nextra blocking this...

ezeikel:
Not sure why this was closed? Would be great if this issue was still open and the docs got updated to reflect this

huozhi:
@ezeikel Good point! I opened https://github.com/vercel/swr-site/issues/264 for this doc requirement

embeddedt:
Can I ask what the reason is for the prefix? I'm using `useSWRInfinite` as a workaround for needing to make a dynamic number of `useSWR` calls and this difference has caused some bugs in my application.

jezmck:
I suggest this be reopened.

Conviley:
I'm using SWR version 2.0.0 and this problem still exists.
+1 on reopening this like @jezmck suggests

koba04:
[...]

PR #4167, opened 2025-08-25
fix: enable global mutate to revalidate useSWRInfinite data with includeSpecialKeys option

### LINKED ISSUE #4149: Global mutate with a matcher function does not receive keys from useSWRInfinite
When using a matcher function with the global `mutate` API from `useSWRConfig`, keys generated by `useSWRInfinite` hooks are not passed to the matcher. This prevents the global invalidation of all pages of an infinite query, which is a necessary pattern for actions like adding a new item to a collection that affects all pages.

The same matcher function correctly receives keys from standard `useSWR` hooks, but it completely ignores the keys associated with `useSWRInfinite` (which are prefixed with `$inf$`).

### Expected Behavior

I expect that a matcher function provided to the global `mutate(matcher, ...)` would be called with **all** active keys in the SWR cache. This should include the internal keys used by `useSWRInfinite`, allowing for global revalidation of infinite queries using a filter, like so:

```javascript
mutate((key) => /boards/.test(key), null);
```

### Observed Behavior

The matcher function is only ever called for keys associated with standard `useSWR` hooks. When logging the `key` argument inside the matcher, the keys managed by `useSWRInfinite` never appear, even though they are confirmed to be present in the cache.

As a result, attempting to globally mutate an infinite query with a regex or function matcher fails silently, as the relevant keys are never processed.

### Steps to Reproduce

1.  **Set up an infinite query:** Create a component (`Component A`) that uses the `useSWRInfinite` hook to fetch paginated data.

    ```javascript
    import useSWRInfinite from 'swr/infinite';

    const getKey = (pageIndex, previousPageData) => {
      if (previousPageData && !previousPageData.content.length) return null;
      return `/boards?page=${pageIndex}&perPage=25`;
    };

    const { data } = useSWRInfinite(getKey, fetcher);
    ```

2.  **Attempt a global mutation:** In a separate component (`Component B`) or global function, call `mutate` with a matcher function designed to target the infinite query keys.

    ```javascript
    // Component B
    import { useSWRConfig } from 'swr';

    const { mutate } = useSWRConfig();

    const revalidateBoards = () => {
      mutate(
        (key) => {
          // This log will NEVER show keys starting with "$inf$/boards..."
          console.log('Key received by matcher:', key);
          return /^\$inf\$\/boards/.test(key);
        },
        null, // Revalidate
        { revalidate: true }
      );
    };
    ```

3.  **Trigger the mutation:** Call the `revalidateBoards` function.

4.  **Observe the console:** Notice that the `console.log` inside the matcher is never executed for the keys belonging to `useSWRInfinite`. The infinite query is not revalidated.

joseph0926:
### Problem

After investigating this issue, I traced its origins back to [[Issue #1670](https://github.com/vercel/swr/issues/1670)](https://github.com/vercel/swr/issues/1670) where it was first reported. Following this, [[RFC #1946](https://github.com/vercel/swr/issues/1946)](https://github.com/vercel/swr/issues/1946) mentioned improvements that would address this problem, and the RFC discussion explicitly stated that "special keys would be excluded." This was then implemented in [[PR #1989](https://github.com/vercel/swr/pull/1989)](https://github.com/vercel/swr/pull/1989) where special keys were indeed excluded.

However, the problem is that global mutate ignores special keys - specifically the `$inf$` keys - causing revalidation failures for `useSWRInfinite` data.

### Investigation Process

To solve this, I added a new `includeSpecialKeys` flag to `MutatorOptions` to create an exception in the `// Skip the special useSWRInfinite and useSWRSubscription keys` conditional. This seemed promising, and I wrote several edge test cases to verify the solution.

However, I discovered an additional problem: `useSWRInfinite` sets an `_i` flag in the cache when calling its own mutate, which the fetcher checks to determine whether to refetch all pages. The issue was that global mutate had no way to set this flag (at least not within the scope of what a regular contributor could modify in a single PR).

While there are several potential solutions - like automatically setting the `_i` flag during fetcher traversal (though this makes responsibility boundaries unclear) or adding a new flag (which would mean adding two flags just for global mutate) - the final solution was to have `includeSpecialKeys` also handle setting the `_i` flag when true.

### Solution

This PR adds an `includeSpecialKeys` option to `MutatorOptions` that
1. Includes special keys (`$inf$`, `$sub$`) in the matcher function
2. Automatically sets the `_i` flag for `$inf$` keys to trigger full revalidation

### Usage
```typescript
mutate(
  key => key.includes('/api/todos'),
  undefined,
  { revalidate: true, includeSpecialKeys: true }
)
```

### Testing
Added comprehensive test cases covering:
- Basic revalidation of infinite queries
- Mixed `useSWR` and `useSWRInfinite` scenarios
- Performance with large cache
- Edge cases with empty matchers and concurrent mutations

---

7599612

- register cache-scoped infinite revalidators using the serialized first-page key
- trigger infinite revalidation alongside the matched page revalidator
- preserve original matcher arguments and keep `$inf$` internal keys hidden
- avoid writing page mutation data into the infinite aggregate cache
- clean up revalidators when infinite hooks unmount or change keys
- cover array keys, disabled revalidation, shared hooks, aggregate shape, and later-page-only matches

Fixes #4149


SuperstrongBE:
Please review and merge this !

joseph0926:
Hi @shuding  @huozhi  — just a gentle bump on this PR.

Happy to adjust the approach, split into smaller PRs, or follow any project preferences.  
If there’s anything blocking review, please let me know. Thanks!

ddoice:
Yup, I can confirm the bug, is a real pain having to delete the caches instead of just revalidate.

PR #4172, opened 2025-09-20
upgrade repo's deps and replace deprecated apis

promer94:


promer94:
Thanks for reviewing this large pr ❤️. 

The changes of this pr are interrelated so i did not create small PRs to update them individually. 

I understant you concerns about this PR. It touches too many files. But most of changes are just adding or removing comments. It won't be too hard to track or revert in the future. 

In future maintenance, we can use Copilot to periodically help us update the repository’s development dependencies, making subsequent upgrades smoother. But for now i guess we need to pay the previous tech debt.

PR #4197, opened 2025-12-10
Add default errorRetryCount of 5

ryohiy:
I apologize for creating this PR without waiting for feedback on the discussion.
This PR implements the change proposed in https://github.com/vercel/swr/discussions/4196.

I was planning to wait for a response in the discussion, but I decided to create this PR because this might be something that needs to be fixed soon, and the change itself is minimal.

I'll keep this as a Draft PR until there's movement in the discussion. There's no need to review this PR until the discussion progresses.

Sorry for the noise. Thank you.


shuding:
SWR uses exponential backoff for error retries (capped at 2^8) and 5s as the default interval. This means that it will retry after 5s, 10s, 20s, 40s, 80s, 160s, 320s, 640s, 1280s (and then fixed at 1280s). 1280s is 21 minutes, and this means that in 40+ minutes there will be only <= 9 retries. The cost is really trivial for a web application and it will be extremely rare to happen, as SWR also pauses it when the app is not focused.

PR #4212, opened 2026-01-25
fix: isHydration will cause unnecessary rerender 

### LINKED ISSUE #4210: Noticed useSWR rendering my component twice - even when not calling the fetcher
# Bug report

- Using SWR in an isomorphic application (Next.js + React)
- Regardless of conditionally calling the fetcher within useSWR, Im noticing my component render twice - even with reactstrictmode disabled

## Expected Behavior

1. I am simulating a fetch call on server side and passing this data down to an Age component
2. Inside the Age component, useSWR will conditionally call the fetcher if the server did not pass in data
3. As expected, the fetch call is not called on the client side - because the server data exists
4. However, Im noticing that my component is rendering twice ie. notice the chrome dev tools logs happen twice in the sandbox - **ive disabled react strict mode to remove re-render noise too**
5. All of the data returned from useSWR is still the same - so Im puzzled by the re-render
6. When I do fetch calls using useEffect instead, my component only renders once

## Repro Steps / Code Example

- https://codesandbox.io/p/devbox/nextjs-ssr-poc-r82x7z

#### Relevant files:
- `app/page.tsx` is the bulk of the page
- `app/components/age-v2.tsx` is the component that uses useSWR
- `app/utils/fetch-age.ts` is a dummy promise to simulate a fetch request with wait times
- `next.config.ts` Ive disabled reactstrictmode here to remove the bloated re-renders for debugging purposes

## Additional Context
- Is this a misunderstanding I have of SWR? ie. is this expected behavior and if it is, why?
- Ive also seen a similar issue but I dont think the answer considered the server side use case Im presenting: https://github.com/vercel/swr/issues/1952

SWR version. 

```
  "dependencies": {
    "next": "15.5.7",
    "react": "19.0.0",
    "react-dom": "19.0.0",
    "swr": "2.3.8"
  }
```

Image

promer94:
close #4210 

The current solution is inspired by seb's previous tweet (The original tweet was deleted) 
Screenshot 2026-01-25 at 16 36 39

### DIFF
--- a/e2e/site/app/render-count/page.tsx
+++ b/e2e/site/app/render-count/page.tsx
@@ -1,0 +1,8 @@
+'use client'
+import useSWR from 'swr'
+
+export default function Page() {
+  useSWR('swr should not cause extra rerenders')
+  console.count('render')
+  return <div>render count pages</div>
+}
--- a/e2e/site/app/server-prefetch-warning/page.tsx
+++ b/e2e/site/app/server-prefetch-warning/page.tsx
@@ -1,0 +1,41 @@
+'use client'
+
+import { useEffect, useState } from 'react'
+import useSWR, { SWRConfig } from 'swr'
+
+const fetcher = () => 'SWR'
+
+function Content() {
+  const [hydrated, setHydrated] = useState(false)
+  useEffect(() => {
+    setHydrated(true)
+  }, [])
+
+  useSWR('ssr:1', fetcher)
+  useSWR('ssr:2', fetcher)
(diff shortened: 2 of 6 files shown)


shuding:
This hack is crazy (and you have snapshotted tweets?!)

shuding (approved):

timtucker-dte:
Not sure what's happening, but in our client-side Next.js app we're seeing frequent failures to fetch data since this change.

PR #4224, opened 2026-02-23
fix: guard against undefined global state in createCacheHelper

### LINKED ISSUE #2986: useSWRMutation is not working with React Native New Architecture.
# Bug report

## Repro Steps / Code Example
1. npx create-expo-app@latest

configure app.json 

``` json 
 "plugins": [
      "expo-router",
      [
        "expo-build-properties",
        {
          "ios": {
            "newArchEnabled": true
          },
          "android": {
            "newArchEnabled": true
          }
        }
      ]
    ],
```

3. home.tsx code example

```tsx

function Home () {
  const {trigger, data, error} = useSWRMutation('https://github.com', (key) => fetch(key).then(res => res.text()))

  console.log({data, error}) // always print {"data": undefined, "error": undefined}

  function submit () {
    // print error: "(NOBRIDGE) ERROR  [TypeError: Cannot read property 'add' of undefined]"
    trigger().then(console.log).catch(console.error)
  }

  return  (
     <View>
       <Button title='mutation' onPress={submit}></Button>
     </View>
     )
}
```

veeceey:
When using SWR with React Native's New Architecture (bridgeless mode), `SWRGlobalState.get(cache)` can return `undefined` during early renders before the provider has fully initialized. This causes a crash when `createCacheHelper` tries to destructure `state[5]` and `state[6]` from the undefined value.

This adds a simple null guard:
- The setter becomes a no-op when state is undefined (the update will happen once the provider initializes properly)  
- The subscriber falls back to a function that returns a no-op unsubscribe handler

Tested locally - all existing unit and integration tests still pass.

Closes #2986

### DIFF
--- a/src/_internal/utils/helper.ts
+++ b/src/_internal/utils/helper.ts
@@ -1,6 +1,6 @@
 import type { Cache, State, GlobalState } from '../types'
 import { SWRGlobalState } from './global-state'
-import { isUndefined, mergeObjects } from './shared'
+import { isUndefined, mergeObjects, noop } from './shared'
 
 const EMPTY_CACHE = {}
 const INITIAL_CACHE: Record<string, any> = {}
@@ -19,7 +19,7 @@ export const createCacheHelper = <Data = any, T = State<Data, any>>(
   cache: Cache,
   key: string | undefined
 ) => {
-  const state = SWRGlobalState.get(cache) as GlobalState
+  const state = SWRGlobalState.get(cache) as GlobalState | undefined
   return [
     // Getter
     () => ((!isUndefined(key) && cache.get(key)) || EMPTY_CACHE) as T,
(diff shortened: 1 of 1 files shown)


veeceey:
### Test results

All existing tests pass:

```
$ pnpm test -- --testPathPattern "test/unit" --no-coverage
Test Suites: 3 passed, 3 total
Tests:       9 passed, 9 total

$ pnpm test -- --testPathPattern "test/use-swr-cache" --no-coverage
Test Suites: 1 passed, 1 total
Tests:       20 passed, 20 total

$ pnpm test -- --testPathPattern "test/use-swr-remote-mutation" --no-coverage
Test Suites: 1 passed, 1 total
Tests:       33 passed, 33 total
```

Type checks also pass (`tsc --noEmit` via pre-commit hook).

veeceey:
hi, just wondering if this could get a look. let me know if anything needs updating!

mahdirajaee (commented):
Good defensive fix for the React Native New Architecture edge case. Casting `SWRGlobalState.get(cache)` to `GlobalState | undefined` and then guarding before accessing `state[5]` and `state[6]` is the right approach — it prevents a runtime crash without altering the behavior for environments where state is always initialized. The fallback `() => noop` for the subscriber is a reasonable no-op, though it's worth noting this returns a function that returns a cleanup noop, so consumers calling the subscriber and then invoking the returned unsubscribe function will still work correctly without throwing. One thing to consider: if state is undefined during initial renders, the setter (second tuple element) will silently skip the `state[5]` broadcast. This means cache updates during that window are effectively lost. It might be worth adding a console warning in development mode if `state` is undefined so developers can diagnose initialization timing issues more easily rather than debugging silent data staleness.

PR #4232, opened 2026-03-18
fix: useSWRImmutable should disable revalidation on focus

### LINKED ISSUE #4225: Unexpected fetcher call on focus when using useSWRImmutable
# Bug report

When using `useSWRImmutable`, the hook should disable revalidation on focus. However, if you leave the app window and go back in (triggering a window focus event), the fetcher function is called again. This is confirmed by the `console.log("getItems called")` inside the fetcher firing every time the app regains focus.

## Expected Behavior

`useSWRImmutable` should implicitly set `revalidateIfStale`, `revalidateOnFocus`, and `revalidateOnReconnect` to false. Leaving the app and returning should not trigger the fetcher function if the cache is already populated and the SWR key has not changed.

## Repro Steps / Code Example

Link to reproduction: [CodeSandbox](https://codesandbox.io/p/sandbox/elegant-rain-78gkmg). Look inside the console! Don’t switch focus too quickly; otherwise, you won’t see it due to the deduping interval in SWR.

## Additional Context

SWR version: `2.4.0`
React version: `19.2.4`

## Update: Potential Workaround Found

I have discovered that explicitly setting the revalidation flags to false within the options object resolves the issue. Even though useSWRImmutable is supposed to handle this by default, adding these options explicitly ensures the hook works as expected:
`useSWRImmutable(key, fetcher, {
  revalidateOnFocus: false,
})`

buyuan-dev:
## Summary

Fixes a bug where useSWRImmutable was not properly disabling revalidation on window focus.

## Problem

useSWRImmutable is documented to disable all automatic revalidation, but when the window loses and regains focus, the fetcher was being called again.

## Root Cause

The immutable middleware was mutating the config object directly, which could be overridden by merged configs.

## Fix

Updated the immutable middleware to create a new config object with explicit overrides, ensuring all revalidation flags are properly disabled:
- 
evalidateOnFocus`n- 
evalidateIfStale`n- 
evalidateOnReconnect`n- 
efreshInterval`n
## Testing

- [x] All existing tests pass
- [x] Manual testing shows no revalidation on focus

Fixes #4225

### DIFF
--- a/src/immutable/index.ts
+++ b/src/immutable/index.ts
@@ -4,11 +4,13 @@ import { withMiddleware } from '../_internal'
 
 export const immutable: Middleware = useSWRNext => (key, fetcher, config) => {
   // Always override all revalidate options.
-  config.revalidateOnFocus = false
-  config.revalidateIfStale = false
-  config.revalidateOnReconnect = false
-  config.refreshInterval = 0
-  return useSWRNext(key, fetcher, config)
+  return useSWRNext(key, fetcher, {
+    ...config,
+    revalidateOnFocus: false,
+    revalidateIfStale: false,
+    revalidateOnReconnect: false,
+    refreshInterval: 0
+  })
 }
 
 const useSWRImmutable = withMiddleware(useSWR, immutable)


mahdirajaee (commented):
Clean fix. The root cause is clear — mutating `config` properties directly means the overrides were being applied to the shared config object rather than producing a new one for this specific hook invocation. Spreading into a new object with the overrides applied is the correct approach and prevents unintended side effects on other hooks sharing the same config reference. This is particularly important because if `useSWRImmutable` was used alongside regular `useSWR` in the same `SWRConfig` provider, the mutation of `config.revalidateOnFocus = false` could leak into sibling hooks. The new spread pattern cleanly avoids this. One minor observation: the trailing comma after `refreshInterval: 0` is missing, which is fine for the current formatting rules but worth noting if the project enforces consistent trailing commas via a linter.

PR #4244, opened 2026-04-13
fix: support async subscribe in useSWRSubscription

### LINKED ISSUE #4214: useSWRSubscription does not work with async subscribe function
# Bug report

useSWRSubscription accepts an async subscribe function as per TypeScript types but does not await it and therefore does not subscribe properly. In addition, if the subscribe function is marked as async, it cannot get returned unsubscribe function as it's wrapped in a Promise.

I encountered this while trying to connect to an AWS Amplify event source. See following code example below for an example.

## Expected Behavior
It should either await async subscribe function (which is preferred) or do not accept one.

## Repro Steps / Code Example

```typescript
import { events } from "aws-amplify/data";

const useSubscribeToEventSource = () => {
  return useSWRSubscription(key, async (_key, { next }) => {
    // .connect function is async
    const channel  = await events.connect("channel-name");
    const subscription = channel.subscribe({
      next: event => next(null, event),
      error: error => next(error, null)
    });

    return () => {
      try {
        subscription?.unsubscribe();
      } finally {
        channelRef.current?.close();
      }
    };
  });
};
```

## Additional Context

SWR version: 2.3.8

solssak:
useSWRSubscription does not support async subscribe functions. When the subscribe callback is async (e.g. to connect to AWS Amplify), the returned Promise is not handled, causing the typeof check to throw.

This PR adds support for async subscribe by detecting Promise returns and resolving them properly, including handling the race condition where the component unmounts before the async subscribe resolves.

Fixes #4214

**src/subscription/types.ts** — Subscribe return type now accepts Promise.
**src/subscription/index.ts** — Detect and handle Promise returns from subscribe().
**test/** — Added runtime and type tests for async subscribe.

All checks pass: pnpm run-all-checks + pnpm test (367 tests passed).

### DIFF
--- a/src/subscription/index.ts
+++ b/src/subscription/index.ts
@@ -72,13 +72,30 @@ export const subscription = (<Data = any, Error = any>(useSWRNext: SWRHook) =>
       subscriptions.set(subscriptionKey, refCount + 1)
 
       if (!refCount) {
-        const dispose = subscribe(args, { next })
-        if (typeof dispose !== 'function') {
+        const result = subscribe(args, { next })
+
+        if (result && typeof (result as any).then === 'function') {
+          // Race condition guard: if cleanup runs before the async subscribe
+          // resolves, the flag tells the resolver to dispose immediately.
+          let shouldDisposeOnResolve = false
+          ;(result as Promise<(() => void) | void>).then(dispose => {
+            if (shouldDisposeOnResolve) {
+              if (typeof dispose === 'function') dispose()
+            } else if (typeof dispose === 'function') {
+              disposers.set(subscriptionKey!, dispose)
+            }
+          })
+
+          disposers.set(subscriptionKey, () => {
+            shouldDisposeOnResolve = true
+          })
+        } else if (typeof result === 'function') {
+          disposers.set(subscriptionKey, result)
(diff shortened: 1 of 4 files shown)


src/subscription/index.ts
```diff
@@ -72,13 +72,30 @@ export const subscription = (<Data = any, Error = any>(useSWRNext: SWRHook) =>
       subscriptions.set(subscriptionKey, refCount + 1)
 
       if (!refCount) {
-        const dispose = subscribe(args, { next })
-        if (typeof dispose !== 'function') {
+        const result = subscribe(args, { next })
+
+        if (result && typeof (result as any).then === 'function') {
```
SukkaW: Last time I checked there is a `isThenable` util within SWR codebase.

SukkaW (changes requested):

src/subscription/index.ts
solssak: Thanks for the catch! You're right — I was reinventing the wheel here. I couldn't find an `isThenable` export, but there's an equivalent `isPromiseLike` util in `_internal/utils/shared.ts`, so I switched to that (along with `isFunction`/`isUndefined` for the other checks). This also let me drop the `as any` cast. 🙏

Issue #4164, opened 2025-08-18
Feature request: exported types / easier augmentation patterns


james-dawson-wegmans:
# Feature request

## Description

useSWR's types are extremely complex and difficult to replicate. I've been working on a variety of wrappers for `useSWR` in our project and because `swr` doesn't export most of its types, it has been nearly impossible to ensure that my wrappers offer the exact same API surface as useSWR.

## Expected Behavior

I'd expect to be able to import types like `Arguments`, `Key`, `State`, etc.

## Additional Context

Rationale for wanting this:

I created an in-project hook named `useAuthenticatedSWR`, which is meant to do everything `useSWR` does, but also taps into my project's authentication system to automatically pass `null` into the `swrKey` if auth isn't ready yet. Because of the complex types for useSWR, I had to limit the options that my custom wrapper can accept even though I'm just passing everything through.

---

I assume that perhaps function overloads are part of this complexity. I'm not sure if swr can provide an easier gateway to being extended in the future, but that's the hope. Wrapping something to configure it would ideally be a lot less difficult than it is.

Issue #2755, opened 2023-08-22
Trigger typing breaks typescript inference because of missing export


ckruppe:
# Bug report

## Description / Observed Behavior

Since Version 2.2.0 the new typing of the trigger function of useSwrMutation breaks type inference if its used in an own NPM module.

Exported variable 'useMutateApi' has or is using name 'TriggerWithArgs' from external module "C:/projekte/Intern/Libraries/nfq-typed-next-api/node_modules/.pnpm/swr@2.2.1_react@18.2.0/node_modules/swr/mutation/dist/index" but cannot be named.ts(4023)
---
Exported variable 'useMutateApi' has or is using name 'TriggerWithOptionsArgs' from external module "C:/projekte/Intern/Libraries/nfq-typed-next-api/node_modules/.pnpm/swr@2.2.1_react@18.2.0/node_modules/swr/mutation/dist/index" but cannot be named.ts(4023)
---
Exported variable 'useMutateApi' has or is using name 'TriggerWithoutArgs' from external module "C:/projekte/Intern/Libraries/nfq-typed-next-api/node_modules/.pnpm/swr@2.2.1_react@18.2.0/node_modules/swr/mutation/dist/index" but cannot be named.ts(4023)

## Expected Behavior

It would help to at least export these types so typescript can use them on compile time to include them in its own generated *.d.ts file.

## Repro Steps / Code Example

```typescript
/**
 * Get request hook.
 *
 * @param url        API URL.
 * @param swrOptions SWR options.
 * @returns SWR response.
 */
export const useMutateApi = <T extends ApiMethod>(
    url: string | null,
    swrOptions?: SWRMutationConfiguration<Jsonify<ApiResponse<T>['data']>, any> | undefined
) => {
    const {
        data,
        error,
        isMutating,
        reset,
        trigger
    } = useSwrMutation<Jsonify<ApiResponse<T>['data']>, RequestError<T>, Key, MutationRequestOptions<T>>(
        url ? `${basePath}${url}` : null,
        mutationFetcher,
        swrOptions
    );

    return {
        data,
        error,
        isMutating,
        reset,
        trigger
    };
};
```

## Additional Context

SWR version. 
Since Version 2.2.0

monolithed:
<img width="1016" alt="Screenshot 2023-08-27 at 03 32 31" src="https://github.com/vercel/swr/assets/803674/23306163-e7c6-469f-a775-5a238fee8aec">

der-daniel:
when will this fix be live?

PR #4245, opened 2026-04-13
fix(types): make useSWRMutation trigger callable in generic wrapper functions

### LINKED ISSUE #4234: useSWRMutation trigger is uncallable when wrapped in a generic function
# Bug report

When wrapping `useSWRMutation` in a generic helper function, the `trigger` return type becomes an uncallable union of `TriggerWithoutArgs | TriggerWithOptionsArgs | TriggerWithArgs`.

TypeScript error:

```
TS2349: This expression is not callable.
  Each member of the union type
  'TriggerWithoutArgs<Data, any, Key, Args> | TriggerWithOptionsArgs<Data, any, Key, Args> | TriggerWithArgs<...>'
  has signatures, but none of those signatures are compatible with each other.
```

The conditional type that determines which trigger interface to use:

```ts
trigger: [ExtraArg] extends [never]
  ? TriggerWithoutArgs<...>
  : IsUndefinedIncluded<ExtraArg> extends true
    ? TriggerWithOptionsArgs<...>
    : TriggerWithArgs<...>
```

cannot be evaluated when `ExtraArg` is a generic type parameter (not yet resolved). TypeScript **defers** the conditional type rather than resolving it to a single branch. The deferred type is opaque and uncallable because TS cannot determine which of the three incompatible call signatures applies.

Note: the `[ExtraArg] extends [never]` wrapping correctly prevents union distribution — this is a deferred conditional type issue, not a distributive conditional type issue.

## Expected Behavior

`trigger` should be callable when `useSWRMutation` is used inside a generic wrapper function.

## Repro Steps / Code Example

```ts
import type { MutationFetcher } from 'swr/mutation'

export const useMutationFetch = <
  Key extends string = string,
  Data = unknown,
  Args = unknown,
>(
  key: Key,
  fetcher: MutationFetcher<Data, Key, Args>
) => {
  const { trigger } = useSWRMutation(key, fetcher)

  // TS2349: This expression is not callable.
  trigger({ someArg: 'value' })
}
```

Currently requires `@ts-expect-error` to work around.

## Additional Context

SWR version: 2.2.4 (also reproduced on 2.4.1)

The three-interface split was introduced in https://github.com/vercel/swr/pull/2604.

Related issues:
- https://github.com/vercel/swr/issues/2755 — trigger types breaking in custom hooks (fixed by exporting types, but the underlying deferred conditional issue was not addressed)
- https://github.com/vercel/swr/issues/4164 — difficulty wrapping SWR hooks with type safety
- https://github.com/vercel/swr/issues/2126 — generic fetcher inference issues

## Proposed Fix

Collapse the three separate interfaces (`TriggerWithoutArgs`, `TriggerWithOptionsArgs`, `TriggerWithArgs`) into a single interface using conditional rest parameters. This pushes the conditional into the parameter position, where TypeScript handles deferral more gracefully — all branches share one callable signature with the same return type.

```ts
interface Trigger<Data, Error, SWRMutationKey extends Key, ExtraArg> {
  // throwOnError: true
  <SWRData = Data>(
    ...args: [ExtraArg] extends [never]
      ? [options?: SWRMutationConfiguration<SWRData, Error, SWRMutationKey, ExtraArg> & { throwOnError: true }]
      : IsUndefinedIncluded<ExtraArg> extends true
        ? [extraArgument?: ExtraArg, options?: SWRMutationConfiguration<SWRData, Error, SWRMutationKey, ExtraArg> & { throwOnError: true }]
        : [extraArgument: ExtraArg, options?: SWRMutationConfiguration<SWRData, Error, SWRMutationKey, ExtraArg> & { throwOnError: true }]
  ): Promise<SWRData>

  // throwOnError: false | undefined (default)
  <SWRData = Data>(
    ...args: [ExtraArg] extends [never]
      ? [options?: SWRMutationConfiguration<SWRData, Error, SWRMutationKey, ExtraArg> & { throwOnError?: false }]
      : IsUndefinedIncluded<ExtraArg> extends true
        ? [extraArgument?: ExtraArg, options?: SWRMutationConfiguration<SWRData, Error, SWRMutationKey, ExtraArg> & { throwOnError?: false }]
        : [extraArgument: ExtraArg, options?: SWRMutationConfiguration<SWRData, Error, SWRMutationKey, ExtraArg> & { throwOnError?: false }]
  ): Promise<SWRData | undefined>
}
```

By keeping all overloads within one interface, the `trigger` property type is always a single interface (not a conditional selecting between three), so it remains callable even when `ExtraArg` is an unresolved generic.

DORI2001:
## Problem

When `useSWRMutation` is used inside a generic wrapper function, the `trigger` return value becomes uncallable with TS2349:

```ts
export const useMutationFetch = <
  Key extends string = string,
  Data = unknown,
  Args = unknown,
>(
  key: Key,
  fetcher: MutationFetcher<Data, Key, Args>
) => {
  const { trigger } = useSWRMutation(key, fetcher)
  trigger({ someArg: 'value' })
  // TS2349: This expression is not callable.
  // Each member of the union type 'TriggerWithoutArgs<...> | TriggerWithOptionsArgs<...> | TriggerWithArgs<...>'
  // has signatures, but none of those signatures are compatible with each other.
}
```

**Root cause:** `trigger` is typed as a conditional type on `ExtraArg`. When `ExtraArg` is a free (not-yet-resolved) generic type parameter, TypeScript *defers* the conditional and represents the result as a union of all three branches — `TriggerWithoutArgs | TriggerWithOptionsArgs | TriggerWithArgs`. Since none of those call signatures are mutually compatible, the union is uncallable.

Fixes #4234. Related: #2755, #4164.

## Fix

Introduce `SWRMutationTrigger<Data, Error, SWRMutationKey, ExtraArg>`, a single interface that uses a **conditional rest-tuple** (`TriggerExtraArg<ExtraArg>`) to encode the argument shape:

```ts
type TriggerExtraArg<ExtraArg> = [ExtraArg] extends [never]
  ? [extraArgument?: null | undefined]
  : IsUndefinedIncluded<ExtraArg> extends true
  ? [extraArgument?: ExtraArg]
  : [extraArgument: ExtraArg]

export interface SWRMutationTrigger<Data, Error, SWRMutationKey extends Key, ExtraArg> {
  <SWRData = Data>(
    ...args: [...TriggerExtraArg<ExtraArg>, options?: SWRMutationConfiguration<...>]
  ): Promise<Data>
  // ... throwOnError overloads
}
```

Pushing the conditional *into the parameter position* means the interface itself is always a single callable type. TypeScript can call it even when `TriggerExtraArg<ExtraArg>` is deferred — it only can't call a *union of incompatible interfaces*.

`SWRMutationResponse.trigger` now resolves to `SWRMutationTrigger` instead of the old three-way conditional union.

## Backwards compatibility

The three legacy `TriggerWith*` interfaces are preserved and re-exported (now marked `@deprecated`). Existing code that references them by name will continue to compile. `SWRMutationTrigger` is also exported as a named type for users who need to annotate trigger callbacks.


DORI2001:
Closing this PR as SWR appears to be in maintenance mode and there's been no maintainer engagement in the 10 days since it was opened. The fix for the `useSWRMutation` TypeScript generics is still valid — happy to reopen if the team is interested.

PR #4252, opened 2026-05-03
fix: guard against double-unsubscribe removing wrong subscriber in cache.ts

### LINKED ISSUE #4251: Bug: double-calling unsubscribe removes wrong subscriber in cache.ts
# Bug report

The unsubscribe function returned by the internal `subscribe` helper in `src/_internal/utils/cache.ts` silently removes an **unrelated** subscriber when called more than once.

```typescript
// src/_internal/utils/cache.ts:59
subs.push(callback)
return () => subs.splice(subs.indexOf(callback), 1)
```

`Array.prototype.indexOf` returns `-1` when the callback is no longer in the array (i.e. already unsubscribed). `splice(-1, 1)` interprets `-1` as the last index and removes the **last subscriber in the array** instead of doing nothing.

## Expected Behavior

Calling an unsubscribe function more than once should be a no-op. The second call should not affect any other subscriber.

## Repro Steps / Code Example

```typescript
import { initCache } from 'swr/_internal'

const provider = new Map()
const [, , , , , , subscribe] = initCache(provider)

const callbackA = () => console.log('A')
const callbackB = () => console.log('B')

const unsubA = subscribe('key', callbackA)
subscribe('key', callbackB)

unsubA() // ✅ removes A correctly
unsubA() // 🐛 indexOf returns -1 → splice(-1, 1) removes B instead of no-op
// callbackB is now gone and will never receive cache updates
```

This can be triggered in practice by:
- React Strict Mode, which double-invokes effect cleanups
- Race conditions during rapid component mount/unmount

## Additional Context

SWR version: 2.4.1

Note: The same unsubscribe logic in `src/_internal/utils/subscribe-key.ts` already handles this correctly with an `index >= 0` guard and an O(1) swap-and-pop pattern. `cache.ts` should use the same approach.

tomohiro86:
## Summary

Fixes #4251

- `subscribe` in `src/_internal/utils/cache.ts` returned `() => subs.splice(subs.indexOf(callback), 1)`
- When the unsubscribe function was called a second time, `indexOf` returned `-1` and `splice(-1, 1)` silently removed the **last subscriber** in the array instead of being a no-op
- Fix adds an `index >= 0` guard and uses the same O(1) swap-and-pop pattern already present in `src/_internal/utils/subscribe-key.ts`

## Changes

- `src/_internal/utils/cache.ts`: align the unsubscribe logic with `subscribeCallback` in `subscribe-key.ts`

## Test plan

- [x] All existing tests pass (`pnpm test`: 360 passed, 5 skipped)
- [x] TypeScript type check passes (`tsc --noEmit`)
- [x] Build succeeds (`pnpm build`)

### DIFF
--- a/src/_internal/utils/cache.ts
+++ b/src/_internal/utils/cache.ts
@@ -56,7 +56,14 @@ export const initCache = <Data = any>(
       subscriptions[key] = subs
 
       subs.push(callback)
-      return () => subs.splice(subs.indexOf(callback), 1)
+      return () => {
+        const index = subs.indexOf(callback)
+        if (index >= 0) {
+          // O(1): swap with the last element and pop
+          subs[index] = subs[subs.length - 1]
+          subs.pop()
+        }
+      }
     }
     const setter = (key: string, value: any, prev: any) => {
       provider.set(key, value)


promer94 (approved):

promer94:
@tomohiro86 Could you signed the commit ?

tomohiro86:
@promer94 
Is this okay?

promer94:
Sorry I did not give you the correct context. 

@tomohiro86 You can follow this documention https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits to sign the commit.

tomohiro86:
@promer94 
I have committed with a signing.

promer94:
@tomohiro86 It’s required that you have to sign all your commits. You might need to rebase your current commits. Coding agent can help you with that.

tomohiro86:
Thanks for the heads-up! I've rebased and signed the commit — GitHub now shows it as verified.

PR #4253, opened 2026-05-09
fix(infinite): keep parallel getKey previousPageData null on setSize

li-jia-nan:
When useSWRInfinite runs with parallel enabled, getKey should always receive null as previousPageData because pages are independent. setSize previously recomputed page keys using the previous page data, which could produce keys inconsistent with the initial parallel load.

Update setSize to pass null in parallel mode and add a regression test.


li-jia-nan:
@shuding @huozhi Hi, could you please review this PR when you have time?

PR #4262, opened 2026-05-29
fix(subscription): delete bookkeeping Map entries on last unmount

### LINKED ISSUE #4261: subscription/index.ts L89-92: disposers and subscriptions Maps not cleared on last-subscriber teardown
## subscription/index.ts L89-92: disposers and subscriptions Maps not cleared on last-subscriber teardown

When the last subscriber unmounts, the cleanup callback calls `dispose()` but never removes the stale entries from the `subscriptions` and `disposers` Maps. In a long-lived app that cycles through many distinct subscription keys (e.g. paginated feeds, rotating live-data channels), both Maps accumulate one dead entry per unique key and grow without bound.

### Affected code

`src/subscription/index.ts`, lines 84-93

```ts
return () => {
  const count = subscriptions.get(subscriptionKey)! - 1

  subscriptions.set(subscriptionKey, count)

  // Dispose if it's the last one.
  if (!count) {
    const dispose = disposers.get(subscriptionKey)
    dispose?.()
    // <-- disposers and subscriptions entries are never deleted here
  }
}
```

### Why this is a bug

`subscriptions` (ref-count Map) and `disposers` (cleanup function Map) are allocated once per cache boundary and live for the lifetime of the cache object. Every key that ever passes through `useSubscription` leaves a permanent entry after its last subscriber unmounts:

- `subscriptions` keeps the key mapped to `0`.
- `disposers` keeps the key mapped to the (now-called) dispose function, preventing the function object from being garbage-collected.

In applications with a large or unbounded set of subscription keys (infinite scroll, user-navigated channels, rotating market tickers, etc.) this is a straightforward memory leak. The leak is proportional to the number of distinct keys ever used, not the number of currently-active subscriptions.

A secondary issue: because the zero-count entry remains in `subscriptions`, a future remount of the same key reads `refCount = 0` correctly (the `|| 0` fallback), so the functional behaviour is not broken -- only the memory footprint grows.

### Repro scenario

```tsx
// Rotate through N distinct keys over time.
// After unmounting each key the Maps retain stale entries.

function RotatingSubscription({ channel }: { channel: string }) {
  useSWRSubscription(channel, (key, { next }) => {
    const ws = new WebSocket(`wss://example.com/${key}`)
    ws.onmessage = (e) => next(null, e.data)
    return () => ws.close()
  })
  return null
}

// Mount/unmount with channel = "ch-0", "ch-1", ..., "ch-9999"
// After all unmounts: subscriptions.size === 10000, disposers.size === 10000
```

### Suggested fix

Delete the entries from both Maps immediately after calling `dispose()`:

```ts
if (!count) {
  const dispose = disposers.get(subscriptionKey)
  dispose?.()
  disposers.delete(subscriptionKey)      // <-- add
  subscriptions.delete(subscriptionKey)  // <-- add
}
```

This restores the Maps to a size proportional to the number of *currently active* subscription keys rather than all keys ever seen.

Reporter: Nexory

spokodev:
Fixes #4261.

The cleanup callback at `src/subscription/index.ts:84-93` calls `dispose()` when the last subscriber for a key unmounts, but never removes the matching entries from the `subscriptions` and `disposers` Maps. In a long-lived app that cycles through many distinct subscription keys (paginated feeds, rotating live-data channels, user-navigated rooms), both Maps accumulate one dead entry per unique key and grow without bound.

This change deletes the entry from both Maps right after calling `dispose()`:

```ts
if (!count) {
  const dispose = disposers.get(subscriptionKey)
  dispose?.()
  subscriptions.delete(subscriptionKey)
  disposers.delete(subscriptionKey)
}
```

The functional behaviour is unchanged: a subsequent re-mount for the same key still hits the `if (!refCount)` branch (because `subscriptions.get(key) || 0` is `0` whether the entry is missing or zero), so a fresh subscribe runs and a fresh disposer is registered.

## Tests

`test/use-swr-subscription.test.tsx` gets a `teardown cleans up bookkeeping Maps (#4261)` describe block with 7 new cases covering both axes of the fix.

Because the bookkeeping Maps are module-private (closed over inside `subscription/index.ts` via a per-cache `WeakMap`), a minimal `@internal` helper is exported so the tests can read their sizes:

```ts
export const _unstable_getSubscriptionStateSize = (
  cache: object
): { subscriptions: number; disposers: number } | null
```

The naming follows the standard `_unstable_` prefix convention and the helper is annotated `@internal` so it does not enter the documented public API. Suggestions on a better location welcome.

Positive (the fix's contract)
- After the last subscriber unmounts, both Maps drop the entry (sizes go to 0).

Negative (existing behavior preserved)
- While another subscriber is still mounted, the entry stays in both Maps (refcount > 0).
- `dispose()` is still called exactly once on the final unmount.
- Re-mounting the same key after teardown runs a fresh subscribe and produces a fresh disposer.

Edge
- Distinct keys are isolated. Unmounting one does not affect another active subscription on the same cache.
- N mount/unmount cycles on distinct keys (the leak scenario from the issue) leave the Maps at size 0.
- Per-cache scope. Cleaning up one `SWRConfig`'s subscription does not touch the bookkeeping of another `SWRConfig`.

All 7 fail on `main` against the relevant assertions and pass on this branch. The 7 pre-existing tests in the same file are unchanged and still pass.

## Verification

```
pnpm test
# 367 passed, 5 skipped (372)

pnpm test test/use-swr-subscription.test.tsx
# 14 passed

pnpm types:check
# clean

pnpm exec eslint src/subscription/index.ts test/use-swr-subscription.test.tsx
# clean

pnpm prettier --check src/subscription/index.ts test/use-swr-subscription.test.tsx
# All matched files use Prettier code style!
```

Opened as draft until CI is green.

### DIFF
--- a/src/subscription/index.ts
+++ b/src/subscription/index.ts
@@ -90,6 +90,10 @@ export const subscription = (<Data = any, Error = any>(useSWRNext: SWRHook) =>
         if (!count) {
           const dispose = disposers.get(subscriptionKey)
           dispose?.()
+          // Drop the stale entries so the per-cache subscription Maps don't
+          // grow without bound when an app cycles through distinct keys.
+          subscriptions.delete(subscriptionKey)
+          disposers.delete(subscriptionKey)
         }
       }
     }, [subscriptionKey])
--- a/test/use-swr-subscription.test.tsx
+++ b/test/use-swr-subscription.test.tsx
@@ -1,7 +1,8 @@
 import { fireEvent, screen } from '@testing-library/react'
 import { sleep, renderWithConfig, createKey } from './utils'
 import useSWRSubscription from 'swr/subscription'
-import useSWR from 'swr'
+import useSWR, { SWRConfig } from 'swr'
+import { render } from '@testing-library/react'
 import { useEffect, useState, act } from 'react'
 import { ErrorBoundary } from 'react-error-boundary'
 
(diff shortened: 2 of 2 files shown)


nahrinoda (commented):
The fix itself (the two `delete` calls) is correct and identical to #4268.

The concern is with `_unstable_getSubscriptionStateSize`: since it's a named export from `src/subscription/index.ts`, it will appear in the `swr/subscription` `.d.ts` declarations and become importable by consumers. The `@internal` JSDoc tag is a convention — TypeScript does not enforce it, and bundlers/IDE autocomplete will surface it. SWR has no precedent for `_unstable_` exports in its source (the existing `@internal` annotations in `_internal/types.ts` are on interfaces within the `_internal` subpath, which is already scoped by convention).

The fix is provably correct without runtime introspection: `subscriptions.get(key) || 0` treats a missing key identically to a zero-count key, so deleting the entry does not change the re-mount path. That reasoning (already stated in the PR body) is the test — no observable behavior changes.

If the maintainers do want a regression test, an alternative that avoids a public export: spy on the `subscribe` return value (the dispose function) and assert it's called exactly once on last unmount, plus assert a fresh `subscribe` fires on re-mount of the same key. Both are observable without internal access and already partially covered by the existing `should deduplicate subscriptions` test. The N2 and N3 tests in this PR do exactly that pattern but could stand alone without the `@internal` helper.

spokodev:
Thanks, fair points. I dropped the `_unstable_` export, so the bookkeeping state sits in an internal module now and nothing is added to the public surface. And you're right that the fix is memory-only, so the size tests can come down to just the behavioral ones (or none) if that reads cleaner. Whichever shape you and the maintainers prefer between this and #4268 is fine by me; the main thing is the leak gets closed.

spokodev:
Followed your suggestion - reworked the test to spy on the `subscribe` dispose return value and assert the public contract (dispose called once on last unmount, fresh subscribe on re-mount), with no internal export. I also inlined the bookkeeping state back so the production diff is just the two deletes, and rebased onto main. It now passes under both the source and build jest configs.

nahrinoda (approved):
Clean rework. The production diff is now just the two deletes, tests assert the public contract through subscribe/dispose call counts, and nothing leaks into the API surface. LGTM.

PR #4271, opened 2026-06-17
Prevent resolved promise to suspend due to missing status

devjiwonchoi:
### Why?

Current behavior: https://github.com/vercel/swr/actions/runs/27748306911/job/82092031539?pr=4271#step:4:714

In SWR Suspense mode, when the promise is already fulfilled, React does not have to suspend. However, there are paths where SWR consumes an internal promise that resolves to `undefined` instead, and because that promise did not have `status` attached, React could not know it's status on the initial consume, so it suspended.

This caused the hook to unnecessarily suspend in conditions where it did not really have to on the client, such as during a sync update or hydration, which could show a Suspense fallback even after the SSR'd content has been rendered.

By marking the internal `undefined` promise as fulfilled, SWR ensures consuming that does not suspend and only suspends when the actual data promise is not fulfilled.

React thenable x-ref: https://github.com/react/react/blob/560db51408f4c186016ccdd8062af75a77c22104/packages/react-reconciler/src/ReactFiberThenable.js#L187-L302

### How?

- Mark SWR's internal `resolvedUndef` promise as fulfilled with an `undefined` value so `use(resolvedUndef)` returns synchronously.
- Add regression coverage for fulfilled promise `fallbackData` and null-key Suspense paths; these tests fail before the fix and pass after it.

### DIFF
--- a/src/index/use-swr.ts
+++ b/src/index/use-swr.ts
@@ -89,7 +89,6 @@ type DefinitelyTruthy<T> = false extends T
   ? never
   : T
 
-const resolvedUndef = Promise.resolve(UNDEFINED)
 const sub = () => noop
 /**
  * The core implementation of the useSWR hook.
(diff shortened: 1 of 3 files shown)


test/use-swr-server.test.tsx
```diff
@@ -88,6 +91,40 @@ describe('useSWR - SSR', () => {
...
+        render(
+          <Suspense fallback={<div>loading</div>}>
+            <Page />
+          </Suspense>
+        )
+
+        // Any fallback render means SWR suspended despite fulfilled fallback data.
+        const loadingFallback = screen.queryByText('loading')
+        expect(loadingFallback === null).toBe(true)
+        screen.getByText('fallback data')
```
lubieowoce, line 127: does this account for the fallback flashing and then disappearing on a rerender? i.e. does this test actually fail before the change?

devjiwonchoi: Yeah I had it red green

https://github.com/vercel/swr/actions/runs/27748306911/job/82092031539?pr=4271#step:4:557

src/index/use-swr.ts
```diff
@@ -790,23 +789,24 @@ export const useSWRHandler = <Data = any, Error = any>(
 
     const req = PRELOAD[key]
 
-    const mutateReq =
-      !isUndefined(req) && hasKeyButNoData ? boundMutate(req) : resolvedUndef
-    use(mutateReq)
+    if (!isUndefined(req) && hasKeyButNoData) {
+      use(boundMutate(req))
```
huozhi: Do we have to make `use()` under condition? Seb raised an issue related to it before so we changed. I'm not sure if this is still an issue with React

test/use-swr-server.test.tsx
```diff
@@ -88,6 +91,40 @@ describe('useSWR - SSR', () => {
         )
       })
     })
+
+    it('should not suspend when fallbackData is a fulfilled promise', async () => {
+      await withServer(async () => {
+        const useSWR = (await import('swr')).default
```
huozhi, line 99: is that possible to make a e2e test instead of mocking a lot with this e2e test, in that way we can avoid mocking react internal states

devjiwonchoi: I tried write e2e but wasn't able to capture the fallback state because React just commits the content before even show fallback, likely as it handles in a micro task and the fixture is lightweight that it just commits the rerender before the fallback

src/index/use-swr.ts
devjiwonchoi: Updated!

x-ref: https://github.com/facebook/react/pull/34030

huozhi (approved):

PR #4275, opened 2026-06-20
feat: rsc fetches and prefill swr cache

### LINKED ISSUE #4065: Feature request: when RSC data fetching set as the fallback, skip onmount revalidation
### Feature Request

https://x.com/asidorenko_/status/1877342799869194603

when there's already an RSC data fetching as the fallback, skip the first revalidation when onmount

huozhi:
Add RSC cache preloading for SWR.

This lets a Server Component call `preload(key, fetcher)` and pass the returned `cacheData` object through the server/client boundary into `SWRConfig`. On the client, `useSWR` consumes that cache data before calling the browser fetcher, returns the server-loaded value, and writes it into the SWR cache after hydration.

Flow:

```txt
Server Component
  preload(key, fetcher)
        | returns { [serializedKey]: dataOrPromise }
        v
Client Boundary
  <SWRConfig value={{ cacheData }}>
        |
        v
useSWR(key, clientFetcher)
  - uses server-loaded cache data for initial data
  - hydrates the SWR cache after hydration
  - skips duplicate initial browser fetch/revalidation
  - later mutate/focus/reconnect revalidation still uses clientFetcher
```

Example:

```tsx
// Server Component
import { preload } from 'swr'
import { ClientRoot } from './client'

const key = '/api/user'

async function getUser() {
  return { name: 'Ada' }
}

export default function Page() {
  const cacheData = preload(key, getUser)
  return <ClientRoot cacheData={cacheData} />
}
```

```tsx
// Client Component
'use client'

import useSWR, { SWRConfig } from 'swr'
import type { CacheData } from 'swr'

const key = '/api/user'

async function fetchUser() {
  const res = await fetch(key)
  return res.json()
}

function User() {
  const { data } = useSWR(key, fetchUser)
  return <div>{data.name}</div>
}

export function ClientRoot({ cacheData }: { cacheData: CacheData }) {
  return (
    <SWRConfig value={{ cacheData }}>
      <User />
    </SWRConfig>
  )
}
```

`cacheData` is object-shaped, so nested configs merge like `fallback` and duplicate keys are replaced by the later config. The tests cover Suspense and default modes, prevent leaking the internal cache-data record into user data, and verify that explicit later revalidation still calls the client fetcher.

Resolves #4065

### DIFF
--- a/e2e/site/app/rsc-unstable-preload-no-suspense/page.tsx
+++ b/e2e/site/app/rsc-unstable-preload-no-suspense/page.tsx
@@ -7,7 +7,7 @@ export const dynamic = 'force-dynamic'
 
 async function getServerData() {
   await sleep(150)
-  return 'SWR_RSC_PRELOAD_FLIGHT_MARKER_20260621'
+  return 'SWR_RSC_PRELOAD_NO_SUSPENSE_MARKER_20260621'
 }
 
 export default function Page() {
--- a/e2e/site/app/rsc-unstable-preload-no-suspense/page.tsx
+++ b/e2e/site/app/rsc-unstable-preload-no-suspense/page.tsx
@@ -1,17 +1,19 @@
+import { connection } from 'next/server'
 import { preload } from 'swr'
 import type { CacheData } from 'swr'
 import { sleep } from '~/lib/sleep'
 import { ClientRoot } from './client'
 import { key } from './key'
 
-export const dynamic = 'force-dynamic'
-
 async function getServerData() {
   await sleep(150)
   return 'SWR_RSC_PRELOAD_NO_SUSPENSE_MARKER_20260621'
 }
 
-export default function Page() {
+export default async function Page() {
+  // Opt into dynamic rendering so each request preloads fresh server data.
+  await connection()
+
   // Runtime resolves the react-server export; this app's type checker still
   // reads the default client preload signature.
   const cacheData = preload(key, getServerData) as unknown as CacheData<string>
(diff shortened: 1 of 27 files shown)


src/_internal/utils/config-context.ts
```diff
@@ -56,6 +55,24 @@ const SWRConfig: FC<
...
+        isUndefined(PRELOAD[key])
+      ) {
+        PRELOAD[key] = {
+          data: cacheData[key],
+          _cacheData: true
+        }
+      }
+    }
+  }
+
```
shuding: This is problematic. We can't mutate the `PRELOAD` object as it can be shared across requests. We need to find a way to create a new config (including its cache and preload) on every server render.

Since we don't strictly require the SWR provider, maybe we'll have to rely on `React.cache`.

e2e/site/app/rsc-unstable-preload-no-suspense/page.tsx
```diff
@@ -0,0 +1,20 @@
...
+  return 'SWR_RSC_PRELOAD_NO_SUSPENSE_MARKER_20260621'
+}
+
+export default function Page() {
+  // Runtime resolves the react-server export; this app's type checker still
+  // reads the default client preload signature.
+  const cacheData = preload(key, getServerData) as unknown as CacheData<string>
+
+  return <ClientRoot cacheData={cacheData} />
+}
```
shuding: How to reproduce the issue:

Keep the same dev server running, add a search param to the page and preload conditionally:

```
const cacheData = params.preload ? preload(key, getServerData) as unknown as CacheData<string> : {}
```

And visit with ?preload first, then without the param. If the page still renders with the preloaded cache it's wrong.

shuding (approved):

yongsk0066:
@huozhi @shuding we've been wanting exactly this, so I installed the beta right away and tried it in our app. Works really nicely with our RSC setup. One small thing I ran into: hooks subscribing with a `null` fetcher never get `cacheData`, which is how most of our `fallback` pages read server data today. Proposed a small fix in #4293.

PR #4281, opened 2026-06-23
Make suspense use branches explicit

devjiwonchoi:
### Why?

Make the Suspense `use(...)` paths easier to debug by replacing ternary-selected promises with explicit branches. This lets a debugger land directly on the preload mutation path, revalidation path, or resolved no-op path.

### DIFF
--- a/src/index/use-swr.ts
+++ b/src/index/use-swr.ts
@@ -796,23 +796,27 @@ export const useSWRHandler = <Data = any, Error = any>(
 
     const req = PRELOAD[key]
 
-    const mutateReq =
-      !isUndefined(req) && hasKeyButNoData ? boundMutate(req) : resolvedUndef
-    use(mutateReq)
+    if (!isUndefined(req) && hasKeyButNoData) {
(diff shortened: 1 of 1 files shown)


saintparish4:
While making the suspense use() branches explicit, the if (suspense) guard in shouldStartRequest (use-swr.ts:228) is worth folding into the same cleanup: both arms return the identical isUndefined(data) || revalidateIfStale, so the guard is dead. It also reads inconsistently against the sibling shouldDoInitialRevalidation block (use-swr.ts:379), which uses isUndefined(data) ? false : revalidateIfStale. I added a characterization test over the suspense initial-revalidation matrix (no- data -> suspends; cached + revalidateIfStale:false -> no fetch; cached + revalidateIfStale:true -> revalidates once); the whole suspense suite stays greeen under both expressions, so they only differ in the suspense+no-data case, which suspends before the snapshot surfaces -- the two are behavior-equivalent and line 228 can be simplified (or aligned with 379) safely. Happy to PR the cleanup + test.

PR #4282, opened 2026-06-24
Avoid creating new promises for deduped revalidation

devjiwonchoi:
> [!TIP]
> Recommended to review commit by commit and hide whitespace.

### Why?

SWR already dedupes the actual fetch, so repeated renders share one network request.

But `revalidate` was an `async` function, which meant every call returned a new promise. With `suspense: true`, React could see a different promise from `revalidate()` on each render, even though SWR was waiting for the same deduped request underneath.

SWR worked around this by instrumenting the promise with `status` and `value`. This change instead uses a cached revalidation promise so `use()` does not see a new promise for the same request.

### How?

Store the current deduped revalidation promise in `FETCH[key]` and reuse it while the same request is active. The existing fetch cleanup still clears the whole `FETCH[key]` entry, including the cached promise.

When previous data is available, start revalidation but skip `use(revalidation)`, so SWR keeps showing the previous data instead of suspending.

### DIFF
--- a/src/index/use-swr.ts
+++ b/src/index/use-swr.ts
@@ -96,6 +96,11 @@ const resolvedUndef = Promise.resolve(UNDEFINED)
 resolvedUndef.status = 'fulfilled'
 // @ts-ignore modify react promise value
 resolvedUndef.value = UNDEFINED
+const resolvedFalse = Promise.resolve(false)
+// @ts-ignore modify react promise status
+resolvedFalse.status = 'fulfilled'
+// @ts-ignore modify react promise value
+resolvedFalse.value = false
 const sub = () => noop
 /**
  * The core implementation of the useSWR hook.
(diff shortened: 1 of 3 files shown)


src/index/use-swr.ts
```diff
@@ -803,16 +827,15 @@ export const useSWRHandler = <Data = any, Error = any>(
...
-      // @ts-ignore modify react promise value
-      revalidation.value = true
+    if (hasKeyButNoData) {
+      // A cache miss for the current key must start its request here.
+      // `returnedData` may be previous data, but it doesn't populate this key's cache.
+      const revalidation = revalidate(WITH_DEDUPE)
+      if (isUndefined(returnedData)) {
+        // No current or previous data is available to render, so suspend on the request.
+        use(revalidation)
+      }
```
eps1lon, line 837: isn't that still conditionally suspending on data usage? If we have data, we don't suspend. If we don't have data, we do suspend. I don't get the relationship between revalidation and returned data. I feel like we should just always use() the data and SWR should set the revalidation Promise into the data if we don't have any currently.

PR #4285, opened 2026-07-02
fix(hash): coerce an invalid Date to a stable string key

spokodev:
## Description

An invalid `Date` in an SWR key hashes to the value `null` instead of a string, so it collides with a literal `null` key and lets two different requests share one cache entry.

`stableHash` hashes a `Date` via `arg.toJSON()`. Per spec, `Date.prototype.toJSON()` returns `null` for an invalid date, whereas every other leaf branch of the hash returns a string. So:

```js
useSWR(['/api/events', filterDate]) // filterDate came from user/URL input and failed to parse
```

hashes to the same key as `['/api/events', null]`. Two distinct requests map to one cache slot and serve each other's data. It also breaks the `stableHash: (arg) => string` return-type contract, since `serialize` then returns a non-string key.

## Fix

```diff
-      ? arg.toJSON()
+      ? arg.toJSON() || arg.toString()
```

Valid dates keep their ISO string (`toJSON()` is truthy); an invalid date falls back to `"Invalid Date"`, a stable string distinct from `null`.

## Test

Added two cases to `test/unit/serialize.test.ts`: an invalid `Date` hashes to a string, and an invalid `Date` in an array key does not collide with `null`. Both fail before the change and pass after.

### DIFF
--- a/src/_internal/utils/hash.ts
+++ b/src/_internal/utils/hash.ts
@@ -64,7 +64,7 @@ export const stableHash = (arg: any): string => {
     }
   } else {
     result = isDate
-      ? arg.toJSON()
+      ? arg.toJSON() || arg.toString()
       : type == 'symbol'
       ? arg.toString()
       : type == 'string'
--- a/test/unit/serialize.test.ts
+++ b/test/unit/serialize.test.ts
@@ -14,3 +14,15 @@ describe('SWR - unstable_serialize', () => {
     )
   })
 })
+
+describe('SWR - stableHash', () => {
+  it('should hash an invalid date to a string', async () => {
+    expect(typeof stableHash(new Date('invalid'))).toBe('string')
+  })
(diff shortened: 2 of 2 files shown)


huozhi (commented):
the different invalid key will be the same tho, it might cause other issue, it's better to handle outside to avoid passing invalid date as key

Image

spokodev:
You are right that every invalid Date maps to the same key. The change is narrower than that: today an invalid Date's `toJSON()` returns `null`, so it hashes to the same key as a literal `null` or an unset field, which is a collision across unrelated keys. Falling back to `toString()` only stops an invalid Date from colliding with `null`. Different invalid Dates sharing one key seems acceptable since they carry no distinct value, but if you would rather reject an invalid Date at the call site, I am glad to close this in favor of that.

PR #4293, opened 2026-07-08
fix: hydrate cacheData for hooks without a fetcher

yongsk0066:
## Summary

Hydrate request-scoped `cacheData` into SWR's cache for `useSWR(key, null)` without treating hydration as revalidation.

- Support plain, promised, and rejected `cacheData` values.
- Keep each key one-shot per request-scoped `cacheData` object.
- Share consumption state across fetcher, fetcherless, and Suspense paths.
- Avoid client fetching, validation state, retries, and revalidation callbacks during hydration.

## Problem

Default-mode `cacheData` was consumed inside `revalidate`. A hook without a fetcher never enters that path, so `useSWR(key, null)` remained undefined and the server-loaded value was never written into SWR's cache.

Tracking consumption only by serialized key also cannot distinguish a fresh request-scoped `cacheData` object from an already-consumed one. Promised values additionally need protection from key changes and newer local mutations.

## Solution

After React commits, a fetcherless hook resolves its matching `cacheData` value and writes the resulting data or error into the external cache. The commit proceeds only while:

- the hook still owns the same key;
- no newer mutation has started; and
- the cache entry remains empty.

A per-hook `WeakMap<CacheData, Set<string>>` tracks consumption by request object and serialized key. This lets a fresh request provide new data for the same key while old request objects remain eligible for garbage collection.

Hydration does not call `revalidate`, so `isLoading` and `isValidating` remain false, revalidation callbacks do not fire, and no client fetcher runs.

## Validation

- Jest: 34 suites passed; 374 tests passed and 5 skipped.
- Playwright: 7 promise-scenario tests plus dev-server warmup passed.
- `pnpm run-all-checks` passed.
- `pnpm format:check` passed.
- `pnpm build` passed.

The browser regression verifies real RSC hydration with `useSWR(key, null)`: the server value reaches both the hook and SWR cache, loading and validating stay false, success callbacks remain at zero, and no client request occurs.

### DIFF
--- a/src/index/use-swr.ts
+++ b/src/index/use-swr.ts
@@ -35,7 +35,8 @@ import type {
   SWRHook,
   RevalidateEvent,
   StateDependencies,
-  GlobalState
+  GlobalState,
+  CacheData
 } from '../_internal'
 
 const use =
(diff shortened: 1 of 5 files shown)


huozhi (approved):
Thanks for the PR, I tweaked the implementation a bit so you don't need the extra effect hook

PR #4296, opened 2026-07-12
fix: allow null for the fetcher config option

### LINKED ISSUE #2888: Type error when overriding global fetcher function with null
# Bug report

When a global fetcher is configured using SWRConfig it can't be overridden locally by setting fetcher to null - a request still happens.

```
  // This results in a request to "/state"
  const first = useSWR("state", null);
```

A workaround is to set fetcher to null in the config parameter, but that does not validate in TS so a ts-ignore comment is needed.

```
  // This works but TS complains:
  // "Type 'null' is not assignable to type '((arg: "state") => any) | undefined'."
  const second = useSWR("state2", null, {
    fetcher: null,
  });

```
## Expected Behavior

No typescript errors displayed. In best case a null passed as fetcher should override global config.

## Repro Steps / Code Example

https://codesandbox.io/p/sandbox/useswr-typing-bug-2d4prc?file=%2Fsrc%2FApp.tsx%3A10%2C32

## Additional Context

SWR version 2.2.4

ankit1324:
Fixes #2888

## Problem

Overriding an inherited fetcher (from `SWRConfig`) with `{ fetcher: null }` already works at runtime — `mergeConfigs` spreads the local config over the context config, so the merged `config.fetcher` becomes `null` and `withArgs` resolves `fn || config.fetcher || null` to `null`, disabling fetching. But the `SWRConfiguration` type declares `fetcher?: Fn`, so the only working override doesn't typecheck and needs a `@ts-ignore`:

```tsx
// runtime: no request. types: error.
useSWR('state', null, { fetcher: null })
```

(Note the `null` second argument alone can't do it: `withArgs` intentionally falls back to the context fetcher there, and changing that would be a behavioral break — so the config override is the supported escape hatch, it just wasn't typed.)

## Fix

Type the option as `fetcher?: Fn | null` and document that `null` disables fetching. `withArgs` is already null-safe (single consumer of `config.fetcher`).

## Tests

- Runtime: with a global fetcher in `SWRConfig`, `useSWR(key, null, { fetcher: null })` renders idle, never calls the global fetcher.
- Type tests: `fetcher: null` accepted for `useSWR` (2- and 3-arg forms) and `useSWRInfinite`.
- `tsc --noEmit`, oxlint, and the config test suite pass.

### DIFF
--- a/src/_internal/types.ts
+++ b/src/_internal/types.ts
@@ -259,9 +259,10 @@ export interface PublicConfiguration<
    */
   strictServerPrefetchWarning?: boolean
   /**
-   * the fetcher function
+   * the fetcher function, or `null` to disable fetching (e.g. to override an
+   * inherited fetcher from `SWRConfig`)
    */
-  fetcher?: Fn
+  fetcher?: Fn | null
   /**
    * array of middleware functions
    * @see {@link https://swr.vercel.app/docs/middleware}
--- a/test/type/option-fetcher.ts
+++ b/test/type/option-fetcher.ts
@@ -516,3 +516,11 @@ export function useReturnReadonlyTuple() {
     }
   )
 }
+
+export function useNullFetcherOverride() {
+  // `fetcher: null` disables fetching, overriding an inherited fetcher
+  // from `SWRConfig` (https://github.com/vercel/swr/issues/2888).
+  useSWR('/api/user', null, { fetcher: null })
+  useSWR<{ id: number }>('/api/', { fetcher: null })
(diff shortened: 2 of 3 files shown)


ankit1324:
Closing — I missed that #4290 already makes the identical `fetcher?: Fn | null` type change (opened a week before this). Deferring to that PR. One thing from here that may be worth lifting into #4290: a runtime regression test that a local `{ fetcher: null }` override actually prevents the inherited `SWRConfig` fetcher from being called.

PR #4303, opened 2026-07-26
fix: skip isLoading when fallback data is available

### LINKED ISSUE #3046: `revalidateOnMount` With Fallback triggers `isLoading` instead of `isValidating`
# Bug report

I have Next App Router Server Fetch to cache the initial response and then use SWR to have live-updates. I pass the initial server fetch as fallback data to simplify the code.

When a user interacts and changes the entire behavior, there are places I need to set `isLoading` to show that the user's interaction triggered a loading (imagine "jump to page 30 of the table"). So I have like `isLoading ? <Skeleton /> : <Content />`. 

However, even when SWR has fallback data, it first triggers `isLoading`.

## Expected Behavior

It should trigger `isValidating` instead of `isLoading`, since the initial data has been provided.

okxint:
When a component provides `fallbackData` or a `SWRConfig` `fallback` map, the hook resolves a non-undefined `fallback` value at line 183. However the `isLoading` guard at line 479 only checks the internal cache — it misses the resolved fallback entirely.

The comment on the guarded block even says "This bypasses fallback data and laggy data", but the condition didn't actually bypass fallback. Result: `isLoading` fires as `true` on initial render even though the UI is already showing fallback data, causing unnecessary loading skeletons in SSR and App Router setups.

Add `&& isUndefined(fallback)` so `isLoading` is only set when there's genuinely nothing to show — no cached data and no fallback.

Closes #3046

### DIFF
--- a/src/index/use-swr.ts
+++ b/src/index/use-swr.ts
@@ -476,7 +476,7 @@ export const useSWRHandler = <Data = any, Error = any>(
       const initialState: State<Data, Error> = { isValidating: true }
       // It is in the `isLoading` state, if and only if there is no cached data.
       // This bypasses fallback data and laggy data.
-      if (isUndefined(getCache().data)) {
+      if (isUndefined(getCache().data) && isUndefined(fallback)) {
         initialState.isLoading = true
       }
       try {


CarlOlson:
Going off of [Key Change + Previous Data + Fallback](https://swr.vercel.app/docs/advanced/understanding#key-change--previous-data--fallback) it seems like the current behaviour is intended, reading further:

> Simply saying you can use isValidating for indicating everytime there is an ongoing revalidation, and isLoading for indicating that SWR is revalidating but there is no data yet to display.
>
> Fallback data and previous data are not considered "loaded data," so when you use fallback data or enable the keepPreviousData option, you might have data to display.

Doing something similar to TanStack's `useQuery` and adding more booleans like `isPending`, `isLoading`, `isFetching`, etc, might be easier to understand. As-is I often need to check multiple values to determine the correct UI and therefore often need to consult the docs.

